PR #43, opened 2015-06-03
Fira Mono: Correct category in METADATA.json

sapics:
I guess that Fira Mono font category is Monospace.


davelab6:
Thanks @sapics :)

PR #48, opened 2015-06-04
Added Digital Numbers font

s-a:


davelab6:
Awesome! Thanks for this - we don't have anything like this and we should :) 

However, there are a few issues that we'll need to resolve before this can be merged; I've posted them in your repo's issue tracker

s-a:
@davelab6  Thank you for reporting the issues. It is my first font I created for a game I am writing. So I could need some help and wrote you back there. I d like to fix the issues and create a new PR.

PR #50, opened 2015-06-08
Update khmer_unique-glyphs.nam

davelab6:
* Adding 0x17DD for Bunong language (CMO, Cambodia, Khmer script) requested by Diethelm Kanjahn at SIL


rsheeter:
Expected to just see addition of one codepoint but files changed is also showing me removal of a bunch (basically latin; 1 addition and 99 deletions.). I don't think we want to merge as-is.

davelab6:
The other name files don't have latin glyphs; eg https://github.com/google/fonts/blob/master/tools/encodings/greek_unique-glyphs.nam

rsheeter:
That is currently intentional (and odd). We could explore changing it but
doing it separate of adding the new codepoint would be best.

davelab6:
Okay I rebased so that 2nd commit is gone :)

rsheeter:
LGTM.

PR #108, opened 2015-08-27
AmaticSC-Regular: various fixes

Ziriax:
* removed GPOS table so it works with Windows/GDI
* changed the usWidthClass OS/2 table to be the same as the bold variant (stretch = medium). This caused the regular font to appear as bold in some applications and browsers
* bumped the version to 1.004

NOTE: This was done using TTX from the Google FontTools repo


davelab6:
Thank you @Ziriax :)

PR #136, opened 2015-10-13
Add instructions for installing on Windows

NomNuggetNom:
I added instructions for installing all the fonts on Windows. Not sure if this fits within the repo or is necessary, but I figured it might be helpful and allow people to use these fonts easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,14 @@ You can also sync the collection with git, so that you can update by only fetchi
 To learn how to use git, Github provides [illustrated guides](https://guides.github.com) and a [youtube channel](https://www.youtube.com/user/GitHubGuides), and a [learning game that takes just 15 minutes](https://try.github.io). 
 Free, open source git applications are available for [Windows](https://msysgit.github.io) and [Mac OS X](http://gitx.laullon.com).
 
+### Install on Windows
+
+You can install all of the fonts using power shell. Change directories to the root folder where you downloaded the package, and run the following command:
+```
+$fonts = (New-Object -ComObject Shell.Application).Namespace(0x14)
+dir ofl/*/*.ttf | %{ $fonts.CopyHere($_.fullname) }
+```
+
 ## Licensing
 
 It is important to always read the license for every font that you use.


tomByrer:
These fonts are optimized for web browsers.  Though perhaps OK for OS use, they've had font info removed that may be used in printing, etc.

Besides, `right-click [install]` works just fine; how many will want to install 1/3Gb of fonts?

(Though info is handy; put in blog/gist/twitter.)

davelab6:
@NomNuggetNom Thanks for this! Would you be happy to see this added to the TRIVIA.md file instead? I think many people in the GF audience won't know what powershell is, but this is a great tip to have around :)

@tomByrer The fonts should work when installed locally; if not, its a bug :)

rsheeter:
I would add that fonts aren't optimized for web use in this repo; they get cut up, converted into optimal format for user-agent, have features removed, and so on downstream.

NomNuggetNom:
Sure thing @davelab6. I submitted a new PR: https://github.com/google/fonts/pull/181

tomByrer:
> they get cut up, converted into optimal format for user-agent, have features removed, and so on downstream

Ah, I thought the output would be here.  So where is that optimization suite please?  I thought  [googlefonts/fontbakery](https://github.com/googlefonts/fontbakery) at first, but now it seems to be more of a test suite.

rsheeter:
I don't think there is something I could point to and say "this is the optimization suite".

The tools (fontTools, woff2, etc) and data (fonts, nam files to say what codepoints go into a given subset) we rely on are all open source. 

Does that help?  Are there specific details that are of interest?

davelab6:
As I understand it, the reason why don't provide such files is because when people want to self-host, they are unlikely to get web font files from us anyway, and likely to roll their own. 

Perhaps what you are expecting is a 'bullet proof' CSS example, and accompanying EOT/WOFF/WOFF2 files. We don't provide that, but https://github.com/majodev/google-webfonts-helper does.

tomByrer:
> The tools (fontTools, woff2, etc) ...we rely on are all open source.

I guess [google/woff2](https://github.com/google/woff2) works.
Sorry to hijack the thread :)

davelab6:
Closed via #181

davelab6:
@rsheeter 

> The tools (fontTools, woff2, etc) and data (fonts, nam files to say what codepoints go into a given subset) we rely on are all open source.

Which tools exactly do we use? 

| format | tool |
|----|----|
| EOT | [sfntly](https://github.com/googlei18n/sfntly) |
| SVG | [Batik ttf2svg](https://xmlgraphics.apache.org/batik/tools/font-converter.html) |
| WOFF | [bramstein/sfnt2woff-zopfli](https://github.com/bramstein/sfnt2woff-zopfli) |
| WOFF2 | [google/woff2](https://github.com/google/woff2) |

davelab6:
@tomByrer This came up in https://github.com/googlei18n/sfntly/issues/49#issuecomment-179781613 - the way to use sfntly yourself to make EOTs is

    $ java -jar sfnttool.jar -e -x input.ttf output.eot

PR #150, opened 2015-11-13
Update Work Sans description

weiweihuanghuang:
Removed Hairline separate family reference


davelab6:
Thank you @weiweihuanghuang

PR #176, opened 2016-01-11
Update some nam files and add a few new ones.

rsheeter:


tools/encodings/khmer_unique-glyphs.nam
```diff
@@ -187,7 +187,6 @@
 0x17D9 uni17D9
 0x17DA uni17DA
 0x17DB uni17DB
-0x17DD uni17DD
```
davelab6: This regresses https://github.com/google/fonts/pull/50

tools/encodings/arabic_unique-glyphs.nam
```diff
@@ -79,74 +79,29 @@
 0x0670 # ARABIC LETTER SUPERSCRIPT ALEF
 0x0671 # ARABIC LETTER ALEF WASLA
 0x0679 # ARABIC LETTER TTEH
-0x067A # ARABIC LETTER TTEHEH
```
davelab6: @khaledhosny I'm curious what you think about which unicode characters should be included in the default arabic subset, and what the effect of not including these characters is? :)

davelab6: It seems that on investigating this a bit more, the larger set that has been here was stale, and this PR simply updates the file to reflect what has been in use since march.

khaledhosny: The removed characters seem to be for small and minority languages, so probably OK to drop from the default subset.

tools/encodings/arabic_unique-glyphs.nam
```diff
@@ -291,27 +198,6 @@
...
-0xFBE1 # ARABIC LETTER KIRGHIZ OE FINAL FORM
-0xFBE2 # ARABIC LETTER KIRGHIZ YU ISOLATED FORM
-0xFBE3 # ARABIC LETTER KIRGHIZ YU FINAL FORM
-0xFBE4 # ARABIC LETTER E ISOLATED FORM
-0xFBE5 # ARABIC LETTER E FINAL FORM
-0xFBE6 # ARABIC LETTER E INITIAL FORM
-0xFBE7 # ARABIC LETTER E MEDIAL FORM
 0xFBE8 # ARABIC LETTER UIGHUR KAZAKH KIRGHIZ ALEF MAKSURA INITIAL FORM
 0xFBE9 # ARABIC LETTER UIGHUR KAZAKH KIRGHIZ ALEF MAKSURA MEDIAL FORM
 0xFBFC # ARABIC LETTER FARSI YEH ISOLATED FORM
```
khaledhosny: I’d drop all Arabic presentation forms completely. If the font uses them in GSUB then `pyftsubset` (assuming that is what is used here) will keep them any way, if they are not used (like in Amiri) then you save file size. The only character that should be kept in this block is U+FDFA.

davelab6: What is fdfa for?

khaledhosny: [ﷺ](https://en.wiktionary.org/wiki/%EF%B7%BA). Unlike the rest of the Arabic Presentation Forms blocks, this is not really a presentation form it is a symbol of its own and can’t (or shouldn’t but people may tray) to be done as a ligature. I’m not sure how wide spread its use on the web, though, Arabic Wikipedia for example uses [an image](https://ar.wikipedia.org/wiki/%D9%82%D8%A7%D9%84%D8%A8:%D8%B5%D9%84%D9%89_%D8%A7%D9%84%D9%84%D9%87_%D8%B9%D9%84%D9%8A%D9%87_%D9%88%D8%B3%D9%84%D9%85) probably because most system fonts have an ugly symbol for it, not what people are used to/expecting.

davelab6: That remains, so we're good :) https://github.com/google/fonts/blob/master/tools/encodings/arabic_unique-glyphs.nam#L210

khaledhosny: Right, I wanted to drop *the rest of the block*, though.

khaledhosny: I.e. drop `0xFB50`–`0xFDF2`, and `0xFDFC`–`0xFEFC`.

davelab6: @roozbehp says, 

> Contrary to what one may expect, the usage of the characters from the Arabic Presentation Form blocks is relatively common on the web. That's why we included them.

So, I'm okay leave this as is, and simply not drawing these characters in new fonts :)

khaledhosny: Not really convinced they are that popular, but I won’t argue more since I was only concerned about Amiri file size and now I remembered that I strip them away in the “web” version anyway.

khaledhosny: Not really convinced they are that popular, but I won’t argue more since I was only concerned about Amiri file size and now I remembered that I strip them away in the “web” version anyway.

davelab6: Cool - not including them is fine

PR #191, opened 2016-01-26
Correct order of subsets in METADATA.pb

sapics:
I guess that subsets are sorted in alphabetical order except 'menu'.
This pr correct order of subsets.


davelab6:
@rsheeter what is the canonical order?

rsheeter:
I don't recall offhand that there is one; we typically view them as a set.
Having the order be consistent in textproto form sounds reasonable.

davelab6:
Alphabetical order sounds good.

sapics:
Thank you for your approval!

PR #201, opened 2016-01-28
tools/add_font.py Include *-ext subsets if any support

davelab6:
>> @davelab6: I also noticed that it doesn't automatically add latin-ext subsets when I would do so; since latin-ext, greek-ext, and cyrillic-ext are 'catch all' subsets defined much larger than most fonts that have characters outside their regular subsets, perhaps a special case for this subset would be good, where if any characters in these 3 -ext subsets are found, the METADATA enables this subset. 
>
> @rsheeter: Today add_font uses 50% coverage as the mark for supporting a subset. We could do per-subset limits but that sounds potentially surprising. What if we just said 25% is the mark or something like that?

I checked and I've used a very low bar for `-ext` subsets. 

    python add_font.py ../apache/calligraffitti/
    Calligraffitti-Regular.ttf latin 212/219
    Calligraffitti-Regular.ttf latin-ext 7/810 

Also this supersedes https://github.com/google/fonts/pull/200 that follows up on #191 to prevent further regressions

### DIFF
--- a/tools/add_font.py
+++ b/tools/add_font.py
@@ -58,8 +58,16 @@ def _MakeMetadata(fontdir):
   """
   file_family_style_weights = _FileFamilyStyleWeights(fontdir)
 
+  # include subset keys when the first font in a family has at least 
+  # 50% of the characters in a subset
   first_file = file_family_style_weights[0].file
   subsets = ['menu'] + [s[0] for s in fonts.SubsetsInFont(first_file, 50)]
+  # make an exception for -ext subsets, include them if any support exists
+  for s in fonts.SubsetsInFont(first_file, 0):
+      if s[0].endswith("-ext"):
+          subsets.append(s[0])
+  # sort and uniq the subsets list
+  subsets = sorted(set(subsets))
 
   font_license = fonts.LicenseFromPath(fontdir)
 
--- a/tools/what_subsets.py
+++ b/tools/what_subsets.py
@@ -18,8 +18,8 @@ flags.DEFINE_integer('min_pct', 0,
 def main(argv):
   for arg in argv[1:]:
     for (subset, available, total) in fonts.SubsetsInFont(arg, FLAGS.min_pct):
-      print '%s %s %d/%d' % (os.path.basename(arg), subset, available, total)
-
+      percent = 100 * float(available) / float(total)
+      print '%s %s %03d/%d %02d%%' % (os.path.basename(arg), subset, available, total, percent)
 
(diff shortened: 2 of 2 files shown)


davelab6:
I set the `latin-ext` threshold very low ([here](https://github.com/google/fonts/pull/201/files#diff-c99bb7185b8b7be6550ac3f6220cd51cR66) because the most common `latin-ext` length (in 110 files) is 8 glyphs:

```
$ python what_subsets.py ../*/*/*tf | grep latin-ext > latin-ext-counts.txt
$ cat latin-ext-counts.txt | awk '{print $3}' | cut -d\/ -f1 | sort | uniq -c | sort | tail ;
  31 140
  32 016
  34 006
  41 151
  42 138
  43 090
  48 208
  58 001
  82 007
 110 008
```

On a percentage basis, the most common percentage coverages start with 0% and increment from there until 30%...

```
$ cat latin-ext-counts.txt | awk '{print $4 $1}'  | sort | uniq -c
 329 00%
 105 01%
  85 02%
  31 03%
  26 04%
   6 05%
  24 06%
   2 07%
   6 08%
   2 09%
  71 10%
  87 11%
  27 12%
  58 13%
  36 14%
  29 15%
  78 16%
 183 17%
  58 18%
  29 19%
  38 20%
  71 21%
  16 22%
  11 23%
  11 24%
  66 25%
  10 26%
  13 27%
  31 28%
   2 29%
   8 30%
   2 33%
   4 36%
   1 38%
   3 40%
   1 41%
  13 42%
   8 45%
   1 46%
   2 49%
   4 50%
   1 51%
   1 61%
   2 65%
  20 67%
   1 73%
$
```

And the many fonts with eg 17% or 25% are from larger families, which as larger projects are bigger along both style and charset axes

```
$ cat latin-ext-counts.txt | awk '{print $4 $1}' | cut -d\- -f1 | sort | uniq -c | sort | tail
  11 17%TitilliumWeb
  12 21%Roboto
  12 28%SourceSansPro
  13 20%OpenSans
  14 21%AlegreyaSans
  14 21%AlegreyaSansSC
  18 13%Exo2
  18 16%Exo
  18 25%Raleway
  18 28%Kanit
```

rsheeter:
I will add two independent flags, --min_pct and --min_pct_ext to add_font and what_subsets. For add_font --min_pct will default to 50 and --min_pct_ext to 10. This allows the desired lower bar for -ext subsets.

davelab6:
:8ball:

Issue #202, opened 2016-01-28
tools/sanity_check.py naming bug


davelab6:
There appears to be a bug when detecting the family names where the spaces in the family name are dropped - for example 

$ python sanity_check.py ~/fonts/ofl/hindcolombo/
FAIL: Hind Colombo normal/300 'name' family[0] expected Hind Colombo, got Hind Colombo Light (/Users/dcrossland/fonts/ofl/hindcolombo)
FAIL: Hind Colombo normal/500 'name' family[0] expected Hind Colombo, got Hind Colombo Medium (/Users/dcrossland/fonts/ofl/hindcolombo)
FAIL: Hind Colombo normal/600 'name' family[0] expected Hind Colombo, got Hind Colombo SemiBold (/Users/dcrossland/fonts/ofl/hindcolombo)

The 400 400i 700 and 700i styles of a large family have only 4 records related to naming. Eg:

    <namerecord nameID="1" platformID="3" platEncID="1" langID="0x409">
      Hind Colombo
    </namerecord>
    <namerecord nameID="2" platformID="3" platEncID="1" langID="0x409">
      Bold
    </namerecord>
    <namerecord nameID="4" platformID="3" platEncID="1" langID="0x409">
      Hind Colombo Bold
    <namerecord nameID="6" platformID="3" platEncID="1" langID="0x409">
      HindColombo-Bold
    </namerecord>

The other 14 of the 18 possible styles have 6 records related to naming, which present a full 9 weight/18 style family as 8 families; one 4-style family as above for 2 weights, and then 7 pairs of Regular/Italic families in IDs 1 and 2. These are then reunited into a single family with the additional name IDs 16 + 17.

    <namerecord nameID="1" platformID="3" platEncID="1" langID="0x409">
      Hind Colombo Light
    </namerecord>
    <namerecord nameID="2" platformID="3" platEncID="1" langID="0x409">
      Regular
    </namerecord>
    <namerecord nameID="4" platformID="3" platEncID="1" langID="0x409">
      Hind Colombo Light
    </namerecord>
    <namerecord nameID="6" platformID="3" platEncID="1" langID="0x409">
      HindColombo-Light
    </namerecord>
    <namerecord nameID="16" platformID="3" platEncID="1" langID="0x409">
      Hind Colombo
    </namerecord>
    <namerecord nameID="17" platformID="3" platEncID="1" langID="0x409">
      Light
    </namerecord>

[...]

PR #203, opened 2016-01-28
tools/sanity_check.py Resolve nameid 16 bug (#202)

davelab6:
@rsheeter please review carefully, I think this is likely not good.

### DIFF
--- a/tools/sanity_check.py
+++ b/tools/sanity_check.py
@@ -1,5 +1,3 @@
-
-
 import collections
 import contextlib
 import os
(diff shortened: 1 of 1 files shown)


tools/sanity_check.py
```diff
@@ -349,6 +347,16 @@ def _CheckFontNameValues(path, name, font, ttf):
...
     actuals = fonts.ExtractNames(ttf, name_id)
     for (idx, actual) in enumerate(actuals):
+      # print "actual is", actual
+      if name_id == 1:
+          # print "name_id is", name_id
+          if expected != actual:
+            # print 'no match for', expected, actual
+            name_id = 16
+            actual = fonts.ExtractNames(ttf, name_id)
+            actual = actual[0]
```
davelab6: Maybe this should be

    actual = actual[idx]

davelab6: Note to self from chat with Rod:

> Ideally I'd like to retain the mechanism of building expectations and comparing with actual. I think, though I haven't actually tried it out, that allows you to leave the verification check alone and just make the construction of expectations more elaborate. That's probably (hopefully?) simpler than plugging more logic into the loop of the actual values.

PR #204, opened 2016-01-28
Add names of unicode characters to nam files

nyshadhr9:


davelab6:
Nice! Do you have a script to generate these?

PR #208, opened 2016-02-04
Update tools. Most notably allow lower threshold for detecting -ext subsets

rsheeter:
This is intended as an alternate implementation of #201


davelab6:
@felipesanches please review

tools/add_font.py
```diff
@@ -9,9 +9,19 @@
 import fonts_public_pb2 as fonts_pb2
 from google.protobuf import text_format
 from google.apputils import app
+import gflags as flags
 
 from util import google_fonts as fonts
 
+FLAGS = flags.FLAGS
```
davelab6: Curiousity nit: Should this variable follow google_var_style_naming?

rsheeter:
No, due to vagaries of when various things were written flags use C++ style naming.

davelab6:
:8ball:

felipesanches:
I read the patch, but @davelab6 merged it faster :-)

davelab6:
@felipesanches right, I mean to read it to inform font bakery development, not to approve :)

felipesanches:
yep! Thanks :-D

rsheeter:
Out of curiosity, how does that play out? Do we implement things twice
(once for font bakery and once here) or do they do different things or are
we sharing code somehow?

davelab6:
> how does that play out? 

Chaotically. :)

The scripts here and in fb are like a venn diagram. Your sanity_check.py is duplicative of some parts of fontbakery-check-*.py and add_font.py becamge duplicative of fontbakery-build-metadata.py when Felipe updated it to handle protocolbuffers instead of json.

I wrote https://github.com/googlefonts/fontbakery/blob/master/BRIEF.md to provide a contextual overview of how I am using the 2 sets of scripts together, and I expect to add some noto-tools scripts (like fontdiff.py to go with ttx files + meld)

davelab6:
> how does that play out? 

Chaotically. :)

> Do we implement things twice (once for font bakery and once here)

Not too much :) 

The scripts here and in fb are like a venn diagram. Your sanity_check.py is duplicative of some parts of fontbakery-check-*.py and add_font.py becamge duplicative of fontbakery-build-metadata.py when Felipe updated it to handle protocolbuffers instead of json.

> or do they do different things 

Not really. They are basically the same. The difference is that fontbakery has been quite monolithic and rather unreliable, and your scripts have been more unix-philosophy and reliable. Felipe is assisting me to make fontbakery more like your scripts.

> or are we sharing code somehow?

I wrote https://github.com/googlefonts/fontbakery/blob/master/BRIEF.md to provide a contextual overview of how I am using the 2 sets of scripts together, and I expect to add some noto-tools scripts (like fontdiff.py to go with ttx files + meld)

rsheeter:
That was an interesting read, ty.

PR #209, opened 2016-02-07
tools/encodings/cyrillic*.nam Remove old afiiXXXX labels

davelab6:
Per https://github.com/adobe-type-tools/agl-aglfn/blob/master/aglfn.txt#L79-L80 these are outdated


rsheeter:
LGTM.

Though ... I wonder if we should instead update cyrillic to the style @nyshadhr9 used in 375f6207?

davelab6:
Sure; please replace https://github.com/google/fonts/blob/master/tools/encodings/namelist.py :)

PR #276, opened 2016-06-19
Update v.01 to GWF Cyrillic encoding 

alexeiva:
Removing two characters that provide only partial support for Kazakh language, which renders them unnecessary:
-0x04B0 uni04B0
-0x04B1 uni04B1
Kazakh language is supported in Adobe Cyrillic-2, and the suggested Google Cyrillic Plus.

List of glyphs that are missing for full Kazakh support:
ә,Ә,ғ,Ғ,қ,Қ,ң,Ң,ө,Ө,ұ,Ұ,ү,Ү,һ,Һ
and their respective Unicode values: uni04D9, uni04D8, uni0493, uni0492, uni049B, uni049A, uni04A3, uni04A2, uni04E9, uni04E8, uni04B1, uni04B0, uni04AF, uni04AE, uni04BB, uni04BA

Adding new symbols: ₴(Hryvnia sign, 20B4), ₽(Ruble sign, 20BD)


davelab6:
Removing is unlikely but the additions look good.

alexeiva:
It's fine to leave as is in all published fonts. Just a recommendation to skip these symbols in future fonts in the Basic set.

rsheeter:
Very happy to see the subsets under active discussion! 

Currently this directory reflects the production subsets so I suggest we put this into it's own dir (cyrillic-proposal or some such) and add a text or markdown file for the rationale. I saw the note in the comment but I think it's too easily lost there.

If there are characters you are recommending not be drawn (skipped in future fonts) lets list those in separate file(s). We would are unlikely to remove them from the production subsets even if we chose not to draw them for future fonts.

alexeiva:
@rsheeter 
Thank you for the feedback. As you suggested I have created a separate folder for the new 2016 encodings in a fork here:
https://github.com/alexeiva/fonts/tree/master/tools/encodings/GF%202016%20Glyph%20Sets

We will make a new pull request when we are ready.

PR #283, opened 2016-06-26
small schwa

Tural:
0x0259  ə LATIN SMALL LETTER SCHWA


alexeiva:
@Tural thank you for the proposal. Small scwha is already included in [GF Latin Plus glyph list ](https://github.com/google/fonts/blob/master/tools/encodings/GF%202016%20Glyph%20Sets/GF-latin-plus_unique-glyphs.nam#L285)

PR #303, opened 2016-07-03
fontbakery check results in markdown syntax

felipesanches:


davelab6:
Thanks Felipe! However, this isn't the best place to store that information so I'm closing this PR :)

PR #361, opened 2016-08-23
Rename METADATA.pb.txt -> METADATA.pb in Inconsolata Font

sapics:
Just a renaming of the file.


m4rc1e:
Was solved in #436 but thanks :-)

PR #376, opened 2016-09-13
lobster: added V1.007

m4rc1e:
hey @davelab6,
Apologies about the previous commit. Still had the old font in repo.

Cheers,
Marc


m4rc1e:
Didn't realise Alexei was modifying the Cyrillic. I will submit a pr once I have this in place

PR #378, opened 2016-09-17
General typographical issues. 

akkatracker:
Mostly its --> it's


davelab6:
@akkatracker thanks for this!

PR #429, opened 2016-11-02
francoisone: v2.000 added

m4rc1e:
Thanks @ThomasJockin


m4rc1e:
Reopening. I've dropped the RFN because it is a Vernon family.

Issue #433, opened 2016-11-03
Neuton font 300 weight displays colon as middle dot


nickcernis:
Typing a `:` character results in a middle dot with Neuton's 300 weight:

<img width="809" alt="neuton_-_google_fonts" src="https://cloud.githubusercontent.com/assets/647669/19959083/db8ad152-a19b-11e6-85c7-8355ed5dce17.png">

I found a report of this in a repo at https://github.com/vladpopa33/googlefontdirectory/issues/215 dating back to 2013 but couldn't see anything in the official repo here.

davelab6:
@nickcernis thanks for reporting!

Will be fixed when PR #440 is pushed to production :)

![screen shot 2016-12-06 at 13 02 44](https://cloud.githubusercontent.com/assets/261579/20937291/4fc21c06-bbb4-11e6-8613-f6174a089ae3.png)

davelab6:
Confirmed fixed :) 
    
<img width="279" alt="screen shot 2016-12-12 at 14 51 57" src="https://cloud.githubusercontent.com/assets/261579/21114273/8cb894ee-c07a-11e6-8cd0-ff9cee051c73.png">

m4rc1e:
Closing.

Can confirm
![screen shot 2017-06-14 at 12 05 03](https://user-images.githubusercontent.com/7525512/27129319-b8981a70-50f9-11e7-92a1-bd181291c865.png)

http://jsbin.com/harehenowa/edit?html,output

Issue #347, opened 2016-08-04
Neuton font displays non-breaking space much wider than a normal space


lukechilds:
If you add have a mix of `&nbsp;` and normal space characters in your markup the content looks really bad. Many WYSIWYG editors will add `&nbsp;` in strange places and they can also end up in content when it's pasted in from other sources.

Demo https://jsfiddle.net/tg67akud/2/

m4rc1e:
I have replicated the bug and fixed it.

1st line: nbsp
2nd line: space
![screen shot 2016-08-11 at 23 34 51](https://cloud.githubusercontent.com/assets/7525512/17605752/3e93f402-601c-11e6-9865-8678fbdeddd0.png)

This has been checked and implemented for each weight. 

Update will be rolled out soon.

Thank you for reporting this.

lukechilds:
@m4rc1e Any idea when this update will be coming out? The jsfiddle I posted still has the issue.

m4rc1e:
Hey @lukechilds,

My repository with the fix is [here](https://github.com/m4rc1e/Neuton). Count this as an unofficial release for the time being. 

We're in the process of upgrading popular fonts and fixing bugs at the moment.

Cheers,
Marc

lukechilds:
Thanks 👍

m4rc1e:
Closing, LGTM.

![screen shot 2017-06-14 at 12 24 20](https://user-images.githubusercontent.com/7525512/27130033-715a4590-50fc-11e7-8bc0-85ffa8801de3.png)

https://jsfiddle.net/tg67akud/2/

PR #440, opened 2016-11-08
neuton: v1.560 added

m4rc1e:
solves #347 #433


davelab6:
@m4rc1e latin-ext regression fixed in ceb3f7b

PR #442, opened 2016-11-08
oswaldheavy: v4.000 added

m4rc1e:
Heavy fonts need to be in their own separate family.


davelab6:

PR #446, opened 2016-11-08
telex: v1.100 added

### LINKED ISSUE #114: Telex has wrong family name
Is has the  family name "Telex-regular", but should just be called "Telex"?

m4rc1e:
RFN has been included. This update will fix #114


davelab6:
@m4rc1e note I fixed a METADATA latin-ext regressoin in 51125d6

PR #448, opened 2016-11-08
bevan: v2.000 added

m4rc1e:
Vernon repo so removed RFN.


davelab6:
@m4rc1e note, fa27864 fixed METADATA filename key update

PR #451, opened 2016-11-09
sharetech: v1.100 added

### LINKED ISSUE #71: Share Tech Mono render problem
Hello,

There is a bug in redering text written as "Share Tech Mono". the combination fi results in a dot instead of the expected letters fi.

m4rc1e:
Fixes #71 in sharetech.

RFN has been moved back onto first line.

Cheers,
Marc


davelab6:
@m4rc1e note that the designer key regressed b88f4fc

PR #453, opened 2016-11-09
nunitoheavy: 4.000 added

m4rc1e:
New weights.

Not included in Nunito family because the api cannot serve Heavy fonts. They have to be seperated into a new family.


davelab6:

davelab6:
Closing as the VF is on the horizon.

PR #470, opened 2016-11-21
metrophobic: v1.100 added

### LINKED ISSUE #407: Metrophobic does not contain space characters
Metrophobic does not contain space characters, which is trivial to fix, and quite urgent :)

m4rc1e:
This will fix #407


m4rc1e:
Need to add Description file.

davelab6:
@m4rc1e this is an existing family so there is a DESC file already; I touched up the line breaks (274a18e) but its all good for now. Thanks!

PR #472, opened 2016-11-24
How to use "Google Font Locally" 

jesusmtzarvizu:


davelab6:
@jesusmtzarvizu Thanks for this :) However, this advice is not ideal. I will leave this PR open and someday follow up with better advice; probably a link to https://github.com/majodev/google-webfonts-helper would suffice :)

m4rc1e:
It would also be handy to squash the commits and write a clearer commit message.

PR #484, opened 2016-12-02
firasanshairline: v4.203 added

m4rc1e:


davelab6:
@m4rc1e I'm going to hold this open because I think we should deal with the super-thin weights in a specific way that I will work with you on soon

davelab6:
Similar to https://github.com/google/fonts/pull/489#issuecomment-264860643 closing this for now

PR #499, opened 2016-12-03
manuale: v0.075 added

m4rc1e:


davelab6:
@m4rc1e Omnibus families should be marked as such in designer key (fixed 5081d27)

PR #501, opened 2016-12-03
asap: METADATA.py: ttfs: fixed style names

m4rc1e:


davelab6:
@m4rc1e thanks! There was a regression on `date-added` (24d8cf1 fixed) we should look at improving the metadata generator next year :) (cc @rsheeter & @felipesanches)

PR #502, opened 2016-12-04
sansita: v1.006 added

m4rc1e:
Last Omnibus font added.

This family is missing the Euro glyph though.


thlinard:
Still not available in production?

PR #517, opened 2016-12-06
Update size for download all google fonts

chezsick:
the size 250Mb is false, because m4rc1e have added the fonts, me i download as tar.gz (not zip).


davelab6:
@chezsick thanks!

PR #518, opened 2016-12-06
firasansextracondensed: v4.203 added

m4rc1e:
I haven't uploaded the ridiculous hairline sub family :-)

Here's what the font menu looks like
screen shot 2016-12-06 at 14 09 32

Only merge if happy.


davelab6:
@m4rc1e winner

PR #520, opened 2016-12-06
Revert "firacode: v3.206 added"

davelab6:
Reverts google/fonts#516

Decided against this 

> Fira Code is Fira Mono 3.206 with less Line Space (1.0) – does not include programming ligatures 

So, not needed, as line-space can be controlled on the web


wesalvaro:
Hi!

Can we revert this revert based on your latest comment in #1460?

Thanks!

tonsky:
@wesalvaro no, this is the wrong Fira Code and should stay reverted. The right one is in #1460

PR #553, opened 2016-12-16
alice: v2.000 added

m4rc1e:
Thanks @alexeiva! this was a solid release, no tweaks needed.

Metrics were also spot on 
screen shot 2016-12-16 at 13 32 43


davelab6:
@alexeiva 👍

PR #555, opened 2016-12-19
Adding a Vietnamese guide

alexeiva:


davelab6:
@alexeiva Awesome! 

(I don't know why Github is seeing your commits from different authors and causing the googlebot to nag, but I guess that you may have 2 different git configs in GUI and CLI tools that don't quite match up :) 

screen shot 2016-12-20 at 14 07 12

alexeiva:
@davelab6 you are right. I have fixed my username in cli recently, so this wasn't my doppelgänger.

PR #598, opened 2017-01-23
merriweather: v2.000 added

m4rc1e:
thanks @EbenSorkin


EbenSorkin:
I think all the relevant merging has happened but the downloadable source isn’t up to date yet.

Does that look correct to you?

Similarly fonte squirrel etc have older files

-e.

m4rc1e:
Hey Eben,

This may take some time to happen. It's not realtime I'm afraid.

PR #617, opened 2017-01-30
encodesansexpanded: v2.000 added

m4rc1e:
Same DESCRIPTION.en_us.html check.


m4rc1e:
@davelab6 Ok, according to FB my family names are incorrect the on Semi Weights:

`expected 'Encode Sans Semi Condensed' but got 'Encode Sans SemiCondensed'`

Keep all the Encode Sans prs open and I will update the DESCRIPTION.en_us.html in each of them.

PR #625, opened 2017-02-03
encodesans: v2.000 added

m4rc1e:


impallari:
I Mark, I've just noticed that Encode (Normal) Thin instance has custom parameter "weightClass: 100".
Any particular reason for it?

impallari:
Oh, I see you just fixed it. Forget about it. Cheers.

PR #635, opened 2017-02-06
cormorantinfant: v3.303 added.

m4rc1e:
Fixed residual damage from Glyphs rescaling bug.

@CatharsisFonts please review these prs. Than you for your patience as well.


CatharsisFonts:
Hi Marc,

sorry for not getting around to reviewing this PR earlier. I visually inspected all glyphs in the Infant Regular font and didn't see any additional issues. The only thing that caught my eye was the fact that the /iigrave-cy does not correspond to the Bulgarian style prevalent in the Infant's Cyrillics, but I believe that was due to Alexei actively discouraging the presence of non-Bulgarian glyphs in the Bulgarian localization.

Thanks for your work! Hope the update will be live soon.

Issue #624, opened 2017-02-02
How to use tools/encodings


graphicore:
I'm working on getting language coverage information for the encodings in tools/encodings. I'm going to ask some questions here in order to make sure I use the provided files correctly.

The [README.md in GF 2016 Glyph Sets](https://github.com/google/fonts/blob/master/tools/encodings/GF%202016%20Glyph%20Sets/README.md) mentions "Core" encodings for Latin and Cyrillic, but it seems there are no files `encodings/GF 2016 Glyph Sets/GF-latin-core_unique-glyphs.nam` and `encodings/GF 2016 Glyph Sets/Cyrillic/GF-cyrillic-core_unique-glyphs.nam`, are the "Core" encodings `encodings/latin_unique-glyphs.nam` and `encodings/cyrillic_unique-glyphs.nam`?

---

[README.md](https://github.com/google/fonts/blob/master/tools/encodings/README.md) says:
> Typically all the Unicodes in each file are in each font. If the fonts go beyond that list, those additional characters will not be available to Fonts API end users.

So, I guess the Fonts API uses these files directly for subsetting.

But from the data I see, some files must be dependent on other files, e.g. [latin-ext_unique-glyphs.nam](https://github.com/google/fonts/blob/master/tools/encodings/latin-ext_unique-glyphs.nam) does not include the chars of [latin_unique-glyphs.nam](https://github.com/google/fonts/blob/master/tools/encodings/latin_unique-glyphs.nam)
Thus, implicitly, `latin-ext_unique-glyphs` includes the charcodes of `latin_unique-glyphs`, this is not documented, as far as I can see.

Is there anywhere a description what exactly the subsetter does for each encoding provided by the Fonts API?

Similarly, my unicode cldr based coverage script produces output like this:

```
googleFontsTools/encodings/sinhala_unique-glyphs.nam
lax language detection: true
language: Sinhala 72 % having: 74 needed: 103 missing: 29 ("-" U+002d,"–" U+2013,"—" U+2014,"," U+002c,";" U+003b,":" U+003a,"!" U+0021,"?" U+003f,"." U+002e,"…" U+2026 … and 19 more) laxSkipped: 3
```

([full output for all nam files](https://gist.github.com/graphicore/00a79eded3553a64ad409559246d853c))

In this case, some very common punctuation marks are missing, it seems impossible that the fonts API would ship a font without e.g. the exclamation mark. Is there a rule, like that all encodings include `latin_unique-glyphs`?

[...]

PR #642, opened 2017-02-08
Improve and extend Namelist file format documentation.

graphicore:
This is a living PR for discussion. **Don't merge it** until we agree that it is good.


davelab6:
Great work! Please assign to me when ready to merge

davelab6:
@graphicore is this ready to merge?

graphicore:
> Great work! Please assign to me when ready to merge

Uh, I missed this. Thanks.

I'm rather happy with this. The biggest three issues for me right now are:

1. The  `tools/encodings/README.md` has a passage at the top "**TODO: clarify:** Typically all the Unicodes in each file are in each font." That sentence is hard for me to understand, it seems that it's just not true. Do I miss something?
2. Since you seem to agree with what I did to the Namelist format, especially the `#$ include` headers. I should remove all parts in `tools/encodings/README.md` that say it's a "proposal". Also, I'd clean it up once more in general, in case I see something more
3. Wrong includes could also be fixed later, but in general, are the dependencies via the  `#$ include` headers declared correctly? At the moment, the newly added Greek encodings don't have these headers, so that should be fixed. And the "optional" encodings don't have these headers. The latter usually don't have unicode codepoints, so it's not important for the CLDR/language detection related stuff. 

Other things can be done in a subsequent PR, like tasks derrived from  #660 and #624

graphicore:
I'll tackle the remarks above later today, as far as I can.

graphicore:
What I did:

1. just removed the "**TODO: clarify:**". This must have made sense before, so maybe I'm reading it wrong. Feel free to comment/ask for a change on this one before you merge.
2. I had another go on that README.md file, I think it's OK now.
3. I added include rules to Greek core, plus and expert. The rest can be dealt with later, I'll have some questions anyways.

graphicore:
I just removed  including `GF-latin-plus_unique-glyphs.nam` into `GF-greek-core.nam` because of: https://github.com/graphicore/specimenTools/issues/19

Note that we should probably create `GF-latin-core.nam` because that should be included by `GF-greek-core.nam`.

I'd rather do this and other changes on the encodings in a separate PR which would however depend on this PR. Or I can just add to this PR. What should I do?

graphicore:
closed by merging #685

davelab6:
Thanks!

> "TODO: clarify: Typically all the Unicodes in each file are in each font." That sentence is hard for me to understand, it seems that it's just not true. Do I miss something?

I guess that there are a couple of things to take from that. 

1. all the unicode characters specified in a glyph set definition should be available as drawn glyphs in each font that meets that definition, and

2. all the fonts in a family should have the same unicode encoded glyphs. There is a FB issue to track that.

graphicore:
> 1. all the unicode characters specified in a glyph set definition should be available as drawn glyphs in each font that meets that definition, and

I think the key here is saying **should**, we'll have to update fonts from time to time to keep the glyph set definitions and the fonts in sync.

@davelab6, somehow related, could you have look at this please? https://github.com/graphicore/specimenTools/issues/20

PR #655, opened 2017-02-14
Add Amstelvar to Early Access

davelab6:
I will keep this open until the time is right to merge it :)


davelab6:
cc @dberlow @djrrb

davelab6:
Google Fonts is ready to start hosting this and Decovar in fonts.google.com/earlyaccess as soon as you are; I put up https://github.com/google/fonts/pull/655 in anticipation of this. 

- [ ] I named the binary `Amstelvar-Roman.ttv` - is this acceptable for you?

- [ ] I made a very brief `DESCRIPTION.en_us.html` file, let me know any further text :)

dberlow:
My only concern is that early access without commitment to completion makes the certainty of final access questionable. Keeping in mind that this font still has a lot of work to do, with only the ASCII having all of the axes, and no agreement to go further.

So maybe you should call it Early Axes;?

Otherwise I'll leave it to you.

PR #656, opened 2017-02-14
Add Decovar to Early Access

davelab6:
I will keep this open until the time is right to merge it, too :)


davelab6:
cc @dberlow @djrrb

dberlow:
Renaming back to DecovarAlpha is fine with me to mark the step into Early Access and to distinguish it from possible later Decorvars.

No documents yet depend on the current name. 

No other script or software requiring the current name are as important as the family name you want. 

My 2¢, 

David

davelab6:
@dberlin thanks! will track upstream

PR #663, opened 2017-02-16
muli: v2.001 added.

m4rc1e:
Fixes MM compatibility.


m4rc1e:
'H' wasn't removing the overlaps correctly on the bar. This would only be noticeable at 200pt+.

screen shot 2017-02-16 at 12 37 19

davelab6:
cc @fonthausen

PR #669, opened 2017-02-17
postnobillsjaffna: v1.220 added

m4rc1e:


davelab6:
@mooniak to prevent the copyright string becoming very long, it would be good in the next release to use the 'project authors' style notice

Issue #411, opened 2016-10-23
Google Open Sans is not supporting Greek extended fonts


minosarp:
I am developing a webpage using greek polytonic. I have imported from google fonts the code for the **greek extended fonts**. However the polytonic letters were displayed in default _Arial, Sans-Serif_. 
I had to download and install different fonts from Google and using Microsoft Word I discovered that only **Ubuntu** supports extended fonts. **Open Sans** and **Roboto** do not support although Google says that they do.
![fonts](https://cloud.githubusercontent.com/assets/23015387/19627780/e433a028-9957-11e6-877c-98f7342b873c.jpg)
Is anyone have gone through the same issue or am I doing something wrong?

davelab6:
Please could you show me a test page? Perhaps with www.jsbin.com or similar :)

minosarp:
Dear Dave,

Thanks for the reply and I am sorry for the delay. Below the test page.

http://jsbin.com/viquwe/2/edit?html,css,output

2016-10-24 18:15 GMT+03:00 Dave Crossland <notifications@github.com>:

thlinard:
Hi Dave,

I confirm,
subsets: "greek-ext"
should be removed from the METADATA.pb of:
opensans
opensanscondensed
roboto
robotocondensed
robotomono
robotoslab
caudex (incomplete support)

You can test on https://fonts.google.com/?subset=greek-ext with this sentence:
Τὴ γλῶσσα μοῦ ἔδωσαν ἑλληνικὴ
τὸ σπίτι φτωχικὸ στὶς ἀμμουδιὲς τοῦ Ὁμήρου.
Μονάχη ἔγνοια ἡ γλῶσσα μου στὶς ἀμμουδιὲς τοῦ Ὁμήρου.

ἀπὸ τὸ Ἄξιον ἐστί
τοῦ Ὀδυσσέα Ἐλύτη

davelab6:
@minosarp thanks! I reviewed and I'm not sure I see the problem... when I use 100 or 900 then I expect that - since Arial doesn't have these - that the weight difference will make the fallback obvious... but I don't see that :)

![screen shot 2017-02-05 at 22 26 28](https://cloud.githubusercontent.com/assets/261579/22633736/21edad04-ebf2-11e6-93cf-7930a95f3a35.png)

[...]

PR #691, opened 2017-03-06
greek-ext Subsets list corrected

thlinard:
See issue #411


davelab6:
Please could you make 2 PRs, one to remove subsets that are being served despite the fonts not having the glyphs, and one to add subsets that can be served but are currently inaccessible?

PR #694, opened 2017-03-07
play: v2.101 added.

m4rc1e:
Readded legacy glyphs which were missing from sprint.

Thanks @alexeiva


m4rc1e:
Let's keep these open until you folks get Travis up and running.

davelab6:
@m4rc1e what are the VFB and glyphs files doing here? :)

PR #695, opened 2017-03-07
greek-ext Subsets list corrected

thlinard:
See issue #411


thlinard:
@davelab6 tell me if this is what you expected, I still have a lot of git commands to discover…

PR #700, opened 2017-03-09
exo: v1.500 added

m4rc1e:
Thanks @NDISCOVER and @rbmntjs


davelab6:
Ah, the fsSelection bold bit should be 1, is 0, in Exo-BoldItalic.ttf

PR #717, opened 2017-03-23
rename "GF 2016 Glyph Sets/" folder => "GF Glyph Sets/"

felipesanches:
at encodings/README.md to match the recent folder rename by @graphicore


davelab6:
Thanks!

PR #728, opened 2017-03-31
vollkorn: v4.014 added

m4rc1e:
New Vollkorn has more weights. This has affected the family weight distribution. New bold appears much lighter because it has become the Black in the latest release.

screen shot 2017-03-31 at 15 24 28
*Left: new Right: old*

People have complained about such issues in the past.


davelab6:
Curious Medium and Extrabold are missing. Do you know why?

I'm ok to shift the weight to make it better

m4rc1e:
@davelab6 Previous release only had:
- Regular
- Bold
- Italic
- Bold Italic
https://fonts.google.com/specimen/Vollkorn

New release has:
- Regular
- SemiBold
- Bold
- Black
- Italic
- SemiBold Italic
- Bold Italic
- Black Italic

I don't think there ever was a Medium.

The weight shift may be ok. The Regular is still the same weight as the old. The new Black weight is the same weight as the previous Bold. The new SemiBold and Bold are in between values.

Imo, the old Bold is more like an ExtraBold, so squeezing in two middle weights makes a better transition.

davelab6:
I'm suggesting even more intermediate weights be generated

davelab6:
Superceded by https://github.com/google/fonts/pull/1011

PR #764, opened 2017-05-03
hotfix-arimo: v1.231 added

m4rc1e:


Qantas94Heavy:
For posterity's sake, if anybody else is trying to work out what this changes, this only sets the fsType to 0. (That basically means the font is allowed to be installed permanently.)

PR #804, opened 2017-05-03
hotfix-abeezee: v1.002 added

m4rc1e:


davelab6:
@m4rc1e lines 7 and 16 of METADATA are unwanted changes, and I suspect the file has the same unwanted 'name' table change. Please check and update :)

m4rc1e:
@davelab6 The font names are good.

screen shot 2017-05-08 at 15 44 59

The problem is in add_font.py. It relies on a family name parser which splits the [camelcase](https://github.com/google/fonts/blob/master/tools/util/google_fonts.py#L349-L364).

I'll fix it by hand for the time being. The family name should be taken from the nametable imo.

davelab6:
OK cool - FB has an exceptions list for the unusual (un camel casable) families we have.

m4rc1e:
@davelab6, I'll do this for other camelcase names which should be ignored, e.g VT323.

PR #882, opened 2017-05-05
hotfix-crimsontext: v0.121 added

m4rc1e:


davelab6:
@m4rc1e note v0.121 is off our spec, we need v1.000+ as a v0.x will cause issues somewhere

PR #883, opened 2017-05-05
hotfix-eczar: v1.103 added

m4rc1e:


davelab6:
@MrBrezina we have to make this hotfix, it is just to fsSelection bitmask, macstyle, and name table :)

m4rc1e:
@MrBrezina description has to match the one currently being served here https://fonts.google.com/specimen/Eczar

MrBrezina:
@m4rc1e makes sense, thanks and thanks for letting me know @davelab6

MrBrezina:
Upon inspecting the TTFs: you are changing our copyright strings (!), trademark (as far as I know, Vaibhav does not have any trademark registered), and removed our vendor code in ID3. I don’t have any particular problem with any of these except the last one (correct code is ` RST`). I have transferred some of these to our source. What is the logic behind the name table changes? Can I change `fsSelection` from AFDKO?

PR #909, opened 2017-05-05
hotfix-oxygen: v0.201 added

m4rc1e:


hroncok:
@m4rc1e Where this did come from originally? I have a bit trouble to find the sources. Thanks

m4rc1e:
@hroncok All the prs which have a "hotfix-" prefix were a batch job to fix the fsSelection, macStyle, usWeigthclass font attribs. Nothing visual has changed. We did it so the fonts will work properly in Google Docs.

hroncok:
Thanks.

PR #166, opened 2015-12-22
SourceCodePro - updated with latest version

futekov:
- Updated the SourceCodePro family with the [latest version](https://github.com/adobe-fonts/source-code-pro/releases/tag/2.010R-ro/1.030R-it) that includes support for cyrillic and greek, and also has italic styles
- Updated both metadata files to include the new language and style support
- Updated the readme with the [official one](http://www.adobe.com/products/type/font-information/source-code-pro-readme.html)

PR #924, opened 2017-05-05
hotfix-sofadione: v1.002 added

m4rc1e:


davelab6:
Also I suggest replacing with https://github.com/google/fonts/pull/166

PR #925, opened 2017-05-05
hotfix-sourcecodepro: v1.018 added

m4rc1e:


davelab6:
Also I suggest replacing with #166

PR #926, opened 2017-05-05
hotfix-sourcesanspro: v1.051 added

m4rc1e:


davelab6:
better to PR the latest update which is overdue :)

davelab6:
https://github.com/google/fonts/pull/950

PR #927, opened 2017-05-05
hotfix-tangerine: v1.301 added

m4rc1e:


onecero:
Hi,

I´m new here. I don't know if this is the place to write some ideas for the improvement of the Tangerine font.

The Tangerine font need some kerning and the letter 'C' need some adjustments.

PR #936, opened 2017-05-07
Revert "exo: v1.500 added"

davelab6:
@m4rc1e I want to revert google/fonts#700 until the fsSelection is fixed


davelab6:
@m4rc1e I think that the Exo 2 work by @rbmntjs hasn't yet been PR'd... but its ready to go from upstream :)

Issue #326, opened 2016-07-18
Update Mada to v1.3


khaledhosny:
https://github.com/khaledhosny/mada/releases/tag/v1.3

ericorruption:
I wanted to contribute on this but couldn't find the font folder on the project.

PR #938, opened 2017-05-08
hotfix-mada: v1.005 added

m4rc1e:
There never was a folder for this family.

The font's copyright string also included the trademark. I'm not sure if this is correct.


m4rc1e:
I'd like to point out that this exists, https://github.com/google/fonts/issues/326. 

The current version has a problem with the 'A'. The hotfix doesn't solve these issues.

screen shot 2017-05-08 at 17 07 22

davelab6:
@m4rc1e okay, I'll merge this now, but please stand up a PR for that because its super bad :)

PR #951, opened 2017-05-15
add_font.py: --update metadata option

m4rc1e:
Currently, when you run [add_font.py](https://github.com/google/fonts/blob/master/tools/add_font.py) It will add placeholders for 'designer', 'category' and 'date added' fields. 
https://github.com/google/fonts/blob/master/tools/add_font.py#L82-L83

I've been upgrading a lot of families so I need to retain these from the previous release. To do this, I've used the metadata api end point.

### DIFF
--- a/tools/add_font.py
+++ b/tools/add_font.py
@@ -29,6 +29,8 @@ import os
 import re
 import sys
 import time
+import requests
+import json
 from argparse import ArgumentParser
 
 import fonts_public_pb2 as fonts_pb2
@@ -48,6 +50,30 @@ flags.DEFINE_integer('min_pct_ext', 10,
                      ' for a -ext subset.')
 
 
+def family_url(name):
+  url_prefix = 'http://fonts.google.com/specimen/'
+  dl_name = name.replace(' ', '+')
+  return url_prefix + dl_name
+
+
+def family_exists(name):
+  """Check if a font family exists on fonts.google.com"""
+  url = family_url(name)
+  request = requests.get(url)
+  if request.status_code == 200:
+      return True
+  return False
+
+
+def _GetFamilyMetaData(name):
+  """Get family information from the frontend GF api"""
+  api_prefix = 'http://fonts.google.com/metadata/fonts/'
+  if family_exists(name):
+    url = api_prefix + name.replace(' ', '+')
+    request = requests.get(url)
+
+    return json.loads(request.text[5:])
+  return None
 
 
 def _FileFamilyStyleWeights(fontdir):
@@ -84,7 +110,7 @@ def _FileFamilyStyleWeights(fontdir):
   return result
 
 
-def _MakeMetadata(fontdir):
+def _MakeMetadata(fontdir, update=False):
   """Builds a dictionary matching a METADATA.pb file.
 
   Args:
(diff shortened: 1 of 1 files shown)


davelab6:
This should probably use the [public developer API](https://developers.google.com/fonts/docs/developer_api) rather than the private fonts.google.com directory API

tools/add_font.py
```diff
@@ -132,21 +198,28 @@ def _WriteTextFile(filename, text):
 
 
 
-def main(argv):
-  if len(argv) != 2:
-    sys.exit('One argument, a directory containing a font family')
-  fontdir = argv[1]
+def main(args=None):
+  parser = ArgumentParser(description=__doc__)
```
rsheeter: Please use gflags, as shown in sanity_check.py, to ensure we can sync the internal copy gracefully.

tools/add_font.py
```diff
@@ -23,6 +55,30 @@
...
+  url = family_url(name)
+  request = requests.get(url)
+  if request.status_code == 200:
+      return True
+  return False
+
+
+def _GetFamilyMetaData(name):
+  """Get family information from the frontend GF api"""
+  api_prefix = 'http://fonts.google.com/metadata/fonts/'
```
rsheeter: This API is subject to change without warning, is there somewhere else you could get the information?

davelab6: I believe it can all be obtained from the public developer API, yes, eg

```json
  {
   "kind": "webfonts#webfont",
   "family": "ABeeZee",
   "category": "sans-serif",
   "variants": [
    "regular",
    "italic"
   ],
   "subsets": [
    "latin"
   ],
   "version": "v10",
   "lastModified": "2017-05-11",
   "files": {
    "regular": "http://fonts.gstatic.com/s/abeezee/v10/mE5BOuZKGln_Ex0uYKpIaw.ttf",
    "italic": "http://fonts.gstatic.com/s/abeezee/v10/kpplLynmYgP0YtlJA3atRw.ttf"
   }
```

davelab6: Ah `lastModified` != `dateAdded` - so, no.

rsheeter: Actually ... can't you just keep the previous values for things like designer if there is an existing file? - that would avoid network and seems simpler.

m4rc1e: @rsheeter Many families exist on fonts.google.com but don't have folders in this repository. 

hopefully, I will have all the missing fonts added to the repository this week. 

For the time being, I think we should leave this open. Once the repo and fonts.google.com are in sync, I can use your file method.

I'll also use gflags. Thank you for the review.

rsheeter: Anything public *should* exist in this repository. We should build out this tool based on that assumption and fix it if it's violated. If the tool can avoid network hits, such as by retaining the values from the pre-existing file that makes it much easier to test and maintain so I would prefer to avoid ever updating it to have network dependencies.

@davelab6 Are you aware of this drift? Should we embark on an effort to bring repo & fonts.google.com back into sync?

davelab6: I chatted to @rsheeter offline to confirm that this PR is part of such embarkation by @m4rc1e :)

davelab6:
@m4rc1e the old fontbakery METADATA.json generator python script would look for an existing file in the same dir path as it took as input, and if found, would (a) take values from that where they should stay the same and (b) wrote out to `METADATA.pb.new`; when we moved from JSON to PB, a fresh `add_font.py` was written which didn't have this. However, @felipesanches did upgrade the old too from JSON to PB. 

But I can no longer see this in the fontbakery source tree; however it was at some point here: https://github.com/googlefonts/fontbakery/blob/7145ce54594fb8b8de8f65e94bae5be76bdf07e3/bakery_cli/scripts/genmetadata.py

m4rc1e:
@rsheeter I've found the time to implement the changes requested in your review.

You're right, the implementation is far simpler.

Thanks again.

m4rc1e:
ping @davelab6 

Could we get this reviewed?

rsheeter:
Sorry, missed this in a cloud of family updates. Will take a look.

tools/add_font.py
```diff
@@ -21,6 +50,9 @@
 flags.DEFINE_integer('min_pct_ext', 10,
                      'What percentage of subset codepoints have to be supported'
                      ' for a -ext subset.')
+flags.DEFINE_boolean('update', False,
```
rsheeter: WDYT about default True or deleting the flag entirely? - I'm thinking that usually if I point add_font at an existing directory with a METADATA.pb I don't want it to destroy information.

rsheeter: I'm going to merge because it looks fine, feel free to send a PR if you decide modifying the flag is worthwhile.

Issue #28, opened 2015-05-15
Josefin Sans regular appears to be italic


baferret:
Not sure if I'm doing anything wrong, but I installed Josefin Sans Regular on Windows 7, and both GIMP and Paint show it as italic, not regular.
Nice font BTW!

jomoos:
Can confirm, appears to be italic on Inkscape in Fedora 20.

digitalmoksha:
From what I can tell, it seems that inside the ttf container, the font's type is set incorrectly - it's set to another type in the family, which is causing a duplicate (and overwrites the previous font when installed).  The actual typography inside the container is correct (based on it's preview), but it's name conflicts with others in the same family.

I'm seeing this on a Mac using Font Book.  I see it with two fonts:

* JosephineSans-Italic tries to install as Regular
* JosephineSans-SemiBoldItalic tries to installs as SemiBold

taip:
Confirmed on Mac. Regular Italic, SemiBold Italic, and Thin Italic have name conflicts; Light Italic and Bold Italic are fine.

davelab6:
Thanks for investigating :)

Behold666:
Any way to fix this? I need the regular weight for a project.

davelab6:
I have an update from the designer, which will solve this

Behold666:
I found that downloading the font from Font Squirrel gave me uncorrupted files. Does the archive that Google Fonts hosts now work?

4ll4n:
Any updates on this? Tried Font Squirrel but got the same problem.

insanelyus:
Had the same problem. http://1001freefonts.com version works fine.

omurphy27:
Having the same problem on Mac OS. 

Switched to Font Squirrel which works fine for me.

untangledjake:
Just wanted to report that I'm also experiencing this problem on a Windows machine in latest versions of Chrome and Firefox. This is on a site that's been up for awhile but the problem seems to have begun in the last week or so.

digitalmoksha:
@davelab6 , you mentioned you had an update from the designer - is there any way that those updates can get checked in?  Would be great to just get this closed out...

just-trey:
I would like to see this corrected as well seems to be an issue on all devices and browsers

squarecandy:
+1

squarecandy:
[...]

PR #1045, opened 2017-06-13
josefinsans: v2.000 added.

m4rc1e:
I am pleased to finally have this out the bag. Thanks @ThomasJockin.

The missing glyphs issue in the upstream repo, has been solved,
https://github.com/ThomasJockin/JosefinSansFont-master/issues/4

This update is a big one. I looked at the fonts in GFR, it looks pretty sweet.

josefin


m4rc1e:
People may complain because the update is so drastic... however, it looks way better.

indolering:
This also contains a bugfix (#28) that has been floating around for 2 years!

m4rc1e:
@indolering A test or evidence would be nice to see?

davelab6:
@m4rc1e yeah I think @indolering is right on this. Josefin has needed the last mile for ages

indolering:
Nope, @m4rc1e is correct, his pull doesn't have that fix.  @m4rc1e could you cherry pick?

davelab6:
@indolering which commit do you want to cherry pick?

indolering:
@davelab6 IDK :man_shrugging:.

thezelja:
@m4rc1e is there any option to revert to v1.0 
or use Josefine Sans v1.0 from fonts.google.com?

davelab6:
You could apply negative letter-spacing to reduce the shift in spacing;
however the new spacing is considered better for readability.

thezelja:
@davelab6 thanx

redranger:
My feedback is that v.1 looked much better. I will need to remove this font as an option on my websites as it is not nearly as nice. Sorry! Any chance we can get v.1 back, or at least have it as an option?

indolering:
@redranger You could host the font yourself and lock it to a specific version.

redranger:
Where would I get the original? Also, what was wrong with the original? I'd like to see the original to compare them again, as I need to decide for definite whether to remove this font, keep the new version, or host version 1 as you suggest. Thanks.

inspire22:
Yeah this broke a bunch of pages on my website.  The new one is MUCH bolder.  How to download the old one for older pages?  Is it even possible since it has been updated at the source?

redranger:
I think there is a bug with the updated font. The tops of some letters go above the top line of their container.

m4rc1e:
@redranger screenshot or jsbin please.

Also, which OS and browser?

redranger:
I'm using Firefox v.55.0.3 on Windows 10. I also tested it in the latest version of Chrome (v.61). See the attached screenshot that highlights the problem. I had a website where the tops of the letters was being cropped off because of this problem.
firefoxwindows10

benjaminparis:
Hi, on Mac OSX 10.11.6, no matter what the browser is (Chrome, Safari, Firefox) the regular font (400) font is bolder and bigger than the bold one (700) as you can see. 
capture d ecran 2017-09-14 a 15 48 18

Here is the file loaded by Google Fonts API : 
```
@font-face {
  font-family: 'Josefin Sans';
  font-style: normal;
  font-weight: 400;
  src: local('Josefin Sans Regular'), local('JosefinSans-Regular'), url(https://fonts.gstatic.com/s/josefinsans/v11/xgzbb53t8j-Mo-vYa23n5hampu5_7CjHW5spxoeN3Vs.woff2) format('woff2');
  unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2212, U+2215;
}
```
which isn't the regular one. Any idea ? 
Thanks

m4rc1e:
@redranger do you have the family installed locally?

On Win 10, FF v55, I couldn't replicate this.

screen shot 2017-09-14 at 15 05 27

@benjaminparis duplicate of #1193

redranger:
I don't have it installed locally. Go to that page in Chrome, then right click on the word Bold (don't select it, just right click on it), then select Inspect. You should see the tops of the letters 'l' and 'd' sticking out.

m4rc1e:
> I'm using Firefox v.55.0.3 on Windows 10

> Go to that page in Chrome, then right click on the word Bold

Using Browser stack so can't right click. Highlighting LGTM on Win 10 Chrome 61

screen shot 2017-09-14 at 15 35 52

redranger:
I mentioned both Firefox and Chrome if you look back. I'm guessing it will probably be an issue with most modern browsers. I'll leave the issue with you. I've removed the font as an option for users on my websites.

thlinard:
Hi @m4rc1e 

I'm unable to reproduce the bug on Chrome 61 for Mac or Safari 10.1 (macOS 10.12.6), but it's visible on Firefox 55.0.3 for Mac, specially for the Bold weight.

The hheaAscent and sTypoAscender value of 750 doesn't seem right for the Bold (something like 789 would probably be better, or 793 in a "Webfont Strategy").

m4rc1e:
@thlinard I can't change this. Vertical metrics visually have to be the same as the previous edition. By changing these values, users will get reflow issues.

davelab6:
If its actually broken we should fix it.

thlinard:
Yes, I understand. But have you see the documents I put in [Google Fonts Discussions](https://groups.google.com/d/msg/googlefonts-discuss/v4RUY3u2HW4/UuvYpIpIBAAJ)? Users already get reflow issues anyway with the version 2.00.

davelab6:
Not vertical reflow though, which is more disruptive

m4rc1e:
Apologies @thlinard, I mean't vertical reflow. We care about this because some DTP apps/text editors make it difficult/impossible to change the 'leading'.

indolering:
@m4rc1e  Maybe it would be better to create a Josefin neue?  Google Fonts doesn't let one stick to a specific version, so these changes have lots of potentially breaking consequences.  That and my inability to get Mozilla to update the repos readme make me think it would be better to just fork 😦 .

fastfasterfastest:
I see the same thing in Edge.
You see it when the text is selected AND the element does not have the focus.  E.g., double-click the word Regular to select it.  The selection has then a blue background and the text is white and the top of the letters are not "sticking out". Then click in the address bar to remove the focus and the selection changes to a gray background and the text is black and the top of the letters are "sticking out".  Another way do see it is to use F12 Developer tools and highlight the span element (with the text Regular) in the DOM Explorer - then you also see the top of the letters "sticking out".

> Go to that page in Chrome, then right click on the word Bold (don't select it, just right click on it), then select Inspect.

The key is NOT to select the word but to right-click on it and select Inspect.

jkyeung:
Hi, I am having the exact same issue as #1193, namely that the current regular weight (400) is too bold for my use case and light (300) is too thin. To me, it's a big gap that would be nicely filled with 350. Would it be possible to include it as a static instance for download?

PR #1060, opened 2017-06-16
baloobhaijaan: v1.443 added.

m4rc1e:
Too many Arabic glyphs are missing for add_font.py to find the codepage.

This needs reviewing before we release, @girish-dalvi


davelab6:
Does add font detect all the other Arabic fonts as such?

PR #1072, opened 2017-07-03
Early Access variable beta fonts update.

m4rc1e:
Following families have been updated by regenerating the projects with the latest version of fontmake, v1.2.7:

- Archivo
- Asap
- Asap Condensed
- Cabin
- Faustina
- Manuale
- Overpass
- Podkova

The following families were not possible to generate with the updated fontmake:
- Dosis
- Encode Sans
- Josefin Sans
- Maven Pro
- Muli
- Nunito Sans
- Nunito
- Oswald
- Raleway
- Rokkitt
- Sansita
- Unna
- Yanone Kaffee Satz

These families were possible to generate with previous version. We should look into this


thlinard:
Hi Marc,
What version of Raleway?

m4rc1e:
@thlinard Raleway V4.025. 

I still think this has to do with fontmake and not the fonts themselves. I guess generation is far stricter than it was a few months back. I'll need to check whether this statement is correct.

rbmntjs:
It should also work with Exo, which was largely built with 4 masters – but I believe the pull request was reverted, so not sure if that’s even in the repo now.

davelab6:
Closing as superceded by https://github.com/google/fonts/pull/1264

PR #1094, opened 2017-07-23
Fix incorrectly-classified Protobuf files

Alhadis:
GitHub currently classifies this repository as a PureBasic codebase, due to the abundance of [`METADATA.pb` files](https://github.com/google/fonts/search?l=purebasic):

Figure 1

Setting a `linguist-language` attribute will fix both miscategorised files and syntax highlighting [currently missing](https://github.com/google/fonts/blob/480630de320cc119f8fc516ea6a55fbfd5232db0/apache/aclonica/METADATA.pb) from every `METADATA.pb` file:

~~~console
λ Google-Fonts (fix-misclassification): ~/Forks/GitHub-Linguist/bin/linguist .
55.31%  Protocol Buffer
44.69%  HTML
λ Google-Fonts (fix-misclassification):
~~~

More information about overriding auto-detected languages [can be found here](https://github.com/github/linguist#overrides). =)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.pb linguist-language=protobuf


Alhadis:
CLA signed!

davelab6:
Awesome, thanks very much :)

Alhadis:
Consider it a small token of gratitude for Noto. =) 

Erm, I couldn't help but notice the syntax highlighting of the `METADATA.pb` files is still missing... those *are* written in the usual Protocol Buffer language, yes?

If so, the site's highlighting grammar for Protobuf might need some fixing... :/

davelab6:
Yes, AFAIK they are the usual Protocol Buffer language.

rsheeter:
They are text format protobuf instances. I imagine the highlighting for protobuf may be limited to .proto files that define the structure of data.

Alhadis:
Do you know of any formal specification that describes this text format?

rsheeter:
Unfortunately I do not. The best reference I can point you to is https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.text_format + the corresponding code.

davelab6:
Maybe YAML is close enough?

Alhadis:
I'm afraid we can only use one highlighting grammar per file-format. 😞  Although if the text format of Protocol Buffers has a definitive file extension (and enough in-the-wild usage), we could register it as a separate filetype.

From the looks of it, though, it doesn't appear that the format is "established" enough to merit formal classification.

davelab6:
We do not use any other protocol buffers format so only using yaml is fine

Alhadis:
Thanks, but it would mean replacing the [main highlighting](https://github.com/pannous/tensor-caffe/blob/master/caffe.proto) used for the format. Which I can tell is much more widespread.

[This repository](https://github.com/pannous/tensor-caffe) mentions the name `prototxt`, which seems to be [the same format](https://github.com/pannous/tensor-caffe/blob/master/lenet.prototxt)... I think it might be worth adding this as a separate filetype. =)

PR #1107, opened 2017-07-28
tools/encodings/armenian_unique-glyphs.nam added.

m4rc1e:
File adds subsetting support for Armenian codepage. File generated with tools/namelist.py
using apache/mardoto/Mardoto-Regular.ttf as a source


davelab6:
@m4rc1e does the namelist.py result for Noto Sans and Serif Armenian have any deltas on this?

m4rc1e:
@davelab6 I will look into this sometime this week

m4rc1e:
Hey Dave,

Mardoto is missing the following glyphs when compared to [Noto Armenian](https://www.google.com/get/noto/)
```
0xFB15 ﬕ ARMENIAN SMALL LIGATURE MEN INI
0xFB14 ﬔ ARMENIAN SMALL LIGATURE MEN ECH
0xFB17 ﬗ ARMENIAN SMALL LIGATURE MEN XEH
0xFB16 ﬖ ARMENIAN SMALL LIGATURE VEW NOW
0xFB13 ﬓ ARMENIAN SMALL LIGATURE MEN NOW
```

We need this investigated before we can merge #1106 

cc @vahanhovh

davelab6:
@m4rc1e I don't see the lack of these ligatures as blocking release, but we can't release until the Armenian subset is supported by the API. I'm not sure when that will happen.

m4rc1e:
@davelab6 my experience with Armenian is limited so take my opinion with a pinch of salt.

I'm assuming these are probably used for ancient manuscripts, hence why they were not included in Mardoto.

If they do appear frequently in text, they're definitely worth having. 

Hrant or John H probably have the answer.

PR #1109, opened 2017-07-31
Add librebarcode families

davelab6:
Thanks @graphicore :)


davelab6:
Thanks @graphicore :)

PR #1112, opened 2017-08-02
Add Sebastian Kosch to Amiri designers

khaledhosny:
The Latin is based on his Crimson Text.


davelab6:
thank you @khaledhosny :)

PR #1113, opened 2017-08-02
Add Aref Ruqaa designers

khaledhosny:
The Arabic is done by Abdullah Aref and Khaled Hosny, while the Latin is based on Hermann Zapf’s AMS Euler.


davelab6:
@khaledhosny thanks for these, attributing designers correctly is important :)

Issue #185, opened 2016-01-25
Update Nobile


davelab6:
https://www.google.com/fonts/specimen/Nobile has only 4 styles but 6 are available

m4rc1e:
Hey @davelab,

You're correct, Vern has an unreleased v1.200 in his [repo](https://github.com/vernnobile/NobileFont).

I will get this fixed and prepped.

Cheers,
Marc

m4rc1e:
Nobile V1.200 is still a major WIP.

![screen shot 2016-09-02 at 13 20 48](https://cloud.githubusercontent.com/assets/7525512/18202429/23dda806-7110-11e6-9dbb-80fe41cf54ca.png)
![screen shot 2016-09-02 at 13 20 40](https://cloud.githubusercontent.com/assets/7525512/18202428/23dc0b54-7110-11e6-89f3-2807c54b816b.png)
![screen shot 2016-09-02 at 13 20 26](https://cloud.githubusercontent.com/assets/7525512/18202427/23d9cca4-7110-11e6-807f-1968e1349881.png)

If we decide to extend the character set. I will prepare the sources properly. Until then, I think it best we stay with the v1.000 release.

Cheers,
Marc

laerm0:
Did some more research on this and it looks like there are a few more things to be fixed (in v1 release as well):

- [ ] accents cut off
- [ ] broken/misordered composites 

i.e.
![screen region 2017-08-02 at 15 16 36](https://user-images.githubusercontent.com/10780161/28890618-a22029c8-7795-11e7-98df-a0d2085b3bb9.png)
![screen region 2017-08-02 at 15 16 49](https://user-images.githubusercontent.com/10780161/28890617-a21dd6b4-7795-11e7-8450-72e745fb0662.png)

davelab6:
So, to fix family naming and ensure they can be installed locally on Windows and Mac, can be done with hotfixing the TTFs currently in the API. 

If the 2 new styles available on Github need design work to ship, that's fine to be queued up, but we need the 'bug' problems fixed themselves.

laerm0:
Okay, I'll work on the aforementioned fixable issues.

laerm0:
Fonts install and work well now.

Windows
![screen region 2017-08-03 at 11 52 01](https://user-images.githubusercontent.com/10780161/28931764-23932a28-7845-11e7-98a9-6ed19dbe263e.png)

OS X
[...]

PR #1118, opened 2017-08-03
ofl/nobile: fixed names and OS/2 values

laerm0:
Fixed names and OS/2 values to get good installs in Win and Mac (#185).


m4rc1e:
@laerm0 What tooling did you use to do this? and is it repeatable?

Issue #1128, opened 2017-08-08
Irish Growler


thlinard:
Shouldn't Irish Growler be removed from the repo? It was removed from the API.

PR #1133, opened 2017-08-09
delete: apache/irishgrowler

davelab6:
Fixes https://github.com/google/fonts/issues/1128


sapics:
I would like to keep these files for consistency in my project.
It would be enough to mention in DESCRIPTION.en_us.html like Duplicated fonts, isn't it?

thlinard:
Irish Growler doesn't exist anymore in the API.

sapics:
@thlinard 

If this git follows the API, it makes sense.
Previous my comment was just my two cent, I will just revert this commit in my project after this is merged :)

rsheeter:
This git should be viewed as upstream of the API, not following. Either way deleting this makes sense.

PR #1145, opened 2017-08-21
Update and add Barcode fonts

davelab6:
thanks @graphicore :)


zealousappex:
I'm not able to get Barcode 39 without text to work, is this a issue see here:
ttps://www.w3schools.com/code/tryit.asp?filename=G4J9BICFPWN9

PR #1202, opened 2017-09-08
vollkornsc: v4.015 added.

m4rc1e:
The sister family to https://fonts.google.com/specimen/Vollkorn

@davelab6 no post processing was needed. The SC family is generated directly from the .glyphs sources, https://github.com/FAlthausen/Vollkorn-Typeface

@FAlthausen apologies for the delay on this. I've submitted an issue to FB to catch these in the future, https://github.com/googlefonts/fontbakery/issues/1568


FAlthausen:
Thanks!

PR #1221, opened 2017-09-24
removing U+02C9 from Latin Plus. Fixes #1088

### LINKED ISSUE #1088: GF Latin Pro: Remove /firsttonechinese
Pinyin is generated using /macroncomb in Glyphs.app. If /firsttonechinese should be used instead, what about the other tones?

alexeiva:
See #1088


davelab6:
Thanks!

PR #1230, opened 2017-09-27
Update Noto Sans Tamil to latest version

nyshadhr9:


davelab6:
The files need an OFL.txt to go with them :) This was committed directly to master, so closing, and will add the license also directly

PR #1237, opened 2017-10-04
Fix mode of non-executable files

Alhadis:
After hard-linking these fonts, I noticed several of them were marked as executable. This PR is essentially a `chmod -x` for every blatantly non-executable file, such that running `find . -type f -perm +111` in the directory displays the following output:

~~~console
./tools/util/filter_lists.py
./tools/util/py_subsets.py
./tools/rangify.py
./tools/namelist.py
./tools/fonts_public_pb2.py
~~~


davelab6:
@Alhadis Thanks so much for this! A great little fix :) It coincided with another repo clean up, moving tools to its own repo, so your branch had conflicts... so I redid the clean up on both repos :) https://github.com/google/fonts/commit/c19f2d37a7466e9da7b64aca5413f0ad5d584c3b here and https://github.com/googlefonts/tools/commit/88dd969241c7181f9f1cfb120ddfa4d17809b760 there.

PR #1252, opened 2017-10-16
lemonada: v4.002 added.

m4rc1e:
Added @Gue3bara update from upstream, https://github.com/Gue3bara/Lemonada


m4rc1e:
Closing since #1952 is newer.

PR #1327, opened 2017-11-15
barlow to v1.104

jpt:
ofl/barlow, ofl/barlowcondensed, and ofl/barlowsemicondensed updated to v1.103 (which fixes a few minor issues that people have opened recently), and then v1.104, which fixed an issue with 103's TTF files :)


jpt:
Signed it!

jpt:
Weird, I guess I need to close this PR until jpt/barlow#19 is fixed -- I suspect it's a Glyphsapp issue as it wasn't a problem before, something is wrong with decomposing glyphs before applying the slant filter

PR #1329, opened 2017-11-16
barlow family to v1.107

jpt:
v1.107 includes a number of fixes (see jpt/barlow/releases), mostly for non-English glyphs


jpt:
Ok, one of these days I am going to submit a PR without someone opening an issue in my upstream right as I do so 😂 ... closing this one, will do a more thorough check before submitting another tomorrow

Issue #1305, opened 2017-11-06
Font Validation report


thlinard:
I tested all the fonts (minus the duplicate, i.e. Alef, EkMukta, JosefinSansStd, MrBedfort, MrsSaintDelafield, Siemreap, and TerminalDosis) from the GF repo in the Font Book application (macOS 10.12.6). Here the Font Validation report:

**Errors** (“14 serious errors were found. Do not use these fonts”)
Artifika-Regular.ttf

- [x] 	‘post’ table usability

Aubrey-Regular.ttf

- [x] 	‘post’ table usability

Brawler-Regular.ttf

- [x] 	‘post’ table usability

Molengo-Regular.ttf

- [x] 	System Validation

Sarabun-Bold.ttf

- [x] 	‘post’ table usability

Sarabun-BoldItalic.ttf

- [x] 	‘post’ table usability

Sarabun-Italic.ttf

- [x] 	‘post’ table usability

Simonetta-Black.ttf

- [x] 	‘post’ table usability

Simonetta-BlackItalic.ttf

- [x] 	‘post’ table usability

Simonetta-Italic.ttf

- [x] 	‘post’ table usability

Simonetta-Regular.ttf

- [x] 	‘post’ table usability

WireOne.ttf

- [x] 	‘post’ table usability

**Warnings** (“32 minor problems were found. Proceed with caution”)
Astloch-Bold.ttf

- [x] 	‘name’ table structure

BM-HANNA.ttf

- [x] 	‘name’ table structure

CaesarDressing-Regular.ttf

- [x] 	‘name’ table structure

Flavors-Regular.ttf

- [x] 	‘name’ table structure

FrederickatheGreat-Regular.ttf

- [x] 	‘name’ table structure

Kantumruy-Bold.ttf

- [x] 	‘name’ table structure

Kantumruy-Light.ttf

- [x] 	‘name’ table structure

Kantumruy-Regular.ttf

- [x] 	‘name’ table structure

LeagueScript.ttf

- [x] 	‘name’ table structure

Padauk-Bold.ttf

- [x] 	‘name’ table structure

Padauk-Regular.ttf

- [x] 	‘name’ table structure

SeoulHangang-Light.ttf

- [x] 	‘sfnt’ required tables
- [x] 	‘vmtx’ table usability

SeoulHangang-Medium.ttf

- [x] 	‘sfnt’ required tables
- [x] 	‘vmtx’ table usability

SeoulHangangCondensed-Bold.ttf

[...]

PR #1340, opened 2017-11-21
ofl/seoulnamsancondensed: v1.01 added. Fixes OS X validation error for kern table

laerm0:
Fixed kern table error, v1.01 added. Reported in #1305 .


davelab6:
There is no `Regular.ttf` which is required.

PR #1565, opened 2018-05-17
markazitextvfbeta: v1.000 added

m4rc1e:
Taken from the upstream repo https://github.com/BornaIz/markazitext, commit 89f9a2fd100e912e572c873c187a854bac4f0635

cc @nyshadhr9


m4rc1e:
I've regenerated the font using fontmake. It also has a hand edited STAT table. The variable font now works correctly in desktop applications.

screen shot 2018-05-30 at 09 42 28

PR #1594, opened 2018-06-06
markazitext: v1.000 added.

m4rc1e:
Taken from the upstream repo, https://github.com/BornaIz/markazitext

@davelab6 I've kept the previous markazitextvfbeta dir and added this family to a new dir markazitext. The family is no longer a beta imo. I can revert this if you want.

We also need to get Vf font meta generation sorted :-)
https://github.com/googlefonts/tools/issues/49


davelab6:
Thanks Marc!

PR #1600, opened 2018-06-20
comforta: test autohinted vf font added

m4rc1e:
Font has been autohinted using VTT 

I've made sure I compiled everything in VTT then hit ship fonts

@laerm0 is the above process still correct?

I can produce a before and after of the VF's origin instance against the static instance served on GF if people are interested?


davelab6:
> I can produce a before and after of the VF's origin instance against the static instance served on GF if people are interested?

Yes, this should be like any other update PR, except that instead of a new set of static fonts, it has a new set of VFs that are smaller in terms of number of files, but have named instances in the STAT table to match the previous static fonts. 

gf-regression images showing what changed, if anything, is important for merging :)

laerm0:
Yep, that's the process, @m4rc1e.

m4rc1e:
Ok cool,

desktop_windows_7_ie_9 0_
*Win7 IE9*

Lots of cap and x-height jumps. 


screen shot 2018-06-20 at 15 32 55

screen shot 2018-06-20 at 15 33 22

desktop_windows_7_ie_9 0_

screen shot 2018-06-20 at 15 36 48


screen shot 2018-06-20 at 15 38 26


GFR:
http://gf-regression.com/compare/6f9be13d-c86d-444e-b795-b01bdf8a23ad

m4rc1e:
STAT and avar are working perfectly. Debatable whether auto hinting is actually better.

m4rc1e:
@davelab6 as I said in my mail. I'd like to keep the other static ttfs until we're happy with the hinting.

m4rc1e:
@nyshadhr9 Thanks, I'll fix and repush.

m4rc1e:
@nyshadhr9 I've just pushed an update. Our fix script will now set these bits based on the default instance. Thanks for spotting this.

PR #1745, opened 2018-10-31
remove dead link to designers' G+ pages

keikoro:
There's a link in the README that results in a 404 because the file/folder it used to link to got removed at some point.

I was specifically looking for info on designers as the API doesn't provide that, and then was extra disappointed when I found out the advertised file didn't exist. (;

I submitted a CLA btw.


davelab6:
Merged as part of https://github.com/google/fonts/pull/1746

PR #1746, opened 2018-10-31
Improve metadata description

keikoro:
This builds on #1745, but is more a suggestion for a slight content improvement than a fix, so I thought it might be better to separate it out.

It adds a little more context to the description of `METADATA.pb` files for anyone looking to find out more about individual font designers.


davelab6:
@keikoro Thanks for this!

PR #1768, opened 2018-11-30
Encode Sans: v3.000 variable & split-variable fonts added – WIP

thundernixon:
This is a work-in-progress pull request for the variable font update to Encode Sans.

Upstream repo: https://github.com/thundernixon/Encode-Sans

Split variable fonts are placed into the five width-based directory of the `fonts/ofl` parent directory. The full variable font (with `wght` + `wdth` axes) is in the `encodesans/` directories.

## Known remaining to-do items
- [x] Explain / ask about remaining FAIL/ERROR/WARN fontbakery checks
- [x] Check that filenames and folder sorting is as expected (Ask)
- [x] Update version to 3.000 in Glyphs sources, rebuild
- [x] update `FONTLOG.txt`
- [x] update `METADATA.pb` for variable font
- [x] Apply autohinting to VFs
- [x] Test hinting in new statics
- [x] Check that font naming/sorting is how others expect in desktop apps
- [x] Methodical Red Arrows check. 
    - [x] Ask: are *any* inflected curves okay (e.g. in accent marks like `/tilde`?)
- [x] Extra proof kerning for refined glyphs (`/Germandbls`es, `/Oslash`es)
- [x] Do I need to fix mark-to-mark positioning? See docs/diffing-against-old-statics
- [x] Upgrade build scripts
    - [x] Make build scripts direct new fonts into `fonts/` directory
    - [x] Add fontbakery checking to static font builds
    - [x] ~~Cue font diffing checks in build scripts~~ (Marc is automating this)
    - [x] (maybe) streamline split-VF build
- [x] Update: Remove width-split VFs from everything but normal width, at Dave's request
- [x] fix overlap in `/oslash` for static fonts
- [x] ~~re-add spaces in width names, to match existing google fonts releases and URLs~~ invalid

## Questions to answer

- [x] Are marks really not stacking well (as shown by fontdiffenator)? These seem to work in Apple Pages. More details at [Encode Docs](https://github.com/thundernixon/Encode-Sans/tree/master/docs/11-diffing-against-old-statics)
  - **according to Marc, I had put in the diff arguments backwards – so, the marks have actually improved**
- [x] Are _any_ inflected curves okay (e.g. in accent marks like /tilde?)
  - **According to Micah, it should be fine. Inflected curves don't cause any big rendering issues; they can make it hard to draw good shapes (e.g. /S), but they don't _need_ extra points if the shapes are achieved without them.**
- [x] Ots-sanitize is tripping on `ERROR: GDEF: bad caret value format: 3`. Is this just a Variable Font issue?
  - Yes, see [MS Typography docs on this](https://docs.microsoft.com/en-us/typography/opentype/spec/gdef#caretvalue-format-3): "In variable fonts, CaretValueFormat3 must be used to reference variation data to adjust caret positions for different variation instances, if needed. In this case, CaretValueFormat3 specifies an offset to a VariationIndex table, which is a variant of the Device table used for variations."
  - There's an issue with OTS: https://github.com/khaledhosny/ots/issues/178
- [x] (This might just take some Googling on my part, but) How can I use the [autohint VF beta](https://groups.google.com/forum/#!msg/googlefonts-discuss/WJX1lrzcwVs/SIzaEvntAgAJ) tool on Windows?
  - **Partial answer: I will experiment with VTT autohint, where I can clean up problems that may occur**
- [x] Is [the current font-menu name ordering problem](https://github.com/thundernixon/Encode-Sans/blob/master/docs/10-testing-and-refining-in-work-apps/assets/2018-11-29-16-20-17.png) the same issue that Marc has already reported?
- [x] If the old source has "Reserved Font Name" in the copyright field, should I change that, or ignore the FontBakery check 103? (I currently have it changed, but I am unsure of this...)
  - Dave said to keep it.
- [x] Ots-sanitize is tripping on `ERROR: GDEF: bad caret value format: 3`. Is this just a Variable Font issue?
  - Yes, see [MS Typography docs on this](https://docs.microsoft.com/en-us/typography/opentype/spec/gdef#caretvalue-format-3): "In variable fonts, CaretValueFormat3 must be used to reference variation data to adjust caret positions for different variation instances, if needed. In this case, CaretValueFormat3 specifies an offset to a VariationIndex table, which is a variant of the Device table used for variations."
- [x] The style names were `Semi Condensed` and are now `SemiCondensed`, etc (with spaces removed to follow current FontBakery recommendations). Should the style names instead retain the spaces in the new versions? Will the Google Fonts specimen URLs change, or stay the same? This is probably a @davelab6 call.
  - @m4rc1e says to stick with names as they are currently on fonts.google.com
- [x] Is it possible to make VFs play better in word processor font menus? So far, I’ve only been getting results which are pretty terrible. For instance, the names tend to be sorted by weight more than by width, regardless of how I order these names or axes in the NAME and STAT tables. Worse, usually the lightest instance is ordered first, and there are seemingly-random other ordering complexities. Do these apps just suck at ordering fonts, or am I doing things incorrectly? 
  - No multi-axis fonts work very (e.g. Bahnschrift, Venn, Barlow, etc). So, it's probably partly a problem with apps.
- [x] It seems logical that a VF should have just its family name as nameID 1 – e.g. “Encode Sans” – and “Regular” set as nameID 2 for the default style, with that instance value flagged as “normal” in the STAT table. However, by default fontmake assigns the default master (Condensed Thin, in this case) as nameID 2. What is best, and why?
  - Answer from Marc: nameID 2 must be the default style, corresponding to the default axis values in the `fvar` table. This must be on a master of a font. In the case of Encode Sans, this means nameID 2 does need to be "Condensed Thin."

FontBakery checks along with all fonts at https://github.com/thundernixon/Encode-Sans/tree/master/fonts


davelab6:
@m4rc1e please could you help answer those questions? :)

davelab6:
Yeah would be good to make them available. I suggest the build script make
them from the TTFs using pyftfeafreeze

thundernixon:
Great, thanks for letting me know! I incorporated pyftfeafreeze into my build for signika, so it shouldn't take too long to adapt it for Encode Sans, as well.

thundernixon:
I've updated fonts and added SC versions. I believe these are ready for diffing checks!

thundernixon:
@m4rc1e you stated in the Signika PR that

> Before we pr here, we should make sure your work is merged into the original repo. If there isn't one, we should create one in googlefonts/

Pablo Impallari hasn't been online for quite awhile, so I can't merge Encode into the original repo, either. I'm guessing we should also make a new Encode Sans repo under the `googlefonts/` org, or leave it under my user account?

davelab6:
You can make a PR to https://github.com/googlefonts/Encode-Sans (freshly forked)

@impallari said he might rekindle his love of letters this year... fingers crossed :)

thundernixon:
My only real remaining uncertainty is whether the default style must be a master, or if it can be an interpolated instance. E.g. can it be "Encode Sans Regular," or must it be "Encode Sans Condensed Thin"?

thundernixon:
Updated the files, including proper naming as discussed with @m4rc1e today. It obviously needs a QA verification, but I think it's good to go.

davelab6:
Great stuff, thank you!

thundernixon:
The latest split VF builds still suffer from the ["Thin" issue in macOS pages](https://github.com/thundernixon/Encode-Sans/issues/8), but work better in MS Word than prior versions due to improved nameIDs 1 & 2.

thundernixon:
In fixing nameIDs 1 & 2, I caused a couple of other issues, which I found and corrected while I was in the process of correcting [ppem bit 3](https://github.com/googlefonts/fontbakery/issues/2338) in these fonts.

thundernixon:
- [ ] UPDATE: now that the Google Fonts API can support multi-axis variable fonts, I will need to change this PR into a PR of the Weight + Width VF, and close the PRs for VFs split along the width axis. I will likely do this in December or January.

PR #1796, opened 2018-12-29
Pt Root UI by Paratype. Initial commit.

paratype-fonts:


davelab6:
@felipesanches @m4rc1e it would be great to have Fontbakery run as CI on PRs like this

m4rc1e:
I've been working on a CI for GF since November. 

It will run FB, Diffbrowsers and Diffenator. It will return a zip containing all the reports and images. I ran a demo of it in November when I was state side.

paratype-fonts:
The corporate CLA for Paratype Inc. is signed

paratype-fonts:
Added GitHub user paratype-git (paratype@live.com) who did the original commit to Google group paratype-google-contributors.

paratype-fonts:
We prefer to keep the RFN restriction because of some issues we experienced in the past. Some of our free fonts were modified in a rather sloppy way and distributed under the original name. It led to us receiving negative feedback from the end users. We believe that Google Fonts service provides a pretty smart sub-setting capabilities that should be sufficient for the majority of users. If someone wants to do a more sophisticated modification, it shouldn't be a big problem for him/her to change the font name. In  fact, there is a true advantage in distributing modified fonts under unique names. It eliminates possible naming conflicts.

davelab6:
We may want to rename the fonts ourselves then. Will that work for you, or
do you want to work on giving Google, Adobe, etc permission to use the rfn?

davelab6:
The OFL is explicit that subsetting is a form of modification that requires
the RFN to be cared for, so yes a formal consent document is required, by
the OFL, to be in place. So if you want to go that route, you ought to put
agreements in place with the major obvious distributors; Google, Adobe,
Monotype, FontSquirel, etc.

I hope to do more to make this easier, but the easiest thing is to rely on
trademarks and not use the rfn :)

paratype-fonts:
Subsetting implemented by Google does not violate OFL. See https://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=ofl-faq_web#b4b7db21 (Question: 2.6): “…newer subsetting technologies may be able to subset in a way that allows users to effectively have access to the complete font…” It is exactly what Google Fonts' subsetting does. The users of Google Fonts indeed do have access to the complete font without any limitations – it’s just the delivery is done is a smart, optimized fashion.

paratype-fonts:
It seems that all checks have passed. When can we expect the font to be published?

Akiyamka:
Any news?  Many people are waiting for the appearance of this font in Google Fonts

paratype-fonts:
For those who are interested, the font is available at: https://www.paratype.com/fonts/pt/pt-root-ui

m4rc1e:
- Medium weight is missing nametable records 16 and 17. These are needed to ensure the font is selectable in MS Word
- All fonts are missing nametable records 13 and 14. These mention that the font is ofl licened.
- fsSelection bits are not set correctly.
- Are the sources available in an upstream git repo? if so, is there a url?

paratype-fonts:
The updated font files are in https://github.com/paratype-fonts/fonts/tree/master/ofl/ptrootui

1. nametable records 16,17 as well as 13,14 are updated
2. This font family doesn't have a standard set of styles (re, italic, bold, bold-italic). The available styles are: light, regular, medium, and bold. With this set of styles, setting fsSelection to 0x20 is not working anyway.

The fonts are ready to be deployed.

PR #1798, opened 2019-01-04
Update TRIVIA.md

ben45c:


ben45c:
Signed.

fitojb:
It would be great to have a more descriptive commit message than GitHub’s default of “Update <filename>”. The log is already hard to parse due to this.

eliheuer:
I like the spirit of this PR, thank you!

There is a need for an index of all know libre fonts. I'm in the planning stages of building something like this at libretype.org. It wouldn't be a site that hosts libre fonts and displays specimens such as what Google Fonts is doing, but would simply be an index of what exists in the libre type commons.

PR #1571, opened 2018-05-21
vollkorn: vollkornvf v4.104 added.

m4rc1e:
These two variable beta fonts replace the previous static ttfs.

Taken from the upstream repo, https://github.com/FAlthausen/Vollkorn-Typeface. PR in upstream is still open, https://github.com/FAlthausen/Vollkorn-Typeface/pull/24

Please note, I simply deleted the old ttfs and made no changes to the METADATA.pb file. our METADATA.pb generator needs some refactoring to work with variable fonts, https://github.com/googlefonts/tools/issues/49

PR #1825, opened 2019-01-25
/ofl/vollkorn Update VF metdata/desc

davelab6:
Way back in August 2018, I merged #1571 which added Vollkorn VFs, but the METADATA file was not updated. 

This PR updates the METADATA with one I just hand-crafted by inspecting the name  and fvar table via TTX XML.

@m4rc1e I'm going to merge this now as @andyjgf asked for an expedited fix, but my guess is that you might want to redo the VF generation on this (and add static files etc) as if it was a "January 2019 VF PR"


m4rc1e:
I'd hold off pushing these until I QA them. These fonts are unhinted btw.

PR #1837, opened 2019-02-04
Add updated and new versions of IBM Plex

thundernixon:
Changes documented at https://github.com/thundernixon/plex/tree/diff-QA/gfonts-qa, which includes diff gifs from `gftools qa --diff`. 

#### Sans
- There is a new "Text" weight
- The new fonts have many new glyphs, in particular, Greek language support
- The new fonts have much-improved placement of accents (see [marks_new.gif](gfonts-qa/currently-hosted-fonts/ibmplexsans/IBMPlexSans-SemiBold-diff_gifs/marks_new.gif))
- The new fonts have kerning between more pairs
- The new fonts make small refinements, e.g. the /gcaron uses a low-eared /g as a base to be more legible

#### Serif

- Similar readability refinements
- Added Text weight
- Glyphs appear to be slightly lighter in weight
- Only three new glyphs, rather than entire Greek alphabet as Sans has

#### Sans Condensed & Mono

- Almost no changes
- Added "Text" weight

#### Devanagari, Hebrew, Thai

- These members of the Plex family will be completely new to Google Fonts.

## Questions to answer before merge

- [x] I assume "Text" styles are important styles to include in the metadata, but if it will break things, I can remove these blocks. Should `METADATA.pb` include or exclude these weights?
  - I am setting the `weight` value as `450` in `METADATA.pb` because this is the weight IBM has specified for this style: [Plex weights, used to generate CSS releases](https://github.com/IBM/plex/blob/80846f5512975b75ff76c7c122159b2a910a5174/scripts/data/weights.js)

- [x] Should the `description` HTML be modified to link to new Devanagari, Thai, and Hebrew versions of Plex?
  - Plex has new versions. Should these be called out in the `description` HTML, or is it good to leave the description as-is? What about for these new families, in which the description is copied over from Plex Sans? I think it still works well as-is, but if it seems useful to update the descriptions, I can do so.

- [x] These have not yet been checked in Browserstack. @m4rc1e, do you want to do this, or would you want me to? (If you want me to do so, I need to get Browserstack credentials).


thundernixon:
(Adding `Priority 0` label based on conversation last week with @davelab6 – if this is inaccurate, feel free to modify the label).

m4rc1e:
> There is a new "Text" weight

It has always existed. Our api doesn't support this style so it was removed. Please remove it and regen the metadata.

> Devanagari, Hebrew, Thai

I think these families append the script name to its family name... I don't think we want this. Could you remove these families from this pr please? we can do them in a separate pr once we understand how to treat them.

Thanks and good work. I'll take a look myself tomorrow.

thundernixon:
Done and done. 

That's strange to me about the API not supporting intermediate styles, but I'll have to ask Nyshadh about it at some point. Thanks for pointing that out to me, though.

Let me know if I can do anything else here!

davelab6:
Nice job, thanks both

Stephen, sorry I didn't mention the script family name or Text style thing.
Per css2, we can only have 9 weights in web font families.

The family thing is worth revisiting. In the past you had to say which
subsets you wanted and if you didn't say you only got a Core Latin
character set. But now only msie users experience that, everyone else has
Unicode Range capable browsers, so they get everything available but only
as they need it. So the risk has largely gone away. So if Stephen adds back
those families, it may be possible to merge them.

The logo thing I think was intentionally removed and as a PUA glyph that's
fine for me.

The respacing is simply unavoidable as a font is improved. We should
probably do some outreach on Twitter and update the description file on
this pr to let people know what happened though.

However: Why did the v metrics change? Were they not compliant with our vm
spec before, or are the only now with this pr?

ofl/ibmplexmono/METADATA.pb
```diff
@@ -1,8 +1,8 @@
 name: "IBM Plex Mono"
-designer: "Bold Monday"
+designer: "UNKNOWN"
```
davelab6: Oops

ofl/ibmplexmono/METADATA.pb
```diff
@@ -1,8 +1,8 @@
...
 license: "OFL"
-category: "MONOSPACE"
-date_added: "2018-03-12"
+category: "SANS_SERIF"
```
davelab6: oops :)

ofl/ibmplexsans/METADATA.pb
```diff
@@ -1,8 +1,8 @@
 name: "IBM Plex Sans"
-designer: "Bold Monday"
+designer: "UNKNOWN"
```
davelab6: Oops

ofl/ibmplexsans/METADATA.pb
```diff
@@ -1,8 +1,8 @@
...
 license: "OFL"
 category: "SANS_SERIF"
-date_added: "2018-03-12"
+date_added: "2019-02-04"
```
davelab6: Oops

ofl/ibmplexserif/METADATA.pb
```diff
@@ -1,8 +1,8 @@
 name: "IBM Plex Serif"
-designer: "Bold Monday"
+designer: "UNKNOWN"
```
davelab6: Oops

ofl/ibmplexsanscondensed/METADATA.pb
```diff
@@ -1,8 +1,8 @@
 name: "IBM Plex Sans Condensed"
-designer: "Bold Monday"
```
davelab6: Oops

PR #1840, opened 2019-02-05
ofl/coiny: v1.00102 added. Updated vert metrics.

### LINKED ISSUE #1629: Coiny font has a wrong ascendent/decendent
Compare the google font verison of coiny, with the same font 'coiny', seems the google API one has a wrong descendent, which is higher than the baseline.

I thought it's special for this font, but I just compare with others, and found that google api's one is special for this, probably i just its wrong with this version.

laerm0:
Vertical metrics have been fixed/properly calculated. Fixes #1629


m4rc1e:
@laerm0 see https://github.com/google/fonts/issues/1629#issuecomment-460973591.

m4rc1e:
Please diff the upstream fonts before commencing work in the future.

laerm0:
Coiny isn't in the upstream repos list.

m4rc1e:
Add it to the upstream doc then.

laerm0:
Well, I will, I just didn't know it existed until your comment.

laerm0:
Hey @m4rc1e, upstream is in the doc now, but I won't fix there: OT is old and a mess and doesn't compile. Save that for another time.

Issue #187, opened 2016-01-25
Check all families that have latin-ext glyphs enable it


davelab6:
There are likely families that have characters only available in the latin-ext subset, but that subset is not yet enabled. It would be good to develop a reliable check tool for this.

rsheeter:
Concept seems valid for all subsets, not just latin-ext?

rsheeter:
sanity_check.py could easily do this provided we can agree on what the bar to "support" the -ext subsets is.

davelab6:
Yes, for sure. Ideally it would be based on language based subsets of the nam files, rather than somewhat arbitrary character total counts.

rsheeter:
If we had a per-subset "supports" bar - lower for latin-ext in particular - would that suffice?

davelab6:
Right, probably made as eg `latin-ext_minimum-glyphs.nam` files

davelab6:
Bumping up the priority of this

@thundernixon this might be something you could tackle

thundernixon:
Sounds like a good project! I'd be up for it.

Are you saying that the `latin-ext` subset should be enabled if there are *any* characters inside that range? And I'm guessing we would do this by including this in the `METADATA.pb` file?

davelab6:
Yes, yes. There is existing software to identify if a font supports the
latin-ext range, I think the add-font script.

daliusd:
Addressing this would improve situation significantly. Now some fonts have latin-ext while they don't support significant range from latin-ext (e.g. Alex Brush) while others don't have latin-ext (Monoton) while situation with them is better.

Still I propose to declare font as latin-ext compatible if at least some characters from latin-ext are available. In some situations something is better than nothing. Problem lies withing Google Fonts API - you can't request latin-ext subset from font that has no latin-ext declared as available (e.g. Monoton) and in result if you want to use this font fully then you should serve this font yourself (vs from Google Gonts servers).

Another problem, I just noticed that Mogra font has zero width for "į". Depending on criterias you choose for latin-ext compatibility this font might lose this status because of this single character.

thundernixon:
@daliusd Thanks for the input here. I filed a separate issue on the Mogra `į`.

thundernixon:
[...]

PR #1861, opened 2019-02-25
rebuild all metadata with gftools add-font to add `ext` subsets where needed

thundernixon:
Work-in-progress PR to resolve https://github.com/google/fonts/issues/187

This uses a slightly-modified version of `gftools-add-font.py` to apply `ext` subset labels where even a single glyph from an `ext` glyph set is present.

Simplified logs of git diffs (removing some of the noise) are available at https://github.com/thundernixon/fix-gf-metadata/tree/master/docs. However, these aren't necessarily a huge improvement on what GitHub shows below, in part because they lack color.

I've written about changes [in these notes](https://github.com/thundernixon/fix-gf-metadata/blob/master/docs/2019-02-22-metadata-diffs.md), but I'll summarize this more concisely here.

### Known issues with the first version of this PR:

- `korean` subsets are dropped due to lack of cp/`.nam` file ([comment in issue gftools issue 92](https://github.com/googlefonts/gftools/issues/92#issuecomment-467132144))
- CamelCased font names are given unwanted spaces (https://github.com/googlefonts/gftools/issues/122)
  - (probably related: "PT Serif" and "PT Serif Caption" become simply "Web")

### Possible/probable issues:

- Other subsets (besides `korean`) are dropped – notably:
  - `chinese-simplified` removed from 3 fonts
  - `devanagari` removed from 7 fonts
  - `greek` removed from 5 fonts
  - `arabic` removed from 3 fonts
- Copyright strings get more escaping than might be necessary
  - I'm not sure whether all of these changes are desired, and simply from the updated Protobuf, or if some/many are incorrect. I need to find a way to see how these would render in practice.
- Adobe Blank gets subset labels, where it excluded these before – I'm not sure whether these will help or hurt its intended purpose
- Language coverage is different between styles, but metadata gets the labels for the more-extensive style (e.g. News Cycle Regular has wider coverage than Bold)

## Primary Current Question

There are a lot of changes introduced by running add-font across the collection. This gives two main paths forward in this PR:
1. Fixing add-font so that it runs as desired on every font in the collection. I'll probably have to loop in Marc and Felipe to assist in these updates. It will take longer and be slightly higher-risk, but will likely have greater overall benefits.
2. Simply adding `ext` labels to applicable fonts, using the existing diffs to guide a script in this, but leaving other possible fixes for later. This will be faster and lower-risk, but may be leaving other existing issues intact, both in metadata and add-font.

@davelab6 / @m4rc1e, which approach do you recommend taking?

---

To-Do items

- [ ] check for `UKNOWN` values, as are currently in the metadata for Open Sans Hebrew + Condensed


thundernixon:
I've updated add-font to make font names in a more intelligent way and have remade metadata files again. Overall, this improves many of the names that were previously getting messed up by being derived from the filenames. Still, there are some aspects that don't seem perfect yet.

Some possible issues:
-  `Cantora One` becomes `CantoraOne`
- `Fjord One` becomes `Fjord`
- `Gemunu Libre` becomes `Gemunu Libre ExtraLight`
- `Hammersmith One` becomes `HammersmithOne`
- `Headland One` becomes `HeadlandOne`
- 3 `Nanum` families lose spaces (e.g. `Nanum Gothic Coding` becomes `NanumGothicCoding`)
- `Nanum Pen Script` becomes `Nanum Pen`
- `Pompiere` becomes `Pompiere ` (it gets a trailing space)
- `Post No Bills Colombo` becomes `Post No Bills Colombo Light`
- `Post No Bills Jaffna` becomes `Post No Bills Jaffna Light`
- `Raleway Dots` becomes `Raleway Dots ` (gets a trailing space)
- `Sansation` becomes `Sansation Light`
- 4 `Seoul` families lose spaces. E.g. `Seoul Hangang` becomes `SeoulHangang` (not sure if this change is incorrect, but it seems off)
- `Sirin Stencil` becomes `SirinStencil`
- `Supermercado One` becomes `Supermercado`
- `Yaldevi Colombo` becomes `Yaldevi Colombo ExtraLight`


I don't know for sure, but I suspect that these issues (at least most of them) are because of odd naming in nameID 16 or 1. As one example, SirinStencil has this for nameID 1, even though it has a space in the GF website:

```
<namerecord nameID="1" platformID="1" platEncID="0" langID="0x0" unicode="True">
      SirinStencil
</namerecord>
```
Likewise, it's "Hammersmith One" on the directory, but:
```
<namerecord nameID="1" platformID="1" platEncID="0" langID="0x0" unicode="True">
      HammersmithOne
</namerecord>
```

thlinard:
So, you get the font nameID 16, else nameID 1, and you obtain:
    * Gemunu Libre ExtraLight
    * Post No Bills Colombo Light
    * Post No Bills Jaffna Light
    * Sansation Light
    * Yaldevi Colombo ExtraLight

This is a kind of bug check for these fonts, isn't it? The fonts should receive a nameID 16, which is obviously lacking.

The other cases are a matter of consistency.

thundernixon:
> This is a kind of bug check for these fonts

That's an interesting way to look at it! NameID 16 is optional, and software just falls back to nameID 1 if it's lacking. But yes, it wouldn't hurt to add 16 to correct the root issue.

m4rc1e:
> That's an interesting way to look at it! NameID 16 is optional, and software just falls back to nameID 1 if it's lacking. But yes, it wouldn't hurt to add 16 to correct the root issue.

NameID 16 and 17 should be used if a font style is non RIBBI. The fonts @thlinard pointed out are all non RIBBI.

Ms spec has more info: https://docs.microsoft.com/en-us/typography/opentype/spec/name#name-ids

You'll find that all of these fonts will also fail Fontbakery on the name checks.

m4rc1e:
This pr enables latin-ext for 668 families. How many families will enable latin-ext if we use the default 10% theshold?

m4rc1e:
@davelab6 If we stick to the original 10% threshold, we'll get an additional 86 families added to latin-ext. This sounds more sensible than adding 233 families which may only contain 1 latin-ext glyph. Shall we proceed with the 86 approach? it also doesn't modify the copyright strings or remove any existing codepages. It will only add the latin-ext codepage of there's more than 80 glyphs.

thundernixon:
*source of "213" added latin-ext families:

Search for the string `+subsets: "latin-ext"` in: 
- https://github.com/thundernixon/fix-gf-metadata/blob/master/docs/git-diffs-ofl.txt
- https://github.com/thundernixon/fix-gf-metadata/blob/master/docs/git-diffs-apache.txt

PR #1872, opened 2019-03-08
rosario: v1.100 vfs added.

m4rc1e:
Taken from the upstream repo, https://github.com/Omnibus-Type/Rosario at commit https://github.com/Omnibus-Type/Rosario/pull/3/commits/9deb1b7d92c7ebb560568f968f4fb7dd9784eeba


m4rc1e:
Closing since #2165 is newer

PR #1876, opened 2019-03-12
manuale: v0.800 vfs added.

m4rc1e:
Taken from the upstream repo https://github.com/Omnibus-Type/Manuale at commit https://github.com/Omnibus-Type/Manuale/pull/5/commits/453aa6d56de3c3652aad79cfe74dced5a6976aa9


m4rc1e:
- ots caret issue will be solved once ots is updated. https://github.com/googlefonts/fontbakery/issues/2268
- Font version record is Version 0.800; ttfautohint (v1.8.1.43-b0c9). FB doesn't like that it's under v1.000
- I'll remove ampersand.001
- There are pixel jumps in the Reg at 14px and 18px. Family has 1.5m views... it should be ok.

m4rc1e:
Closing. #2167 is newer.

PR #1881, opened 2019-03-15
ebgaramond: v1.000 vf added.

m4rc1e:
Taken from the upstream repo https://github.com/octaviopardo/EBGaramond12 at commit https://github.com/octaviopardo/EBGaramond12/pull/25/commits/cbc4c4e7fa4031cb9c900e67bf5cb7f86e591986


ofl/ebgaramond/static/METADATA.pb
```diff
@@ -0,0 +1,103 @@
+name: "EB Garamond"
+designer: "Georg Duffner"
```
davelab6: `designer: "Georg Duffner, Octavio Pardo"`

PR #1894, opened 2019-03-19
add fresh split VF and statics for ofl/encodesans

thundernixon:
I had PR'ed this from my fork, before remembering that it might have to be PR'ed from within the same repo.

If this one runs the automated diff checks, etc, I'll follow suit with the rest.


davelab6:
I think it would be useful to have the ttfautohint version and args shown in the images; I believe this is coming from the before using an older version of ttfautohint with a different increase-xheight default/arg, and the jump could perhaps could be avoided by passing a different increase-xheight arg.

ofl/encodesans/METADATA.pb
```diff
@@ -10,81 +10,15 @@ fonts {
...
+  copyright: "Copyright 2018 The Encode Project Authors (https://github.com/thundernixon/Encode-Sans), with Reserved Font Name \'Encode Sans\'."
 }
 subsets: "latin"
 subsets: "latin-ext"
 subsets: "menu"
 subsets: "vietnamese"
+axes {
+  tag: "wght"
+  min_value: 100.0
+  default_value: 100.0
```
davelab6: I believe this should be 400, since CSS requires the 400 Regular to be the default.

thundernixon: Thanks for the sanity check! I do think this is correct...

As far as I know, a variable font must have a default axis values that align with its default instance. 

At the very least, this seems to be a requirement to build with FontMake. If I set a default of `400` for `wght` in the designspace, then try to generate a variable font, it fails.

This:

```
<axis tag="wght" name="Weight" minimum="100" maximum="900" default="400">
```

...results in this:

```
$ fontmake -o variable -m master_ufo/EncodeSans.designspace
 
# build information
 
File "/Users/stephennixon/type-repos/google-font-repos/Encode-Sans/venv/lib/python3.7/site-packages/fontTools/varLib/__init__.py", line 748, in load_designspace
    assert base_idx is not None, "Base master not found; no master at default location?"
AssertionError: Base master not found; no master at default location?
```

I assume that the default instance must be described for legacy apps that can't handle variable fonts. In the past, we have had conversations around whether this should be the light or bold master, if there is no middle master to provide the default values.

I made a quick test of this for browsers, just to check whether a default `wght` value of 100 would cause any issues. As far as I can tell, it doesn't.

Chrome:
image

---

Safari:
image

---

Firefox:
image

---

Windows 10, Edge:
image

thundernixon: I _think_ the MS opentype spec also stipulates this must be the case, though their definition is admittedly somewhat complex. I take this to mean that the default value of an instance should represent what is in the glyph table, before any deltas are applied:

> The default value interacts with the glyph and glyph variations tables in a particular way: the variation instance that has the default coordinate value for each axis will use glyph outlines as defined in the glyph table, without any variations from the glyph variations table applied. This instance is referred to as the default instance.

[from the fvar spec](https://docs.microsoft.com/en-us/typography/opentype/spec/fvar#variationaxisrecord)

thundernixon:
@davelab6 made the suggestion that it would be worth making a shell script to generate fonts with different TTFautohint arguments.

> I believe this is coming from the before using an older version of ttfautohint with a different increase-xheight default/arg, and the jump could perhaps could be avoided by passing a different increase-xheight arg.

I've started testing this with a script that hints with a bunch of different options applied. I can extend this script and make it more systematic, if useful.

https://github.com/thundernixon/Encode-Sans/tree/8ae7b5b41b150125d8b7815527f0aad4daebe2da/docs/16-autohint-settings-test

However, a blocker is that I don't know what the current system is for accessing BrowserStack to use [diffbrowsers](https://github.com/googlefonts/gf-docs/blob/57f28ebf288d5c4529bf957181be00b94c6e215d/ReviewingFamilies/README.md#diffbrowsers). @m4rc1e should I get a browserstack account, or is there a way I can use the existing Google Fonts account? I assume it wouldn't be optimal to blast this PR with 20+ different pushes to test different hinting strategies, but I could do so if you think it's the simplest way forward.

ofl/encodesans/METADATA.pb
thundernixon: Something I was curious about that I hadn't tested earlier ... if the `font-weight: 100 900;` is not declared in the `@font-face`, Encode Sans still shows up as a `400` weight in Chrome, while all weights just continue to show up as before in Safari.

image

thundernixon:
Closing in favor of weight + width PR: https://github.com/google/fonts/pull/1768#issuecomment-547558237

Issue #1460, opened 2018-02-25
Add Fira Code - on deck to merge!


tonsky:
https://github.com/tonsky/FiraCode

please. Let me know if you need something changed/altered

davelab6:
I decided not to include this (https://github.com/google/fonts/pull/520)

tonsky:
Oh, sorry, this is different.

Fira Code (original, my repo) is a significantly different version that adds more than 100+ ligatures to Fira Mono, designed to improve how programs (code) looks on screen. It’s a big change and huge value added for programmers, compared to “just monospaced letters”.

I released it Nov, 11 2014 and it’s been growing in popularity ever since. To date it’s the most popular programming font with ligatures (20k+ stars on GitHub) and it’s been the pioneer of that movement from the beginning, forcing IDEs to add support for it, for example (Atom, Idea, VS Code, iTerm 2, Sublime Text—none of them had it in 2014, and added ligatures support because of Fira Code pressure). You can browse through some feedback here: https://twitter.com/search?q=Fira%20Code&src=typd

There’s been a long-standing request to get Fira Code on the web https://github.com/tonsky/FiraCode/issues/18, and the Google Fonts is amongst most popular ones and easiest to use. Using ligatures on web is both possible and desirable, and hosted Fira Code would be the best way to do so.

On names: on Oct 17, 2016 (2 years after my Fira Code release) Carrois released another version of Fira Mono, as you noticed, with altered line-height, which they unfortunately called Fira Code as well. As far as I know, nobody is very attached to that version of Fira Mono, either as a name or as a font. 

I hope my arguments convince you that Fira Code name is de facto associated with ligatures font, which I ask you to add in this issue. Please reconsider. And thank you for what you’re doing!

Best,
Nikita

eriawan:
@davelab6 
I support @tonsky comment above. Fira Mono is simply a fork of Fira Code, and both of them are different.

> Fira Code (original, my repo) is a significantly different version that adds more than 100+ ligatures to Fira Mono, designed to improve how programs (code) looks on screen

Yes, his comment above is also true.
Please reconsider to add the original Fira Code to Google Fonts.

AronT-TLV:
👆👆🏻👆🏼👆🏽👆🏾👆🏿

Fira Code is THE best font for code editing. Fira Mono doesn’t come close. Please reconsider

ai:
Ligatures of Fira Code is very useful

pozda:
[...]

Issue #1817, opened 2019-01-18
How to best set the filenames of VFs


davelab6:
Currently we name VFs filenames as "Family-Regular.ttf"

I think filename "Family-Roman.ttf" and "Family-Italic.ttf" is better, especially as esteemed superuser @thlinard, @eliheuer and others have pointed out (at https://github.com/google/fonts/pull/1814#issuecomment-454889240 and elsewhere) that Family-Regular.ttf (400 static) is confusable with Family-Regular.ttf (roman VF)

Also, if the file is named differently on Github, will the download-ZIP get a renamed file?

For now, Marc has set up PRs for VFs with "normal" names (Family-Regular.ttf) and I'm going to merge these, with static fonts in each family's subdirectory, but @rsheeter suggested on a IM chat to file this separate issue to ponder VF file naming, as it's very much something we can alter and the GF eng team has not yet fully considered it.

m4rc1e:
Agreed.

Atm VFs are named by including the family name and the **default** instance, FamilyName-DefaultInstance.ttf.

m4rc1e:
I did start writing about this issue in #1580

thundernixon:
A related question: Should we have any strategy around naming the family in the nameID 16 (typographic family name, for apps which separate family and style names into separate menus), in order to differentiate VFs from statics? For instance, we could have "Encode Sans Var" and just "Encode Sans." Presumably, it might make sense for the suffix to be the same in the filename and nameID 16. 

Related to fontbakery fails described at https://github.com/googlefonts/fontbakery/issues/2297

davelab6:
We want the vf to be transparent upgrade, so don't do this :)

davelab6:
To clarify: 

I want the VFs to be transparent upgrades, so don't do this *to the name tables* :)

I think the filenames should be different so it is possible to have in the same directory both a static 400 roman style (eg `Family-Light.ttf` or `Family-Regular.ttf` or `Family-Bold.ttf`) and a VF that supplies those (and other) family-styles (`Family-Roman.ttf`)

Whereas, I think the "expected" family name metadata in the `name` table should be found in the VF, because that is the latest version of the font family. I can see that it would be good to adjust the static fonts' `name` metadata so that they can also be installed together for legacy apps/platforms; sadly the version numbers in fonts are not used very much by platforms/apps to allow sets of the same family  with the same `name` table metadata to be installed concurrently. 

[...]

PR #1911, opened 2019-03-28
fira code: v1.207 added.

thundernixon:
Fira Code is a highly popular monospace font, famous for including "code ligatures."

A variable font has been built from it, and a few pieces of metadata have been cleaned up to pass FontBakery checks.

This PR isn't immediately ready to merge – remaining to-do items are:

- [x] Finish vector outline check + fixes (Light master was extrapolated, and has a few issues) – [quick pass correcting outline issues summarized here](https://github.com/thundernixon/FiraCode/blob/qa/googlefonts-qa/notes/outline-checks.md)
- [ ] Wait for https://github.com/mozilla/Fira/pull/219 to merge, so we have permission to use name "Fira"

---

Upstream Repo: https://github.com/tonsky/FiraCode
Working QA branch: https://github.com/thundernixon/FiraCode/tree/qa
PR Closes: https://github.com/google/fonts/issues/1460


thundernixon:
@tonsky You're welcome to take a look at this, if you like! I know you care about this project a lot, so I want to make sure you have some say before we merge it. Is the variable font working as you expect? Is the html description looking good to you? Are you finding any issues I may have missed? 

If you're busy with other work; no worries – I think things are solid here, but I just wanted to welcome input from you, if you have any.

tonsky:
@thundernixon thanks for pinging me! Looks great!

BTW not sure if this is intentional or not, but shouldn’t [this file](https://github.com/google/fonts/blob/9555b351342cd003c28c60ed88b365ceaac8b45f/ofl/firacode/FiraCode-Light.ttf) be called FiraCode.ttf (no -Light suffix)?

thundernixon:
@tonsky cheers! Thanks for taking a look.

> shouldn’t this file be called FiraCode.ttf (no -Light suffix)?

Good question! Variable font filenames will soon be converted to a [format described in GF issue 1817](https://github.com/google/fonts/issues/1817#issuecomment-466556000). For Fira Code, I think that will look like `FiraCode[wght].ttf`. However, I haven't imposed that in this PR, because we (probably @m4rc1e) will just make a new [gftools](https://github.com/googlefonts/gftools) script to convert filenames for many VFs at once.

In general, VFs contain some default naming metadata based on their default instance, so that they'll make sense as a fallback in legacy software that doesn't understand the new OpenType tables used for variable fonts. As a random example, if you load the Fira Code VF in a 2011 MS Word, it will (as far as I know) show up as `Fira Code Light` and show the associated glyph shapes.

tonsky:
Shouldn’t Regular be the default then?

thundernixon:
Another good question! 

To make this work as a variable font, it was necessary to set up the font file so that it wouldn't rely on extrapolation for the extremes. In your version of the source, there are Regular and Bold masters which output Light through Bold instances. Those can be built as static fonts, but VarLib doesn't currently support such a setup.

So, the two choices were to 
1. Make a light master and let the be the default (alternatively, the bold could be the default)
2. Make a light master *plus* keep the Regular master so it could be the default. This would create a significantly larger VF output, with little advantage. So long as the font is being viewed on modern software, almost no users will even realize that the light master is a default state, because the browser and CSS will provide their own styles.

tonsky:
I just dived into your PR and noticed there’s no Regular master anymore. That explains it, thanks!

thundernixon:
The latest update adds fresh statics to the build – previously, these were accidentally left out.

davelab6:
@thundernixon mozilla/Fira#219 is merged!

m4rc1e:
The VF contains 5 instances. One of them is the following style, which is in between Regular and Medium:
```
    <!-- Retina -->
    <NamedInstance flags="0x0" subfamilyNameID="259">
      <coord axis="wght" value="450.0"/>
    </NamedInstance>
```

@garretrieger Will having such an instance cause issues? if not, we can merge this. I've attached the full [fvar dump](https://github.com/google/fonts/files/3367252/FiraCode-fvar.txt)

rsheeter:
We don't rely on named instances, shouldn't matter

twhite96:
Is this live because I search for it and don't see it.

thundernixon:
@twhite96 There are internal engineering processes that happen to further optimize fonts (especially variable fonts) to make things work with the API / generate CSS, statics, etc. This PR merging into the public repo means were a lot closer than before! It also means that the font has gotten some useful improvements, and should be solid if you'd like to use them for self-hosting before they make it onto the public directory & API.

twhite96:
>@twhite96 There are internal engineering processes that happen to further optimize fonts (especially variable fonts) to make things work with the API / generate CSS, statics, etc. But, this PR merging into the public GF repo means we're a lot closer than before! It also means that the font has gotten some useful improvements, and should be solid if you'd like to use it for self-hosting before it makes it onto the public directory & API.

@thundernixon okay! No worries.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #1934, opened 2019-04-10
publicsans: v1.000 added

eliheuer:
Taken from the upstream repo https://github.com/uswds/public-sans at commit uswds/public-sans@b871432


davelab6:
> rename publicsans to legacy vf filenames

Rejecting this until that is fixed.

eliheuer:
I'm closing this to create separate PRs for Roman and Italic.

PR #1962, opened 2019-05-03
Update README.md

Swivelgames:
**GitX** hasn't been in active development for 10 years, and the **Laullon** fork died five years ago. **MSYSGIT** also ceased development around the same time.

However, `git-scm.com` provides a comprehensive list of user-friendly and intuitive Git GUIs for those who don't want to use the command line or might not be as savvy with it. It's probably best to send users there.

Either way, these links need to be updated, because they're both out of date.


Swivelgames:
I signed it.

PR #2016, opened 2019-06-10
gayathri: v1.000 added.

m4rc1e:
Taken from the upstream repo https://github.com/smc/gayathri at commit https://github.com/smc/Gayathri/commit/5ad0cb435bd61227de1e95dc54ee85bfca671c53


m4rc1e:
@santhoshtr are these fonts autohinted? the rendering doesn't look too good imo.


Desktop_Windows_10_chrome_69 0_
_Win 10 Chrome 69_

Everything else seems fine to me.

m4rc1e:
View the image full size to see what I mean,.

PR #2020, opened 2019-06-12
literata: v2.200 added.

m4rc1e:
Taken from the upstream repo https://github.com/googlefonts/literata from the tagged release https://github.com/googlefonts/literata/releases/tag/2.200


m4rc1e:
Fixed in #2021

PR #2021, opened 2019-06-12
literata: v2.201 added.

m4rc1e:
Taken from the upstream repo https://github.com/googlefonts/literata from tagged release https://github.com/googlefonts/literata/releases/tag/v2.201


davelab6:
> com.google.fonts/check/canonical_filename
> 🔥 FAIL This is a variable font, but it is using a naming scheme typical of a static font.
> 🔥 FAIL Please change the font filename to use one of the following valid suffixes for variable fonts: VF, Italic-VF, Roman-VF

Accept this as the upstream is named this way

PR #2035, opened 2019-06-21
lexend: v1.000 added.

m4rc1e:
Taken from the upstream repo https://github.com/ThomasJockin/lexend at commit https://github.com/ThomasJockin/lexend/commit/6cfd6902018178dc06d0bd4f3d59f165c407d62f


m4rc1e:
@davelab6 fails are false positives.
- The description is over 1000 bytes. However, there are many hyperlinks so the visible text is actually much shorter.
- FB is reporting that https://revreading.org/wp-opt-perf/ is broken, it isn't. However, there are broken links in the description because I'm referencing the sibling families which don't exist yet. I'll file an fb issue.

The other sibling families will also have these issues. Please feel free to adjust the descriptions if needed.

cc @ThomasJockin

m4rc1e:
I've closed these prs because the family names are going to change.

PR #2049, opened 2019-06-27
livvic: v1.000 added.

m4rc1e:
Taken from the upstream repo https://github.com/Fonthausen/Livvic at commit https://github.com/Fonthausen/Livvic/commit/885959cd198c1f825d110c125f73c70ca3a14244


davelab6:
Would be great to have a VF with wght axis here

PR #1675, opened 2018-09-03
archivonarrow: v1.010 added

m4rc1e:
Taken from the upstream repo https://github.com/Omnibus-Type/ArchivoNarrow at commit https://github.com/Omnibus-Type/ArchivoNarrow/commit/5c10afd57eccd85f15f7d872c57221a4d39d3bbf

---
## Fontbakery report

<details>
<summary><b>[2] ArchivoNarrow-MediumItalic.ttf</b></summary>
<details>
<summary>:fire: <b>FAIL:</b> Checking OS/2 Metrics match hhea Metrics.</summary>

* [com.google.fonts/check/042](https://github.com/googlefonts/fontbakery/search?q=com.google.fonts/check/042)
* :fire: **FAIL** OS/2 sTypoAscender and hhea ascent must be equal. [code: ascender]

</details>
<details>
<summary>:fire: <b>FAIL:</b> Checking with Microsoft Font Validator.</summary>

* [com.google.fonts/check/037](https://github.com/googlefonts/fontbakery/search?q=com.google.fonts/check/037)
* :fire: **FAIL** MS-FonVal: The value doesn't match the calculated value DETAILS: maxPoints = 96, calculated = 98
* :fire: **FAIL** MS-FonVal: The value doesn't match the calculated value DETAILS: maxComponentElements = 4, calculated = 0
* :fire: **FAIL** MS-FonVal: The value doesn't match the calculated value DETAILS: maxComponentDepth = 3, calculated = 0

</details>
<br>
</details>
<details>
<summary><b>[2] ArchivoNarrow-SemiBold.ttf</b></summary>
<details>
<summary>:fire: <b>FAIL:</b> Checking OS/2 Metrics match hhea Metrics.</summary>

* [com.google.fonts/check/042](https://github.com/googlefonts/fontbakery/search?q=com.google.fonts/check/042)
* :fire: **FAIL** OS/2 sTypoAscender and hhea ascent must be equal. [code: ascender]

</details>
<details>
<summary>:fire: <b>FAIL:</b> Checking with Microsoft Font Validator.</summary>

* [com.google.fonts/check/037](https://github.com/googlefonts/fontbakery/search?q=com.google.fonts/check/037)
* :fire: **FAIL** MS-FonVal: The value doesn't match the calculated value DETAILS: maxPoints = 82, calculated = 92
* :fire: **FAIL** MS-FonVal: The value doesn't match the calculated value DETAILS: maxComponentElements = 4, calculated = 0
[...]

PR #2066, opened 2019-07-12
archivonarrow: v2.001 added.

m4rc1e:
Taken from the upstream repo https://github.com/Omnibus-Type/ArchivoNarrow at commit https://github.com/Omnibus-Type/ArchivoNarrow/commit/af3f04719d253d7f88dca83de71ecca00fd82f92

Replaces #1675


m4rc1e:
The hhea metrics are now the same as the typo. I'll need to test this from sandbox on real sites.

Issue #2085, opened 2019-07-26
Quicksand font: letters cropped in <select>


TomOne:
The text of `<select>` elements with `font-family: "Quicksand"` is cropped and only partially visible even with no additional CSS is applied.

It looks like this: ![image](https://user-images.githubusercontent.com/4357973/61937613-85b61e80-af8f-11e9-8b64-60fddca253e0.png)

Live example: https://codepen.io/zoyara/pen/GVjdNq (tested in Chrome v75)

Interestingly this only happens when at least two font-weights are loaded from Google Fonts, e.g.
`<link href="https://fonts.googleapis.com/css?family=Quicksand:400,700&display=swap" rel="stylesheet">`

With only one font-weight, the text in the `<select>` element renders normally:
`<link href="https://fonts.googleapis.com/css?family=Quicksand&display=swap" rel="stylesheet">`

Example with one font-weight: https://codepen.io/zoyara/pen/aemGvX. This example is identical to to previous one except it only loads `https://fonts.googleapis.com/css?family=Quicksand:400&display=swap`

m4rc1e:
Thanks for reporting this. Can confirm the issue for Win Chrome only.

This issue is being caused by the MVAR table. If I strip out the table, the results are fine.

<img width="672" alt="Screenshot 2019-07-26 at 10 37 41" src="https://user-images.githubusercontent.com/7525512/61942491-6cae6d00-af91-11e9-9021-fffd300a0fd5.png">

I reported this issue to Microsoft last year.

For our other variable fonts, Comfortaa and Oswald, I stripped the MVARs out before releasing them.

@nyshadhr9 Shall I quickly push an update where I simply strip out this table? For this family, the MVAR is simply changing the strikeout offset by 12 units which is rather negligible. This issue isn't present for Dosis and Kreon which we also just rolled out.

@felipesanches In Fontbakery can we add the "MVAR" table to our check [com_google_fonts_check_unwanted_tables](https://github.com/googlefonts/fontbakery/blob/master/Lib/fontbakery/profiles/universal.py#L531)?

m4rc1e:
@TomOne thank you once again. The issue should now be solved.

<img width="1136" alt="Screenshot 2019-07-27 at 00 16 44" src="https://user-images.githubusercontent.com/7525512/61985987-07479400-b004-11e9-984a-fabc2e82563a.png">

TomOne:
Great, thank you

arrowtype:
[...]

PR #2090, opened 2019-07-26
firacode: v1.208 added

m4rc1e:
Hotfix to remove MVAR table


thundernixon:
Was this table causing rendering issues, or just a waste of filesize?

m4rc1e:
It causes the following issue #2085

thundernixon:
Ohh right. I believe that was also responsible for really extreme line heights in MS Word, for Inter. Thanks for pointing me to that example, and thanks for this patch!

PR #2149, opened 2019-09-02
jura: remove MVAR table

m4rc1e:


m4rc1e:
Closing since #2163 is newer.

PR #2220, opened 2019-10-22
adding the license file for the Tomorrow family

felipesanches:


davelab6:
Thanks Felipe! I would love a check for this so that new PRs like the original PRs for this family would not have been merged and we avoid a PR like this in the future :) Filed https://github.com/googlefonts/fontbakery/issues/2675 to track :)

PR #2243, opened 2019-11-07
nunito: v3.504 added (Adds Cyrillic)

m4rc1e:
Taken from the upstream repo https://github.com/googlefonts/nunito at commit https://github.com/googlefonts/nunito/pull/20/commits/76794b6d58d68864dbbea3a1a0f9d4bb0c3cd06c


davelab6:
Nice to see this Cyrillic work go out

PR #2254, opened 2019-11-13
dancingscript: v2.000 added

m4rc1e:
Taken from the upstream repo https://github.com/impallari/DancingScript at commit https://github.com/impallari/DancingScript/pull/3/commits/a6d4598b189b8a882fc61f218cbb37fcfb830e47


thlinard:
If there is a script in FontBakery to warn that static fonts should not be deleted but placed in a /static/ folder, it doesn't work.

m4rc1e:
Thanks. Statics added.

We do have a check which is still wip, https://github.com/googlefonts/fontbakery/pull/2662

PR #2258, opened 2019-11-16
Fix typo on Mike LaGattuta's name & add upstream repo URL to DESCRIPTION file

felipesanches:


felipesanches:
This PR does not change the actual font files. It only fixes a typo on the name of one of the designers on the METADATA.pb file and adds a reference to the upstream git repo on the DESCRIPTION.en_us.html file. So it is completely safe to merge, as it does not affect the font rendering.

felipesanches:
There is a FontBakery FAIL in the report that could be addressed in a future PR which would require someone to review the quality of the rendering, as the FAIL is related to a hinting setting.

Issue #2259, opened 2019-11-20
Please restore Jura static fonts


thlinard:
The static fonts were suppressed by PR #2070

Issue #1763, opened 2018-11-22
 Please restore Vollkorn static fonts


thlinard:
The static fonts (4 weights + italics) were suppressed in PR #1571

laerm0:
Hmm. How are we handling the static data stored going forward? @nyshadhr9 @davelab6 @m4rc1e

m4rc1e:
@laerm0 Moving the static fonts to a new dir in each family called 'static'. This is temp btw.

Issue #1666, opened 2018-08-27
Don't suppress non-VF version


thlinard:
Isn't a bit too early to suppress regular font files in favor of VF beta files?

Example with this Vollkorn commit:
https://github.com/google/fonts/commit/652b3a586c077a715184aa9751847d44fd2ba213

PR #2260, opened 2019-11-20
Restore static files

thlinard:
Fix:
Issue #2259
Issue #1763
Issue #1666


m4rc1e:
Where did you get the static fonts from?

m4rc1e:
We're also thinking of deprecating this static dir folder since users should be getting the fonts directly from [Google Fonts](https://fonts.google.com)

thlinard:
Hi @m4rc1e,

From my personal archives of this repo. If you feel more confident to get them back from a diff, fine, of course.

m4rc1e:
Thanks for the quick response.

The fonts we add to static dirs are not the previous fonts. They're generated from the same source as the VF.

We face a bit of a conundrum here because Google Fonts generates these static fonts automatically from the VF and lets users download them from Google Fonts directly. We're adding them ourselves to this repo but they're not getting shipped to end users. By dropping this dir, we'll probably upset users like yourself so I'm hesitant to remove these dirs at the moment.

thlinard:
> We're also thinking of deprecating this static dir folder since users should be getting the fonts directly from [Google Fonts](https://fonts.google.com)

Well, are the tools mature enough? I just checked a couple of VF typeface, but I don't obtain the same files:

- When I download [Vollkorn](https://fonts.google.com/specimen/Vollkorn), I obtain only static files, but not of the same weight as the static files of this repo.
- When I download [EB Garamond](https://fonts.google.com/specimen/EB+Garamond), I obtain VF + static files, but not of the same weight as the static files of this repo.

thlinard:
> Thanks for the quick response.
> 
> The fonts we add to static dirs are not the previous fonts. They're generated from the same source as the VF.
> 
> We face a bit of a conundrum here because Google Fonts generates these static fonts automatically from the VF and lets users download them from Google Fonts directly. We're adding them ourselves to this repo but they're not getting shipped to end users. By dropping this dir, we'll probably upset users like yourself so I'm hesitant to remove these dirs at the moment.

Are the process of generating static files for this repo and the process generating the static files on Google Fonts really identical? I didn't go further than comparing the size of the files, but obviously they are not the same files.

m4rc1e:
> When I download Vollkorn, I obtain only static files, but not of the same weight as the static files of this repo.

Vollkorn (not the beta) is still a static family. We'll release a VF later this year.

> When I download EB Garamond, I obtain VF + static files, but not of the same weight as the static files of this repo.

This shouldn't be the case. I'll look into this.

thlinard:
> Vollkorn (not the beta) is still a static family. We'll release a VF later this year.

In https://github.com/google/fonts/tree/master/ofl/vollkorn I only found VF files, and not beta.

m4rc1e:
> Are the process of generating static files for this repo and the process generating the static files on Google Fonts really identical? 

They are not the same. We typically generate static fonts using fontmake. The Google Fonts backend uses fontTools varLib Instancer. I think there may be subtle differences in rounding but nothing which should be apparent to the majority of end users.

m4rc1e:
> Vollkorn (not the beta) is still a static family. We'll release a VF later this year.

> > In https://github.com/google/fonts/tree/master/ofl/vollkorn I only found VF files, and not beta.

Well it hasn't been pushed to production. We usually do an internal review before this happens. I'll take another look at the family but I don't think it's ready.

thlinard:
OK, thanks Marc.

PR #2276, opened 2019-12-07
Add Tibetan font DDC_Uchen

andyjgf:


davelab6:
Please add the OFLtxt, a second paragraph to description file with a link to the up-stream, rename the ttf to follow the Family-Style.ttf canonical form, and check the font with FontBakery to ensure there are no FAIL results, and review the WARN results :)

ofl/ddcuchen/METADATA.pb
```diff
@@ -0,0 +1,17 @@
+name: "DDC Uchen"
```
davelab6: Per the naming requirements in https://github.com/google/fonts/blob/master/CONTRIBUTING.md we may want to remove the DDC

andyjgf: done

m4rc1e:
@andyjgf I've hotfixed the fonts so they pass the serious checks.

It seems Tibetan requires very loose vertical metrics.

Screenshot 2019-12-11 at 14 42 13

_Uchen tested on https://r12a.github.io/scripts/tibetan/ using fontdragr_

This looks great for Tibetan but the Latin does appear very loose.

Screenshot 2019-12-11 at 14 44 53

_Uchen test on http://www.cyreal.org/Font-Testing-Page/_

NotoSans Tibetan has similar metrics so this should be ok.

PR #2294, opened 2019-12-13
Cairo v3.000 added

eliheuer:
From the upstream repo: https://github.com/eliheuer/titillium-web-vf
At commit: 109d49f


m4rc1e:
Closing. #2380 is newer

PR #2350, opened 2020-02-19
HOTFIX tradewinds: threesuperior and threequarters needed outline swap

### LINKED ISSUE #1939: TradeWinds ¾ and ³ are switched
I noticed that the ¾ and ³ are switched in the TradeWinds font. Can anyone switch them to the correct codepoints. For reference: ¾ is u+BE, ³ is u+B3.

laerm0:
fixes #1939 

I swapped the outlines of these two glyphs via ttx.


RosaWagner:
Looks good, no other regression.

glyphs_modified

PR #2351, opened 2020-02-19
hammersmithone: name table updated to match API

### LINKED ISSUE #1954: HammersmithOne nameID1
> Hi, I’m working on a font library in .NET and I noticed a strange behavior: the font family names returned by the Google Font API and the font family names saved inside the related TTF files are not the same. As an example, “Hammersmith One” TTF file contains “HammersmithOne” as font family name. Can you fix this behavior, please?

laerm0:
Fixes #1954 

Updated name table to match font names served by the API and in `METADATA.pb`


RosaWagner:
@m4rc1e, since name ID 1 has changed, does it mean we should create a new family for this fix to avoid conflicts? If yes… does it really need to be fixed?

Sorry @laerm0 for all the notifications you get, I am doing a small end-of-summer-cleaning.

davelab6:
@RosaWagner its the opposite - as reported in #1954, the font's name table data was malformed and does not match what is used in the API, so this brings the font into alignment with the API to solve the existing conflict. This needs to be fixed, somewhat urgently.

RosaWagner:
AH well let's merge then

PR #2401, opened 2020-03-25
cabin: v3.000 added.

m4rc1e:
Taken from the upstream repo https://github.com/impallari/Cabin at commit https://github.com/impallari/Cabin/pull/20/commits/a0cd53b13f2fff0e257b064347fa49b8b65e0993


thlinard:
Shouldn't the new (version 3.000) Condensed static files be put in the static folder also, and the old (version 2.200) [Cabin Condensed folder](https://github.com/google/fonts/tree/master/ofl/cabincondensed) be removed?

PR #2412, opened 2020-04-07
Remove stale families (Baloo v1, old renamed ones)

davelab6:
They are available as part of baloo*2 families

I also removed the other stale families


thlinard:
Hi @davelab6 

Could you also add my commit  c12a3b4acfa90821d9dba09787434d6d6a4b7636 ?

PR #2414, opened 2020-04-08
Update Roboto to v3 ("Roboto Classic")

davelab6:
- Add VFs with weight and width axes
- Upgrade license from Apache to OFL
- Remove legacy fonts
- Add old static files from [v2.136 roboto-hinted.zip release](https://github.com/google/roboto/releases/download/v2.136/roboto-hinted.zip) as the VF files are not subset as previous web releases were


davelab6:
> Add old static files from v2.136 roboto-hinted.zip release as the VF files are not subset as previous web releases were

@rsheeter note that these are not "web subset" per your internal request to do so, but this will inflate the size of the `latin-ext` and `cyrillic-ext` subsets generated in the API

davelab6:
@m4rc1e OK, @rsheeter confirmed that he wants to proceeed ASAP so I'll add the web-subset statics here and merge

m4rc1e:
This PR has the following issues:

- Version number is not v3.000
- Apache license info is still present in fonts
- No viz diffs/sign off have occurred.

@davelab6 would you like the above done and pushed in a new pr?

davelab6:
yes please :)

PR #2441, opened 2020-05-01
Removed unintented line from the block

keshavgbpecdelhi:
Unnecessary # and empty line removal makes it nice and if not gap without # can be implemented


RosaWagner:
This line is intended

PR #2444, opened 2020-05-04
heebo: v3.000 added

m4rc1e:
Taken from the upstream repo https://github.com/TypeNetwork/heebo at commit https://github.com/TypeNetwork/heebo/commit/00a501d1d1d7cc4ce97127dd4e7c309849acdf1b


richardhsueh:
https://github.com/google/fonts/issues/2453
Hi, I am not sure is the problem related to this update (but I think this most recent update related to Heebo ), can you help to verify the problem ?
Thanks a lot. 🙏

PR #2461, opened 2020-05-18
museomoderno: v1.001 added

m4rc1e:
Taken from the upstream https://github.com/Omnibus-Type/MuseoModerno at commit https://github.com/Omnibus-Type/MuseoModerno/commit/922413068096b117652824e82ec544d779fa69c6


m4rc1e:
@davelab6 thoughts on the name "MuseoModerno". Should it be "Museo Moderno"?

cc @omnibus-type

Omnibus-Type:
From a conversation between @davelab6, Jos Buivenga and me (07.20.2019):
Me: To avoid confusion, we'll propose that the name of the typeface be _MuseoModerno_ (without space). 
DC: That works for me
JB: I hope & think this will cause the least confusion if the Museo name is used.

davelab6:
Confirmed.

PR #2474, opened 2020-05-26
Fix DESCRIPTIONs for cormorantupright and notable

davelab6:
cormorantupright has `N/A` in production, so that needs to be pushed 

notable, the description I've added here is taken from production, so nothing to do there


m4rc1e:
Thanks. A period was missing at the end of Cormorant's description so I added it. I'll get eng these changes this week.

PR #2438, opened 2020-04-29
encodesans: v3.0 added.

thundernixon:
This PR replaces the current `encodesans` directory with a wght+wdth variable font, and includes a `static` folder with autohinted static instances.

This PR relates to and supercedes a previous PR: https://github.com/google/fonts/pull/1768. For a record of what has changed in the upstream from that PR, see [my notes on finishing this upgrade](https://github.com/thundernixon/Encode-Sans/tree/master/docs/17-finishing).

Upstream: https://github.com/thundernixon/Encode-Sans

## To do:
- [x] add `variable font` to description
- [x] Fix copyright string. `copyright 2020 the Encode Sans project authors (stephen@thundernixon.com), with reserved font name "encode sans”.`
- [x] update instance style names to include width descriptors
- [x] check that the 'Regular' instance has a 'wdth' (Width) axis coordinate must of 100, fix if needed
- [x] Update license URL with `https` URL in OFL & in font info
- [x] Check copyright namerecords match license file.
- [x] **semi-blocked:** I can't simply follow FontBakery's suggestion of adding `Axis Location` custom parameters to each master in the Glyphs source, as this causes glyphsLib to fail to generate the correct axis map, resulting in bad middle-instance wght values on the fvar table. See https://github.com/googlefonts/fontmake/issues/543. I'll probably have to either edit the designspace in the build, or edit the final OS/2 table with TTFont. I'll probably do the latter. [Solution found](https://github.com/googlefonts/fontmake/issues/543#issuecomment-621951478)
- [x] `🔥 FAIL TYPOGRAPHIC_SUBFAMILY_NAME for Win "Thin Condensed" is incorrect. It must be "Thin". [code: bad-typo-win]` Solution: this name was incorrect in the masters and had to be corrected.
- [x]  FAIL: Checking OS/2 usWeightClass matches weight specified at METADATA.pb.
- [x] check & fix alignment of vietnamese accents. These are off in certain masters.
- [x] correct flipped numerators & denominators.

## Questions
- [x] Is a "reserved font name" okay? We never heard back from Impallari on whether we had his blessing to change that. UPDATE: in https://github.com/google/fonts/pull/1768, Dave said that we could keep the RFN for this.

## Incorrect FontBakery checks

> 🔥 FAIL: A static fonts directory with at least two fonts must accompany variable fonts

There are static fonts in the `static` folder.

> ⚠ WARN: Combined length of family and style must not exceed 27 characters. com.google.fonts/check/name/family_and_style_max_length

[...]

PR #2497, opened 2020-06-11
encodesans: v3.001 added.

thundernixon:
This makes almost no changes to the fonts from https://github.com/google/fonts/pull/2438, except:
- Static filenames have been improved (the main reason to upgrade).
- The glyph `dotlessi.sc` has been added to the smallcap glyphs


m4rc1e:
Thanks. LGTM.

thlinard:
Hi @thundernixon 

I don't want to delay publication, as this version is much better than the previous one, but why Encode Sans "**SC**" in nameID 1, 3 and 4 (static files)?

m4rc1e:
Good spot @thlinard.

```
    <namerecord nameID="1" platformID="1" platEncID="0" langID="0x0" unicode="True">
      Encode Sans SC Cnd Black
    </namerecord>
    <namerecord nameID="3" platformID="1" platEncID="0" langID="0x0" unicode="True">
      3.001;GOOG;EncodeSansSC-CndBlack
    </namerecord>
    <namerecord nameID="4" platformID="1" platEncID="0" langID="0x0" unicode="True">
      Encode Sans SC Cnd Black
    </namerecord>
```

@thundernixon would you mind updating the static files?

thundernixon:
😬 Thanks for spotting that, @thlinard.

I’ve done a bit of digging, and found that the problem is caused by unwisely combining logic to fix SC names with logic to abbreviate names more generally (done to meet character length limitations) in [a single Python script](https://github.com/thundernixon/Encode-Sans/blob/011cec896d1131512265193c34cbb8bf283c170c/sources/scripts/helpers/fix-SC-names.py).

I have some urgent tasks right now, but I will separate these parts of the flow to the static files this weekend.

PR #2498, opened 2020-06-12
sourceserifpro: v3.001 added

m4rc1e:
Taken from the upstream repo https://github.com/m4rc1e/source-serif-pro at commit https://github.com/m4rc1e/source-serif-pro/commit/5e368e18cac90df9bec383d6ce50a25483a3d109


thlinard:
Hi @m4rc1e 
You don't want to use the VF version?

m4rc1e:
Will push it later on. Not eta yet though since we'll need to tweak some hot fix scripts.

PR #2522, opened 2020-06-29
Sora: Version 2.000 added

graphicore:
* Sora Version 2.000 taken from the upstream repo https://github.com/sora-xor/sora-font.git at commit https://github.com/sora-xor/sora-font/commit/efeb79b4eff1090588196321b53bc9fea722d6f9.


graphicore:
Sorry for the noise. This was just a test ...

PR #2541, opened 2020-07-12
GF Axis Registry setup

nathan-williams:
Initial setup of the `axisregistry/` directory containing GF Axis Registry metadata.

### DIFF
--- a/axisregistry/README.md
+++ b/axisregistry/README.md
@@ -1,18 +1,33 @@
 # GF Axis Registry
 
-This package contains a collection of metadata files that collectivefly form
-the GF Axis Registry.
+This package contains a collection of metadata files that collectivefly form the
+GF Axis Registry.
 
 ## Why does Fonts have its own Axis Registry?
 
-Any font foundry or distributor library that offers variable fonts has a
-latent/de-facto axis registry, which can be extracted by scanning the library
-for axes' tags, labels, and min/def/max values and publishing a document that
-makes this info explicit. We are doing the same in order to help our users adopt
-variable fonts.
+We plan to support a superset of the OpenType axis registry and additional
+metadata. Individual variable fonts support a subset of our axis registry. Axes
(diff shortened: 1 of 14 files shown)


axisregistry/README.md
```diff
@@ -0,0 +1,18 @@
+# GF Axis Registry
+
+This package contains a collection of metadata files that collectivefly form
+the GF Axis Registry.
+
+## Why does Fonts have its own Axis Registry?
+
+Any font foundry or distributor library that offers variable fonts has a
+latent/de-facto axis registry, which can be extracted by scanning the library
```
rsheeter, line 9: This is only meaningfully true if the vendor uses unregistered or custom axes

axisregistry/README.md
```diff
@@ -0,0 +1,18 @@
...
+for axes' tags, labels, and min/def/max values and publishing a document that
+makes this info explicit. We are doing the same in order to help our users adopt
+variable fonts.
+
+While in 2016 Microsoft originally offered to include more axes in the OpenType
+1.8 specification (https://github.com/microsoft/OpenTypeDesignVariationAxisTags)
+as of 2020-05 they have abandoned that effort. We hope that in publishing
+explicit "Google Fonts Axis Registry" docs, we will reinvigorate Microsoft's
+effort to update the OpenType specification's axis registry.
```
rsheeter, line 18: Suggest delete paragraph. Overall, this why section seems a bit verbose. IIUC the key points are:

*   We plan to support a superset of the OpenType axis registry and additional metadata.
*   The axes Google Fonts supports are documented here.
*   The meaning of the fields is <either explain or link to where explained>.
*   Clarify individual families support a subset of our axis registry and that axes not in our registry won't function via our API

nathan-williams: Done. Let me know if you have further suggestions and I can make further revisions after merging the branch.

PR #2544, opened 2020-07-13
Fix email address in copyright line of OFL.txt for librebaskerville

TimWolla:
This also makes sure that the OFL.txt is valid UTF-8 instead of ISO-8859.


TimWolla:
Note: I'm not perform the bureaucratic CLA process for a single character typo fix. Please either take the PR as is or just close it if the CLA is an absolute requirement.

PR #2551, opened 2020-07-15
s/SAN_SERIF/SANS_SERIF for Encode Sans.

garretrieger:


m4rc1e:
Thanks Garret. I've submitted an issue to FB since it didn't catch it.

Issue #2557, opened 2020-07-18
URL Inconsistency Issue with the fonts


anushbhatia:
If you search a font in any other browser will face a major bug that is if a font is typed in First Word capital for example: https://fonts.google.com/specimen/Montserrat **(It will work fine**)😊
https://fonts.google.com/specimen/montserrat  **(this won't)**😥
This creates an issue of URL inconsistency. Might cause a surge of bounce back rate so to reduce this we can have a single named allocated to a font.

Preferred method will be to have the URL in lowercase.

On Uppercase (https://fonts.google.com/specimen/Montserrat)
![Montserrat - Google Fonts and 10 more pages - Personal - Microsoft​ Edge 18-07-2020 16_36_06](https://user-images.githubusercontent.com/40017559/87851279-e7943480-c914-11ea-9ac4-3d2572c5d127.png)

On Lowercase (https://fonts.google.com/specimen/montserrat)
![Google Fonts and 7 more pages - Personal - Microsoft​ Edge 18-07-2020 15_34_03](https://user-images.githubusercontent.com/40017559/87851293-04c90300-c915-11ea-8329-9cdad5a2bbf5.png)

**I would like to work on this can you please assign me this issue🎉🎉** 
_Thanks for going through this Issue and all suggestions are welcome_

anushbhatia:
@davelab6 @m4rc1e any suggestions?

anushbhatia:
@davelab6 @m4rc1e

PR #2568, opened 2020-07-22
merriweathersans: v2.000 added.

m4rc1e:
Taken from the upstream repo https://github.com/TypeNetwork/Merriweather-Sans at commit https://github.com/TypeNetwork/Merriweather-Sans/commit/b37362977e331ffce377ba94b96fd866b76d59e4


anushbhatia:
sir reply to my pr #2557

thlinard:
Hi @m4rc1e 

Could gftools-fix-vf-meta.py be applied to these two VF? Their STAT table is almost empty.

m4rc1e:
@thlinard we're going to talk to Adobe soonish regarding this matter. Even if you get the STAT tables "correct", the sliders may not work in Indesign etc. We can easily push an update once the apps have sorted out their support. Anyways, pdfs and postscript don't support VFs yet so I recommend not using them for DTP apps.

m4rc1e:
@thlinard that's not to say we won't fix this. It's just we're not sure what's the right approach, given the current state of vf support in these apps.

thlinard:
@m4rc1e well, at least with gftools-fix-vf-meta.py the fonts work in Photoshop (named instances and sliders), and the named instances work in InDesign.

There, only the named instances work.

Adobe is certainly at fault, but on the other hand I understand them: they need properly made fonts first to test their own bugs.

PR #2577, opened 2020-07-28
Newsreader 1.000

yanone:
Upstream: https://github.com/productiontype/NewsReader
My contributions are merged as I worked directly on the repo.

The vertical metrics FAIL is okay because within each `opsz` family they are consistent.


davelab6:
In https://github.com/google/fonts/blob/master/ofl/newsreader/METADATA.pb#L30 opsz min of 0 seems wrong, and per https://github.com/google/fonts/blob/e637ce0ab9f82042f3e63061eed1fe77223a2b34/axisregistry/optical_size.textproto#L6 it won't work

thlinard:
Newsreader-Italic[opsz,wght].ttf lacks a Regular entry in the `fvar` table (or, better, `gftools fix-vf-meta`).

thlinard:
Oh, and caretSlopeRun/caretSlopeRise are wrong in Newsreader-Italic[opsz,wght].ttf (should be something like 611/2000, 0/1 is for Regular, not Italic).

thlinard:
And, in light of @m4rc1e comments https://github.com/googlefonts/fontbakery/pull/3021, fsSelection bit 7 should be enabled.

davelab6:
Thanks @thlinard, all these small changes sounds good to me for @RosaWagner to do

RosaWagner:
> Oh, and caretSlopeRun/caretSlopeRise are wrong in Newsreader-Italic[opsz,wght].ttf (should be something like 611/2000, 0/1 is for Regular, not Italic).

I actually don't know how to solve this

davelab6:
Its just values in the `hhea` table that you can set with fonttools

https://docs.microsoft.com/en-us/typography/opentype/spec/hhea

davelab6:
I would guess there is a GlyphsApp CuustomParameter or UFO fontinfo.plist to set them in source

RosaWagner:
The scripts called in the build script are actually quite complicated, so I gonna do it directly with fonttools on the variable. @thlinard can you confirm the values you gave are correct?

thlinard:
> The scripts called in the build script are actually quite complicated, so I gonna do it directly with fonttools on the variable. @thlinard can you confirm the values you gave are correct?

Yes, 611/2000. Or 11/36, which is the same, as only the value of the fraction is important. But it's more conventional to take head.unitsPerEm as a starting point, so 611/2000 (post.italicAngle=−17.0, so tan(73°)=3.2708526, and 2000 ÷ 611 = 3.2733224, which is the closest value — formula in https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6hhea.html).

Btw, a good hhea.caretSlopeRun value has a good side effect: InDesign uses caretSlopeRun/caretSlopeRise to order its styles menu and needs the values to be different between Roman and Italic.

RosaWagner:
Thank you for the explanation!

thlinard:
De rien. 😉

It's strange, as this is more of a FontLab bug (fixed in 7.1.4: "Previously, when you exported font instances, FontLab would always export hhea.caretSlopeRun as 0 and hhea.caretSlopeRise as 1, like in upright fonts that have the italic angle of 0°" https://help.fontlab.com/fontlab/7/manual/Release-Notes-7515/), Glyphs has a single italicAngle parameter which "Affects the CFF ItalicAngle, the post italicAngle, the x offsets of the OS/2 subscript and superscript values, as well as the hhea caretSlo-peRise and caretSlopeRun entries" https://glyphsapp.com/downloads/handbook/Glyphs-Handbook-2.3.pdf).

Roboto Mono has the same bug: https://github.com/TypeNetwork/RobotoMono/issues/21

moyogo:
The UFOs should either not have info.openTypeHheaCaretSlopeRise and info.openTypeHheaCaretSlopeRun defined or should have them with the correct values.

Right now their values are 1 and 0 even in the italic UFOs.

They can be removed because fontmake+ufo2ft will generate them from info.italicAngle.

See ufo2ft’s [openTypeHheaCaretSlopeRiseFallback()](https://github.com/googlefonts/ufo2ft/blob/e1c08f6efb07a2fb5202d65a90af94dbd599e7e8/Lib/ufo2ft/fontInfoData.py#L135) and [openTypeHheaCaretSlopeRiseFallback()](https://github.com/googlefonts/ufo2ft/blob/e1c08f6efb07a2fb5202d65a90af94dbd599e7e8/Lib/ufo2ft/fontInfoData.py#L155).

ofl/newsreader/METADATA.pb
```diff
@@ -0,0 +1,37 @@
...
+  full_name: "Newsreader Italic"
+  copyright: "Copyright 2020 The Newsreader Project Authors (http://github.com/productiontype/Newsreader)"
+}
+subsets: "latin"
+subsets: "latin-ext"
+subsets: "menu"
+subsets: "vietnamese"
+axes {
+  tag: "opsz"
+  min_value: 0.0
```
davelab6, line 30: This value should not have been accepted, and has been set to the axis minimum here

https://github.com/google/fonts/commit/92ec7cb6699bcdc28a8c23a38805820a2161cd66

PR #2585, opened 2020-07-30
Remove source keys from METADATA.pb files

andyjgf:


davelab6:
This broke the staging servers as the updated protobufs definition has not propagated everywhere. This will likely be reverted in future. 

cc @m4rc1e @graphicore

PR #2590, opened 2020-08-01
ofl/petrona/METADATA correct date

davelab6:
Ringo asked for the date to be updated, since the family was completely redrawn


davelab6:
@ringoseeber it is done :)

PR #2596, opened 2020-08-04
Remove ofl/newsreader/static/*ttf

davelab6:
These family names don't match our naming spec, so we shouldn't offer them.

cc @yanone @m4rc1e


davelab6:
(For comparison, Literata is good)

PR #2600, opened 2020-08-04
ofl/kufam/METADATA.pb add arabic subset

davelab6:
In https://github.com/originaltype/kufam/issues/2 I proposed adding the arabic design to the italics, duplicated from Roman.

Since the arabic characters are not present in half the files, then `add-font` didn't include the arabic subset METADATA enumeration, so I'm adding it in this PR. 

When pushed to the catalog, the result will be tofu for arabic text in the Italic styles.

### DIFF
--- a/ofl/kufam/METADATA.pb
+++ b/ofl/kufam/METADATA.pb
@@ -21,6 +21,7 @@ fonts {
   full_name: "Kufam Italic"
   copyright: "Copyright 2019 The Kufam Project Authors (https://github.com/originaltype/kufam)"
 }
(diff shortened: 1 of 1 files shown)


davelab6:
Screen Shot 2020-08-04 at 5 16 32 PM

m4rc1e:
The Italic fonts lack Arabic. Shall we merge this anyway?

davelab6:
@m4rc1e and I discussed and since Italic is a foreign concept for Arabic script, we will proceed

PR #2608, opened 2020-08-08
Upgrade Recursive description to better highlight the specimen website

thundernixon:
The current Google Fonts [specimen page for Recursive](https://fonts.google.com/specimen/Recursive) has a description which I believe was meant to only be temporary, and fails to say almost anything about the project:

image

This PR adds a description that improves this by adding a bit of detail and making the website link slightly more prominent. It will now look like this:

image

An earlier version of this PR incorporated ASCII art, but after some offline conversations, I have agreed to simplify. Thanks to everyone for the help, here!

<details>
<summary><b>Earlier version of this PR</b> (Click to expand)</summary>

The Google Fonts specimen page buries the link to the Recursive website, so basically any casual viewer would likely see only a few basic styles but totally miss most of what the project has to offer. This seems like a big miss (having commissioned a font and having worked on the full API, presumably you want to help people make use of both of these). Obviously, though, it is a tricky thing to design a specimen pages that works equally well for all 1000 fonts in the collection, including outliers like Recursive. So, this PR uses a bit of ascii art to make the page design bring a little more attention to the Recursive specimen website.

This is specifically designed to work with the Roboto UI font of Google Fonts, and tested to make sure it displays well down to small screens like the iPhone 5.

image

image

</details>

### DIFF
--- a/ofl/recursive/DESCRIPTION.en_us.html
+++ b/ofl/recursive/DESCRIPTION.en_us.html
@@ -1,12 +1,13 @@
 <p>
-Recursive draws inspiration from single-stroke casual signpainting, a style of brush writing that is stylistically flexible and warmly energetic.
-Adapting this aesthetic basis into an extensive type system, Recursive is designed to excel in digital interactive environments, including data-rich user interfaces, technical documentation, and code editors.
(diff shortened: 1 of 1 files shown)


davelab6:
@m4rc1e please help shephard this to sandbox and prod :)

m4rc1e:
Sure thing!

PR #2423, opened 2020-04-15
Add detail to description for Recursive

thundernixon:
This description assumes we will publish Recursive around May 1, after an update to the Recursive site that will help users configure advanced Google Font API calls.

If we decide to publish it earlier, we should switch the final paragraph here to:

```html
<p>
To learn more, visit the Recursive website at <a href="https://www.recursive.design/">recursive.design</a>. To configure Google Fonts embed code from Recursive’s full set of styles and variable axes, please see <a href="https://developers.google.com/fonts/docs/css2">developers.google.com/fonts/docs/css2</a>.
</p>
```

PR #2617, opened 2020-08-12
Add recursive to to_sandbox

davelab6:
It seems the updated description in https://github.com/google/fonts/pull/2423 has not made it to sandbox and prod, but should go through soon :)

cc @thundernixon


m4rc1e:
LGTM

Issue #2649, opened 2020-08-26
Unblock publication of Newsreader


davelab6:
As describe in https://github.com/google/fonts/pull/2577#issuecomment-669502406

RosaWagner:
Will be resolved with #2837

PR #2661, opened 2020-09-02
Newsreader: version 1.002 added

RosaWagner:
Taken in upstream repo <https://github.com/productiontype/NewsReader> at PR <https://github.com/productiontype/NewsReader/pull/9>.

It fixes issue <https://github.com/google/fonts/issues/2649> as describe in PR <https://github.com/google/fonts/pull/2577>.

- Enable fs selection bit 7
- Correct caretRun and caretRise values
- Add Italic fvar instance
- Corrected opsz axis min value
- Correct STAT table to get along better with Indesign


m4rc1e:
- Rerun `gftools add-font path/to/google/fonts/ofl/newsreader`
- add static fonts in a new dir called /static
- add OFL.txt

When you add a family to the google/fonts repo, I recommend running fontbakery again on the family dir because it will now run additional checks for the metadata etc.

thlinard:
>     * Correct caretRun and caretRise values

Oh, sorry for not being clearer: 611/2000 means 2000 for caretSlopeRise and 611 for caretSlopeRun.

Like this, in ttx code:

```
    <caretSlopeRise value="2000"/>
    <caretSlopeRun value="611"/>
```

RosaWagner:
Yes yes. the ofl text was already there but in lowercases… I don't understand why the v.1.001 didn't have these these fails, the font-origin is the same for the 2 versions and I didn't touch Yanone's name-table script. Are you sure the statics have to be there for this project ? There were not there at the first place, I thought it was somehow intentional.

RosaWagner:
There are 3 families for the statics: Caption, Text and Display. The variable font origin is NewsreaderText-Regular. It looks like the font doesn't match the spec and it should be named Newsreader6pt, Newsreader16pt and Newsreader72pt. 

1. Shall we take the same process as Literata (and in that case I review sources, designspace and build script) 
2. Or is there something special about this project with the naming and we make 3 families, with the variable in the NewsreaderText family ?

davelab6:
(1) is what I prefer, please chat with @productiontype about it

yanone:
@RosaWagner 
Can you please describe what `1) Shall we take the same process as Literata (and in that case I review sources, designspace and build script)` means, as I wasn't involved in Literata and am now taking over the process again? Thank you

RosaWagner:
It was about the naming mostly: Literata has a workflow allowing to export instances with proper naming and a STAT table with optical size axis. We discussed it with Production Type and they wanted to do the corrections themselves. They closed the issues I raised in their repo to guide them in the process, so maybe the font are ready now…

yanone:
@googlebot I consent.

RosaWagner:
@ googlebot I consent.

RosaWagner:
About the variable font name… we had this conversation recently : https://github.com/googlefonts/gf-docs/issues/83
It basically says that the master origin (default master) shouldn't contain, for example, "16pt" because the default optical size should be elided.

yanone:
Okay, I'll change that. Give me a moment.

yanone:
Done

RosaWagner:
Ah the OFL.txt file has to be set in caps for Fontbakery to take it into account

m4rc1e:
Both fonts still have their opsz min value set to 0 but the metadata has 6. Will this cause any problems?

yanone:
I changed the opsz minimum to 6pt, renamed all 16pt to 18pt. Thus, the fonts are already in line with the upcoming unpublished fontbakery checks regarding the axis registry.

RosaWagner:
Always according to this conversation googlefonts/gf-docs#83, now the default optical size shouldn't be elided.

Since the initial default value for opsz was 16, we should hold this PR until a decision is made about allowing values out of the axis registry.

PR #2673, opened 2020-09-08
Delete expression.textproto

davelab6:
We no longer use this axis, so let's remove it from the registry


davelab6:
Duplicate of https://github.com/google/fonts/pull/2594

PR #2674, opened 2020-09-09
Delete Syne static fonts

davelab6:


RosaWagner:
I would like to merge but apparently "some checks are not successful"… do you know why ? Can it be safely merged ?

PR #2561, opened 2020-07-21
Added Commissioner 1.000

yanone:
Upstream PR: https://github.com/kosbarts/Commissioner/pull/34

PR #2675, opened 2020-09-09
Commissioner: version 1.000 added

RosaWagner:
This PR follows this one : https://github.com/google/fonts/pull/2561 which I will close now.

- The spacing issue is not fixed, we decided to let it this way. Although while searching for a solution, the designer updated the kerning and some other component stuff.

- I also added some informations about the designer in designers/kostasbartsokas


davelab6:
Note that this was missing all the axes in the METADATA file. Its now been pushed to production, so needs a METADATA update and new push.

PR #2677, opened 2020-09-11
Update SOFT axis registration

davelab6:
Based on discussion with @sponcey in https://github.com/undercasetype/Fraunces/issues/209


cozysd:
Hey all… we're eagerly watching Fraunces development and can't wait to see it on Google Fonts. Any timeline for that?

PR #2718, opened 2020-10-06
Ballet: metadata fixed

RosaWagner:
Added default_value to opsz axis in METADATA.pb


RosaWagner:
Thanks! after changing this file 3 times, I preferred waiting for you to validate it haha

PR #2737, opened 2020-10-09
Commissioner: METADATA.pb updated with all axes

RosaWagner:
So this is a great opportunity to clarify the metadata default values (related issue: https://github.com/googlefonts/gf-docs/issues/83#issuecomment-706113386)

- Default master has `weight=100` and `METADATA.pd` displays `weight=400`. What should I do ? change it to 100 manually ? or use `registry_default_overrides`?
- VOLUME and FLAIR should be added to the axis registry ?

### DIFF
--- a/ofl/commissioner/METADATA.pb
+++ b/ofl/commissioner/METADATA.pb
@@ -20,6 +20,21 @@ subsets: "latin-ext"
 subsets: "menu"
 subsets: "vietnamese"
 axes {
+  tag: "FLAR"
+  min_value: 0.0
+  max_value: 100.0
+}
+axes {
+  tag: "VOLM"
+  min_value: 0.0
+  max_value: 100.0
+}
+axes {
+  tag: "slnt"
+  min_value: -12.0
+  max_value: 0.0
+}
+axes {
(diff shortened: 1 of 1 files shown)


davelab6:
> registry_default_overrides

This is for over-riding values in axis registry; its not related to the `glyf` table design.

davelab6:
> Default master has weight=100 and METADATA.pd displays weight=400. What should I do ? change it to 100 manually ? or use registry_default_overrides?

@m4rc1e is going to look into the previous discussion of this, we couldn't find it on a call just now, but any family with this is going to be blocked until this is resolved

davelab6:
OK, the issue is here: https://github.com/googlefonts/fontbakery/issues/2683

The conclusion was that while the Default `glyf` master has `weight=100`, since the `wght` range covers `400` then its fine for `METADATA.pb` to have `weight=400`. 

Therefore nothing is needed - no change to 100 manually and no use of `registry_default_overrides` :)

RosaWagner:
This is blocked until we figure out the axis registry / metadata / zip file issue

PR #2739, opened 2020-10-12
volume.textproto added to axisregistry

RosaWagner:


m4rc1e:
Thank you for adding these!

axisregistry/volume.textproto
```diff
@@ -0,0 +1,14 @@
...
+display_name: "Volume"
+min_value: 0.0
+default_value: 0.0
+max_value: 100.0
+precision: 1
+fallback {
+  name: "Normal"
+  value: 0.0
+fallback {
+  name: "Loud"
```
felipesanches, line 11: @RosaWagner, you need to add closing curly bracket here before the second `fallback` keyword. Otherwise it breaks the Protobuf parser.

I noticed this typo when integrating these files to the FontBakery GFAxisRegistry-related checks at https://github.com/googlefonts/fontbakery/pull/3057

RosaWagner: Thanks @felipesanches for the review 👍

m4rc1e:
@RosaWagner did you make this by hand or did you use `gftools add-axis`?

Either way, we should have a validation check for these files.

RosaWagner:
I used the script, of course, but I modified the file manually for this one axis because Commissioner has FLAR and VOL linked in the STAT and so there was an additional fallback name "flair" at 100 in the VOL axis.

davelab6:
> we should have a validation check for these files

Done in https://github.com/googlefonts/fontbakery/pull/3056

PR #2740, opened 2020-10-12
flair.textproto added to axisregistry

RosaWagner:


axisregistry/flair.textproto
```diff
@@ -0,0 +1,18 @@
+tag: "FLAR"
+display_name: "Flair"
+min_value: 0.0
+default_value: 0.0
+max_value: 100.0
+precision: 1
```
davelab6, line 6: I believe this should have been -1

PR #2750, opened 2020-10-14
Categorize Laila as sans-serif

jtbandes:
The description of Laila states: "Laila is an informal sans serif design with brush terminals." However, it's currently categorized as a serif font. Change the categorization to sans.


jtbandes:
@googlebot I signed it!

PR #2754, opened 2020-10-16
[hotfix] ofl/elsie/Elsie-Regular.ttf updated

### LINKED ISSUE #515: Elsie Font
When I type the letters f and i together (fi - lowercase)  it becomes another letter.

RosaWagner:
Supposes to fix #515 by:
- Removing f_i and f_j from file
- updating version (1.002 to 1.003)

few FAILS since it's an old font


RosaWagner:
I should probably add the DSIG table to both fonts, at least…

m4rc1e:
LGTM

Issue #2756, opened 2020-10-17
Self-hosting Alternative - Fontsource


DecliningLotus:
The main [README](https://github.com/google/fonts#self-host-fonts-available-from-google-fonts) has its own individual section addressing self-hosting fonts. 

I was curious whether that section should be updated with references to other solutions such as [Fontsource](https://github.com/fontsource/fontsource) which somewhat addresses other issues and concerns within this repository and serves as useful information for the developers that frequent here. 

Additionally, if there are elements in Fontsource that seem to be concerning, it'd be great to get the feedback and input from the Google team to improve the repository to get it up to standard since there does seem to be demand for a format such as this.

I'd be more than willing to make a PR to update the main README if this sounds like a valid acceptable idea 

Another thing to note is that Kyle Mathews' [Typefaces Project](https://github.com/KyleAMathews/typefaces) did deprecate itself in favour of Fontsource.
References: #222, #472, #1790.

davelab6:
I'm happy to review and accept such a PR. Nice project :)

DecliningLotus:
Resolved in #2757.

PR #2757, opened 2020-10-19
Update readme with self-hosting alternatives

DecliningLotus:
As discussed in #2756, I reformatted the self-hosting section into a list of developer projects that could grow in the future. The descriptions used are taken from their official GitHub repository descriptions.

Seems like Prettier also removed all the extra whitespace on the rest of the README while I worked on this, which I don't believe is an issue either.

Let me know if there are any amendments needed for this!


DecliningLotus:
@googlebot I signed it!

davelab6:
@DecliningLotus thank you so much for this!

PR #2761, opened 2020-10-21
Update grade.textproto

### LINKED ISSUE #2710: Revaluation of grade, GRAD
I propose the values for Grade (`GRAD`) be changed, from min/default/max of -1/0/1, to 1/400/1000, so as to match the value system of the weight (`wght`) axis. The current solution was proposed with no clear alternatives at the time, so to match the UI value range common to all axes, was decided.

Now it’s clear, from questions surrounding implementation issues for dark mode automation in a mobile OS, that having relative and absolute values, for both weight and grade, is required long term. So:

* wght = relative weights 1..1000
* GRAD = relative +/- from 400 at same scale as wght
* XOPQ = weight in 1/1000 of em, measure of typical feature.
* PGRD = grade in 1/1000 of em, measured on same feature as XOPQ

The advantages I can see of a duo of relative axes include:

1. Knowing, from the `GRAD` axis range, whether it’s a narrow ranging axis, e.g. `350..400..450` for tuning, or a broad grad axis range, e.g. `100..400..700` for tuning and weight changes, which is programmatically useful, as well as providing more specific information linking Weight and Grade.
2. A `wght`-relative `GRAD` value enables synthetic `GRAD` axes to be automatically conjured at the same scale as `wght` in monospaced fonts, where `wght` and `grad` share the same variation and programs might not “know” that.
3. A future `avar2` implementation, where values of `wght` shift along `wdth` and `opsz` axes, could maintain accurate `wght`-relative values as the `GRAD` axis needs to be tailored and trimmed around a design space.
4. Being able to communicate and specify the `wght` with a grade change, e.g. saying “400 over 380”, allows for more complex fallback solutions, when solving dark mode issues between font families, and when mixing use of scripts between families.
5. Specifying type with size/leading, e.g. saying “10 over 16 points”, to communicate both size and line spacing, or ordering type “10 point with 12 point bodies”, to communicate type with additional letter spacing applied, are traditional in metal typesetting practice. This addition, of a nominal weight adjusted for particular conditions, without changing other conditions, fits that model.

davelab6:
Closes #2710

### DIFF
--- a/axisregistry/grade.textproto
+++ b/axisregistry/grade.textproto
@@ -1,33 +1,17 @@
 # GRAD based on https://github.com/TypeNetwork/Opentype-1.8-Axis-Proposal/blob/gh-pages/proposal/Grade/ProposalSummary.md
 tag: "GRAD"
 display_name: "Grade"
-min_value: -1
-max_value: 1
+min_value: -1000
+max_value: 1000
 default_value: 0
-precision: -1
-fallback {
-  name: "Full Negative"
-  value: -1
-}
-fallback {
-  name: "Half Negative"
-  value: -0.5
-}
+precision: 0
 fallback {
   name: "Normal"
   value: 0
-}
-fallback {
-  name: "Half Positive"
-  value: 0.5
-}
-fallback {
-  name: "Full Positive"
-  value: 1
 }
 description:
   "Finesse the style from lighter to bolder in typographic color, by varying"
   " stroke thicknesses or other forms, without any changes overall width,"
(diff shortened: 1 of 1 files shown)


davelab6:
cc @dberlow

dberlow:
FB has deployed grades with:
Simple values, 0,1,2,3,4, 1989.
Measured values, per/1000, 2016.
Relative values, per/ 400/1000ths, 2019.
and 
Relative Relative values, -/+, 2020.

Google, Adobe, Microsoft and Apple need to lead on from here.

Issue #2760, opened 2020-10-20
Proposal: update stylenames for SOFT axis to sort alphabetically


arrowtype:
Request: approval from Dave. I can make a PR if that is better.

## Current problem

With the `SOFT` axis style names of Sharp, Soft, and ExtraSoft, many apps will list the resulting static fonts in a non-intuitive order, going from Sharp to ExtraSoft to Soft:

![image](https://user-images.githubusercontent.com/45946693/96608137-07961400-12c7-11eb-85bb-de2992404c53.png)

It seems unclear in the axis definitions whether certain style names should be elided (default names excluded from font/style names, like "Regular" or "Roman"), but if we didn’t elide Sharp, the ordering would be even worse, as `ExtraSoft, Sharp, Soft`.

## Proposed solution

Instead, we should change the [`SOFT` axis definition](https://github.com/google/fonts/blob/b6de8a8ab79225da43df46d7c9dde18a3576878c/axisregistry/softness.textproto) in two ways:

1. Make “Sharp” an elided name (it’s just the default, and @sponcey specifically wishes to avoid the name “Sharp” displaying in the Fraunces font naming because the concept of Fraunces is softness.
2. Make the `SOFT=100` stylename `SuperSoft`, rather than `Extra Soft`, for better sorting of font names. (See the discussion at https://github.com/undercasetype/Fraunces/issues/186#issuecomment-709465964 where we come to this conclusion)

![image](https://user-images.githubusercontent.com/45946693/96608859-dd912180-12c7-11eb-8a15-8d41f1acee3a.png)

PR #2764, opened 2020-10-21
update SOFT axis to close issue #2760

arrowtype:
Summary: this updates the `SOFT=100` style name to `SuperSoft` to allow fonts to sort alphabetically in a way that is more logical & useful for users.

See https://github.com/google/fonts/issues/2760 for details.

Question: do we mark elided style names in axis definitions? If not, should we?

### DIFF
--- a/axisregistry/softness.textproto
+++ b/axisregistry/softness.textproto
@@ -14,7 +14,7 @@ fallback {
   value: 50
 }
 fallback {
-  name: "Extra Soft"
+  name: "SuperSoft"
   value: 100
 }
 description: "Letterforms become more soft and rounded."


axisregistry/softness.textproto
```diff
@@ -14,7 +14,7 @@ fallback {
   value: 50
 }
 fallback {
-  name: "Extra Soft"
+  name: "SuperSoft"
```
m4rc1e, line 17: Shouldn't it be "Super Soft". 

What's wrong with "Extra". We use this in both the weight and width particles.

arrowtype: See https://github.com/google/fonts/issues/2760 for an explanation.

At least on Mac, weights are sorted correctly, probably using OS/2 data. But, these SOFT style sub-families are part of family names, so they won't/can't sort properly without alphabetic names.

davelab6:
I chatted with Evan and he said he agreed with the idea of getting alphabetical sort when possible. 

I also want to remove all the spaces in fallback names

davelab6 (commented):
LGTM

axisregistry/softness.textproto
davelab6: @m4rc1e I think its better to not have spaces in fallback names; I'd like to remove all of them, as we had in static fonts' style names, because then it keeps it easy to identify each particle as a unit.

m4rc1e: @davelab6 When we load the axisreg in fontbakery and gftools, we [remove the spaces in the particle names](https://github.com/googlefonts/fontbakery/blob/master/Lib/fontbakery/profiles/googlefonts_conditions.py#L561-L562). I don't want a situation where we're sending PRs with a naming schema which isn't in line with what already exists. I have no problems if we go and remove all spaces in the existing particles then merge this.

Issue #2587, opened 2020-07-31
Reintroduce METADATA.pb source fields whenever possible


felipesanches:
https://github.com/google/fonts/pull/2585

graphicore:
Looks like some of the packager PR commits are squashed into one commit! this will make it impossible to reproduce the METADATA source commit hash info from this history.

Better use "rebase and merge"

![Screenshot from 2020-12-04 01-29-38](https://user-images.githubusercontent.com/393132/101107784-5546b180-35d3-11eb-8bf7-5d6ab9a29c0d.png)


it may be possible that the commits that "reference this issue" stay accessible, e.g. https://github.com/google/fonts/commit/63217f3db073eb5c5cf076149bd6e4e6f150812e but there's a clear warning:

> This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.

So maybe garbage collection will remove them at some point. Also, from "simply" going through the history, we create a task of maybe fishing in the github api of commits that don't have parents any more. I don't really know what the implications are here.

Not sure if it makes sense (effort related) to restore the info of these squashed commits, but going on, we could just use the "rebase and merge" option and have a really simple way to "undo" these 

> [gftools-packager] {family} remove METADATA "source"

 commits: `git revert <commit>`.

(may or may not work with these "dangling commits" as well)


CC: @m4rc1e @vv-monsalve @RosaWagner

PR #2767, opened 2020-10-23
Texturina: Version 1.002 added

RosaWagner:
d8da2de: [gftools-packager] Texturina: Version 1.002 added

* Texturina Version 1.002 taken from the upstream repo https://github.com/Omnibus-Type/Texturina at commit https://github.com/Omnibus-Type/Texturina/commit/85d61e3424c37595c2c6c8cac44bcc63fd9cf52f.

 5308e46: [gftools-packager] ofl/texturina remove METADATA "source".  google/fonts#2587


RosaWagner:
So, I would like to be sure: this font has its default opsz value at 12 (and not 14 as in the axis registry). Do I need to add the registry_default_overrides key, or, since 14 is covered in opsz values range, it is not necessary ?

Omnibus-Type:
@RosaWagner from our side there are no problems with the value at 14 :)

m4rc1e:
> So, I would like to be sure: this font has its default opsz value at 12 (and not 14 as in the axis registry). Do I need to add the registry_default_overrides key, or, since 14 is covered in opsz values range, it is not necessary ?

I think so, let's add an override key like we did for Ballet.

PR #2794, opened 2020-11-05
designers: converted jpg to png

RosaWagner:
catalog/designers/tirotypeworks
catalog/designers/paulhanslow
catalog/designers/kajaslojewska
catalog/designers/johnhudson


alerque:
Why? JPEG compression is better suited to these kinds of images and now you have the worst of both — the loss from the JPEG source plus the worse compression of the PNG.

PR #2822, opened 2020-11-19
designers/yellowtype: fixed typo

RosaWagner:


felipesanches:
thanks, @RosaWagner 

There's a second typo that was found by the same fontbakery check here: https://github.com/googlefonts/fontbakery/pull/3091#issuecomment-730021677

Issue #2849, opened 2020-12-02
Inclusion of new custom Sharp (SHRP) axis


vv-monsalve:
This new custom axis ins included in [Geologisk](https://github.com/monokromskriftforlag/geologisk#variable-axis) to modify the top terminals of stems.
This would be the draft of the axis registry so we can review the ranges, fallbacks and values.

```

#SHARP based on https://github.com/monokromskriftforlag/geologisk/#variable-axis
tag: "SHRP"
display_name: "Sharp"
min_value: 0.0
default_value: 0.0
max_value: 100.0
precision: 1
fallback {
  name: "Flat"
  value: 0.0
}
fallback {
  name: "Sharp"
  value: 100.0
}
description: "Detailing and terminal treatment."
  "Modifies the slope of the sans serif stem terminals from flat to pointed"
  "As the sharpness increase, different levels of acuteness can be set"

```

![Sharp](https://user-images.githubusercontent.com/48698976/100890763-98473e80-3486-11eb-89f7-c169eca4b0e2.gif)

cc @m4rc1e @davelab6 @rsheeter

vv-monsalve:
Note: The "Sharp" fallback name is also used in [SOFT](https://github.com/google/fonts/blob/master/axisregistry/softness.textproto) axis. Not sure if this could be a source of any conflict, in any case, for us to be aware of it.

thlinard:
Perhaps, before creating the custom axes "Lightness" and "Narrowness", it would suffice to use "Weight" and "Width", and for "Soft" it would suffice to have a fairly broad definition, encompassing sharpness to one end and softness to another?

m4rc1e:
Thank you for including the axis def and providing a gif. It's very clear.

> Note: The "Sharp" fallback name is also used in SOFT axis. Not sure if this could be a source of any conflict, in any case, for us to be aware of it.

Good point. I reckon this may cause some confusion so it's worth exploring other names if we do decide to include this axis.

How many glyphs do this axis modify and could you take a screenshot of a paragraph of text please?

vv-monsalve:
**Letters**
![Sharp-letters](https://user-images.githubusercontent.com/48698976/100913844-4d3a2500-34a0-11eb-93c7-bc5ebf82a172.gif)

**Paragraph**
![Sharp-Paragraph](https://user-images.githubusercontent.com/48698976/100913818-47444400-34a0-11eb-8f15-ecc8615b019c.gif)

**SOFT behaviour in Fraunces**
[...]

PR #2853, opened 2020-12-03
Adding Terminal Slope axis registry

vv-monsalve:
This axis modifies the slope and sharpness of stem terminals.
The inclusion of it was discussed in https://github.com/google/fonts/issues/2849

### DIFF
--- a/axisregistry/terminal_slope.textproto
+++ b/axisregistry/terminal_slope.textproto
@@ -1,6 +1,6 @@
 #SHARP based on https://github.com/monokromskriftforlag/geologisk/#variable-axis
-tag: "TSHR"
-display_name: "Terminal_Sharpness"
+tag: "TSLO"
+display_name: "Terminal_Slope"
 min_value: 0.0
 default_value: 0.0
 max_value: 100.0
@@ -10,10 +10,10 @@ fallback {
   value: 0.0
 }
 fallback {
-  name: "Sharp"
+  name: "Sloped"
   value: 100.0
 }
 description: "Detailing and terminal treatment."
-  "Modifies the slope of the sans serif stem terminals from flat to pointed"
-  "As the sharpness increase, different levels of acuteness can be set"
+  "Modifies the slope of the sans serif stem terminals from flat to sloped"
+  "As the slop increase, different levels of acuteness can be set"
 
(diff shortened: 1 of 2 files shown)


vv-monsalve:
Proposal updated according to this [comment](https://github.com/google/fonts/issues/2849#issuecomment-738939428) in #2849

RosaWagner:
Is it the final proposition?

axisregistry/terminal_slope.textproto
```diff
@@ -0,0 +1,19 @@
+#SHARP based on https://github.com/monokromskriftforlag/geologisk/#variable-axis
```
m4rc1e, line 1: Change `SHARP` to `TSLO`.

axisregistry/terminal_slope.textproto
```diff
@@ -0,0 +1,19 @@
...
+  name: "Flat"
+  value: 0.0
+}
+fallback {
+  name: "Sloped"
+  value: 100.0
+}
+description: "Detailing and terminal treatment."
+  "Modifies the slope of the sans serif stem terminals from flat to sloped"
+  "As the slop increase, different levels of acuteness can be set"
```
m4rc1e, line 18: Perhaps the description could simply be "Description: Adjust the slope angle of sans-serif stem terminals."?

axisregistry/terminal_slope.textproto
vv-monsalve: > Is it the final proposition?

No, I haven't updated it with the latest changes. I was waiting for final comments on the issue where it was discussed.
If we agree on the [latest proposal](https://github.com/google/fonts/issues/2849#issuecomment-739977724), and we don't need any further specific approval I would update it, including the suggestion on the description.

axisregistry/terminal_slope.textproto
vv-monsalve: Done

axisregistry/terminal_slope.textproto
vv-monsalve: I've just updated it

vv-monsalve:
We are giving this a second revision to make it broader or "universal" like `Terminal Shape` so that it opens the possibility to include other possible shapes or forms in the future, e.g. rounded. 

So, I'm closing this here for now.

PR #2854, opened 2020-12-04
Tilt Neon: Version 1.000 added

yanone:
ee3df73: [gftools-packager] Tilt Neon: Version 1.000 added

* Tilt Neon Version 1.000 taken from the upstream repo https://github.com/yanone/Tilt-Typeface.git at commit https://github.com/yanone/Tilt-Typeface/commit/475bebdb230721b357805d4eedc7b762020b4339.


yanone:
Note: This is a preliminary PR from my fork of the upstream repo.
Please review the fonts. I'm particularly interested in how to handle the lack of a weight axis, which throws a few fails.

The family consists of three fonts, each with their own family name, so they would come in separate PRs. I've submitted only one here. The other two are technically identical, just with different outlines.

One fail (regarding the glyph name `gnrl:hyphen`) is going to be resolved in the future. Most likely, it's going to be removed from the fonts. I'm in touch with Andy about it.

RosaWagner:
The last font I tried to onboard without a weight axis (Ballet) never got to the sandbox, no reason given… We should probably raise an issue about that to know more how to handle these cases.

PR #2860, opened 2020-12-08
Libre Barcode EAN13 Text: Version 1.006; ttfautohint (v1.8.3) added

graphicore:
53e1edd: [gftools-packager] Libre Barcode EAN13 Text: Version 1.006; ttfautohint (v1.8.3) added

* Libre Barcode EAN13 Text Version 1.006; ttfautohint (v1.8.3) taken from the upstream repo https://github.com/graphicore/librebarcode.git at commit https://github.com/graphicore/librebarcode/commit/5a3e6055e01b9a2f8b9bccd635faa4e844b8767a.

 23541ae: [gftools-packager] ofl/librebarcodeean13text remove METADATA "source".  google/fonts#2587


graphicore:
This changes the "interface" of the font, hence it would be awesome to have it up ASAP, so that not so many start using the prior contract. Description follows.

Automatically calculating the checksums can be confusing for users who e.g. one would just type in the complete numbers of an ISBN. in that case one would get an EAN13 code plus one digit. *It's a bad idea to break copy and paste.* I think this the best argument for this update. (cc @davelab6)

Hence, now explicitly checksums are supported again. Also a marker char "?" (question mark) can be used to explicitly request check-digit calculation. Hence, e.g.for EAN-13, the two following inputs will produce the same barcode:

`0012345678905` (explicit checksum = 5)
`001234567890?` (triggers checksum calculation)

PR #2876, opened 2020-12-09
roboto: v3.004 added

m4rc1e:
Taken from the tagged release upstream repo https://github.com/TypeNetwork/Roboto at tagged release https://github.com/TypeNetwork/Roboto/releases/tag/v3.004

Please note: We don't have time to push this family into production this year. We need it to go to sandbox only. I will then conduct further tests on the instantiated static fonts for older browsers.


RosaWagner:
Being in 2022, is it now okay to push this PR to production @m4rc1e ?

PR #2883, opened 2020-12-10
Ringo Romei bio info added

vv-monsalve:
This includes info.pb file updated plus bio.html and png pic files added


RosaWagner:
In Petrona's metadata.pb, designer is `Ringo R. Seeber`. Either we change the designer name in Petrona, either we change the name in the catalog. 2nd option is better, right ?

PR #2885, opened 2020-12-10
ringoromei changed to ringorseeber

vv-monsalve:
For consistency sake, the name was updated even in the pic to Ringo R Seeber (ringorseeber)


catalog/designers/ringorseeber/info.pb
```diff
@@ -0,0 +1,6 @@
+designer: "Ringo R. Seeber"
+link: "https://glyph.co"
+avatar: {
+  file_name: "ringoromei.png"
```
RosaWagner, line 4: should be ringorseeber.png

RosaWagner:
@m4rc1e can we remove "ringoromei" from the catalog and add "ringorseeber"? or we should just add "ringorseeber" and leave "ringoromei". 

It would not harm to remove it cause it is not used in any font. But want to be sure before merging

catalog/designers/ringorseeber/info.pb
vv-monsalve: Done. I forgot to commit that file

RosaWagner:
Cool, so I can merge this and we'll see later if we can remove the doppelganger

PR #2890, opened 2020-12-11
Typeland bios

yanone:


RosaWagner:
catalog/designers/alessiamazzarella/AlessiaMazzarella
catalog/designers/typeland

(just to make path copy pasting easier later)

PR #2924, opened 2020-12-17
[gftools-packager] Qahiri: Version 1.01 added

### LINKED ISSUE #2893: Add Qahiri
Once https://github.com/alif-type/qahiri/pull/4 is merged and a new OFL release is out :)

@khaledhosny do you want to [learn how to] do the PR here, or should I ask eg @RosaWagner to do?

khaledhosny:
* Qahiri Version 1.01 taken from the upstream repo https://github.com/alif-type/qahiri at commit https://github.com/alif-type/qahiri/commit/94030f40802dff92cb4a559018836dd1b2df0282. Fixes #2893


khaledhosny:
Fixes #2893

m4rc1e:
- I see some overlap issues:

Screenshot 2021-01-08 at 10 47 04

- The nbspace is set to 0

- There seems to be an issue with the space if quickly test the font on http://www.cyreal.org/Font-Testing-Page/

Screenshot 2021-01-08 at 10 48 48

khaledhosny:
>     * I see some overlap issues:
> 
> 
> Screenshot 2021-01-08 at 10 47 04

Where do you see this (I think I need to set `OVERLAP_COMPOUND` flag for components, but I need to check if it fixes the issue).

>     * The nbspace is set to 0

This is intentional. Both space and no-break-space has default width of 0 to match the traditional Arabic spacing, there is an alternate glyph for Latin with positive advance width.


>     * There seems to be an issue with the space if quickly test the font on http://www.cyreal.org/Font-Testing-Page/
> 
> 
> Screenshot 2021-01-08 at 10 48 48

What browser are you testing with? Looks like `locl` feature is not being applied.

m4rc1e:
I saw both issues in OS X 10.15.7 Chrome 87.

khaledhosny:
I updated the font file to fix the overlap issue.

The missing space in Latin is a Chrome bug, but since the whole point of the Latin part of the font is to give Latin the second class citizen status of other scripts, this is rather fitting and I’m not going to debug it any further.

m4rc1e:
> but since the whole point of the Latin part of the font is to give Latin the second class citizen status of other scripts, this is rather fitting and I’m not going to debug it any further.

Unfortunately many users will use multi-script families for the Latin only. There are more people who use Poppins solely for the Latin than the Devanagari. Due to this, I don't want to merge this family because it will lead to many support issues.

khaledhosny:
The Latin is already unusable, no one is their right mind would want to use it, [it started as a joke](https://twitter.com/KhaledGhetas/status/1281328837750554626) (and a cultural statement) and was never intended to be of any use, and Dave is very well aware of this.

khaledhosny:
Anyway, if you have a solution for the Chrome bug that does not compromise Arabic spacing, I’m happy to use it, otherwise it is not my fault that some software is buggy.

davelab6:
Sadly I am going to require you to set the space glyph to work with Latin, and to match the traditional Arabic spacing your alternate glyph should be for Arabic.

khaledhosny:
The Chrome issue that was blocking this seems to be fixed, I also made the Latin space wider (200 units instead of 80).

PR #2926, opened 2020-12-18
Benne: Version 1.000 added

yanone:
43a3782: [gftools-packager] Benne: Version 1.000 added

* Benne Version 1.000 taken from the upstream repo https://github.com/googlefonts/Benne.git at commit https://github.com/googlefonts/Benne/commit/3fed638b3bd8c6229b54d3e5007fac10645a4b53.


RosaWagner:
The version of this font is the same as the one already published. @yanone can you bump up the version?

PR #2927, opened 2020-12-18
Alegreya: Version 2.009 added

m4rc1e:
b20592b: [gftools-packager] Alegreya: Version 2.009 added

* Alegreya Version 2.009 taken from the upstream repo https://github.com/TypeNetwork/Alegreya at commit https://github.com/TypeNetwork/Alegreya/commit/4ac591fc21fe06ae1d99b9970d1f12c6f47a318b.

 5b8742f: [gftools-packager] ofl/alegreya remove METADATA "source".  google/fonts#2587


thlinard:
Please add the static folder.

m4rc1e:
Since the downloadable zips on fonts.google.com have been fixed, we're no longer adding static dirs unless the statics are manually hinted.

thlinard:
> Since the downloadable zips on fonts.google.com have been fixed, we're no longer adding static dirs unless the statics are manually hinted.

I understand, Marc, but it's so much more convenient to have them here, from where I can gather the 4,000 ttf files from my local Github copy!

m4rc1e:
I see now :-)

Perhaps the site should offer a download all button?

thlinard:
> Perhaps the site should offer a download all button?

That would be great, but is it actually going to happen, given the size of the file to download?

m4rc1e:
@davelab6 wdyt on us offering a download all button on fonts.google.com?

PR #2948, opened 2021-01-11
Fix designer attributes in ofl/lexend*/METADATA.pb

davelab6:
This was changed in the internal METADATA files 2019-8-14, but this wasn't back-propagated to this repo, and then regressed to what is in this repo 2020-11-19.


RosaWagner:
LGTM
We could add Bonnie Shaver-Troup to the catalog.

RosaWagner (approved):

PR #2949, opened 2021-01-11
Kufam font designer's info added

vv-monsalve:
This PR includes the catalog designer's information for Artur Schmal and Wael Morcos, authors of Kufam Family.


RosaWagner:
catalog/designers/arturschmal
catalog/designers/waelmorcos

LGTM

RosaWagner:
Artur Schmal got to sandbox, but not Wael Morcos. I checked info.pb and it misses the dieresis in the designer's name.

vv-monsalve:
> Artur Schmal got to sandbox, but not Wael Morcos. I checked info.pb and it misses the dieresis in the designer's name.

Inspecting this in detail, which dieresis do you mean? The designer sent the information as it is and he doesn't use a dieresis in his name, nor here neither in his [webpage](http://morcoskey.com/about)

RosaWagner:
In the metadata.pb of the font: the designer name has to match the one in the catalog, otherwise they are not linked. 

If if the designer doesn’t want to use the dieresis, then we have to remove the dieresis from kufam’s metadata.

vv-monsalve:
> In the metadata.pb of the font: the designer name has to match the one in the catalog, otherwise they are not linked. If if the designer doesn’t want to use the dieresis, then we have to remove the dieresis from kufam’s metadata.

So the error is in the metadata of the original PR, no info.pb of this one. I'll change the metadata in another PR.

PR #2953, opened 2021-01-13
Commissioner: METATA.pb updated with FLAR and VOL axes

RosaWagner:
We'll have to check the ZIP file in Sandbox

### DIFF
--- a/ofl/commissioner/METADATA.pb
+++ b/ofl/commissioner/METADATA.pb
@@ -20,6 +20,16 @@ subsets: "latin-ext"
 subsets: "menu"
 subsets: "vietnamese"
 axes {
+  tag: "FLAR"
+  min_value: 0.0
+  max_value: 100.0
+}
+axes {
+  tag: "VOLM"
+  min_value: 0.0
+  max_value: 100.0
+}
+axes {
   tag: "slnt"
   min_value: -12.0
   max_value: 0.0


RosaWagner:
Mh. @kosbarts, maybe we could fix the font while we are at it… we can ignore the 2 first fails, but the other are a bit problematic.

kosbarts:
Hi Rosalie,

OK. Let's ignore the first two.
Nr.3 & 6 - I will check the STAT table against Google Spec. I kind of like the use of Upright and the inclusion of Italic but I will change it according to GF.
Nr. 4 - Report says the link if broken however it seems to be working fine. (except that I need to update my website are some point in my life)
Nr. 6 - Will check the componented glyphs again.

kosbarts:
I am perplexed with something re the STAT table.
https://github.com/google/fonts/blob/master/axisregistry/slant.textproto
I had Upright for value=0 and the spec says Default. I changed that. 
When the value is any different from 0 there is no name assigned to it. In my STAT I am using Italic. Shouldn't there be some value there to indicate that something is happening?

kosbarts:
@RosaWagner did you get a chance to see my response?

RosaWagner:
Yes, I am a bit perplexed too with the "Default" fallback name… This is how I understand it: there is a lack of support from browsers for the slant axis, so Google doesn't serve it yet. Although we can implement it in fonts, but the axis registry only take the default value with an elided fallback name, so it is not exported in the ZIP file at the end. When there will be better support for slant axis, GF will just have to update the slant.textproto file. I'll talk about it with a colleague who has the same issue. We can also open a discussion to see how to anticipate things the best (maybe fallback names like "0deg", "8deg" or "Upright" and "Slanted"…) but whatever fallback names you will chose in the STAT may have to change in the future… In any case "Italic" can't be one of the fallback names cause it is reserved to actual italics for style linking. 

The other option, my favourite, is to ditch the slant axis for now and have two ttf variable files: a Roman and an Italic, with an ital axis (0,1). So the Zip contains at least an italic.

I checked with Dave yesterday, and we actually also have to wait for confirmation for FLAR and VOL. And I don't how much time we have to wait.

kosbarts:
@RosaWagner yeah I understand the problem when using "Italic". I could opt to have it named Slanted. As for breaking it into two separate variable files I kind of disagree with that since it seems to be breaking the whole efficiency of variable fonts. 

Since we still need to wait for confirmation on the other axes I will just push the current fixes. The only one that remains is to figure out what to do with the non-default name in the slnt axis in the STAT table.

RosaWagner:
We try to coordinate with @arrowtype and @vv-monsalve who both work on projects with slant axis so we keep some kind of a consistency. For now the most obvious choice would be "Oblique" (and "Upright" instead of "default" looks better and relevant, but I will see with people to see if it can be changed).  

What doesn't feel good with Oblique is that there is not really possible stops in between like "0deg", "5deg", "10deg" would allow. But it is also a very used name and most projects for now doesn't need in-between fallbacks… we are going to inspect that deeper and see what makes more sense for now. I keep you updated.

kosbarts:
@RosaWagner OK. Thanks.

kosbarts:
In the meantime I have pushed the latest version with all fixes, except from the STAT table entry for the slnt axis ≠ 0.

RosaWagner:
Boh I close this since we are not close to resolution with the slant axis thing

PR #2955, opened 2021-01-14
Update info.pb

yanone:


RosaWagner:
I think we need more specific PR titles for the catalog with the directory name

PR #2979, opened 2021-01-28
Archivo: Version 2.001 added

RosaWagner:
40a2dc9: [gftools-packager] Archivo: Version 2.001 added

* Archivo Version 2.001 taken from the upstream repo https://github.com/Omnibus-Type/Archivo at commit https://github.com/Omnibus-Type/Archivo/commit/555fa4a86de865637a2e5c35123ddb8972333467.

 7201068: [gftools-packager] ofl/archivo remove METADATA "source".  google/fonts#2587


m4rc1e:
Horizontal advance widths have changed a lot (I guess the wider space glyph is causing this).

RosaWagner:
Yep, I know. There is also kerning between K and space, which makes it even more obvious with this sample string. We can show it to Dave tomorrow to have confirmation. I personally think it is better this way, but I also spent quite some time reversing weight and v metric regressions, so this one is like okay compare to previous upstream version.

davelab6:
I think this is good, the spacing is not overall changed in personality, but improved

PR #2991, opened 2021-02-04
Updated catalog info for Rasa (3x)

yanone:


RosaWagner:
I can only see the update for Rosetta in sandbox, cause David and Anna are not in Rasa's metadata.pb.

yanone:
The update is not PRd to GF yet. Waiting for some upstream work. I have the two other names in the new metadata.

RosaWagner:
ok perfect

PR #2997, opened 2021-02-10
Open Sans: Version 3.000 added

m4rc1e:
b440e73: [gftools-packager] Open Sans: Version 3.000 added

* Open Sans Version 3.000 taken from the upstream repo https://github.com/googlefonts/opensans at commit https://github.com/googlefonts/opensans/commit/c071e9696751f2d69925d66f3b8c581f2dc041c2.

 f6b53ed: [gftools-packager] apache/opensans remove METADATA "source".  google/fonts#2587


m4rc1e:
Summary of differences:
- Light Italic and Italic are wider.  In many cases, documents will reflow as demonstrated below.
- A few glyphs have been redesigned (I guess they are fixes)
- Glyphs with apostrophe like marks have had their spacing fixed e.g dcaron
- Minimal rendering differences (imo a pass)
- Vertical metrics values have changed. The visual appearance is the same but it may cause issues (will research further)

**Win 10 Chrome 71**


winxp_ie_7 0
**Win XP IE 7**

m4rc1e:
I think these changes have occurred due to the evolution of the family and its history. Noto Sans, Open Sans and Droid Sans are all related. I assume they were created in the following order, Droid Sans, Open Sans and then Noto Sans. In order to create this family, Type Network started with Noto Sans VF. They then made some modifications in order to reduce the amount of regressions. I agree with this approach since Noto Sans fixed several issues in Open Sans.

cc @moyogo are my assumptions here correct?

moyogo:
> @moyogo are my assumptions here correct?

Yes that is correct.

Here's what I had noted in https://github.com/TypeNetwork/opensans-1/pull/29#issue-427854740
> This is to test Using Noto Sans 4 masters to produce 4 masters for Open Sans.
> 
> Mapping between Open Sans weights and Noto Sans master values:
> Open Sans (GF) | wght | Noto Sans (master) | wght 
> ---------------|------|--------------------|-----
> –              |      |               Thin | 26
> Light          |   50 |               –    |     
> Regular        |   83 |               –    |     
> –              |      |            Regular | 90
> SemiBold       |  117 |               –    |     
> Bold           |  151 |               Bold |   151
> ExtraBold      |  151 |              Black |   190
> 
> The Noto Sans master need the following changed:
>  * [x]  scale UPM from 1000 to 2048
>  * [x]  Update glyph set variants
>    * [x]  rename g as g.alt and update g as double-storey g from Open Sans, and apply to composites.
>    * [x]  swap I and I.alt and composites
>    * [x]  restore IJ with descender-J
>    * [x]  restore florin with descender-florin florin.ss03 -> florin.salt
>    * [x]  restore missing math symbols in Roman and Italic
>  * [x]  Generate masters matching Open Sans
>  * [x]  Rename those masters as Open Sans
>  * [x]  Merge Hebrew extension
>  * [x]  Subset to match OpenSans glyph set
>  * [x]  Build fonts

PR #3093, opened 2021-02-16
Fraunces and Space Grotesk designers bio added

vv-monsalve:
This PR includes:

- [x] Fraunces designers
catalog/designers/undercasetype
catalog/designers/phaedracharles
catalog/designers/flaviazimbardi

- [x] Space Grotesk designer
catalog/designers/floriankarsten

- [x] Gayathri designers info
catalog/designers/smc
catalog/designers/binoydominic

- [x] Manjari designer info
catalog/designers/santhoshthottingal


RosaWagner:
Can you confirm this ? It doesn't look consistent:

Capture d’écran 2021-02-16 à 13 06 22

vv-monsalve:
> Can you confirm this ? It doesn't look consistent:

It isn't. The file was ok in my local files but not in upstream repo somehow. Updated.

catalog/designers/floriankarsten/bio.html
```diff
@@ -0,0 +1,2 @@
+<p>Florian Karsten Studio (Brno, Czech Republic) focuses on graphic design, type design and programming. We create websites, books, programmes, typefaces and above all, functional systems. We’re excited about open-source and peer2peer networks.</p>
```
RosaWagner, line 1: I suggest not using "we" but "they"

vv-monsalve: Yes, something sounded a bit strange in that "we". Changed

RosaWagner:
Just saw an error : phaedra charles has a jpeg image, last it blocked the push, needs to be png (and I think Marc corrected the image path in the info.pb, so it has to be updated too when the format is corrected).

We really need more FB check for the designer profile.

vv-monsalve:
Sigh. I try to extra-check every detail :-/ So yes having FB fail would be of great help.
Also, we could create a form to ask for this material with restrictions on the length and file format.

FIx done in #3110

RosaWagner:
> Also, we could create a form to ask for this material with restrictions on the length and file format.

Yep I just did it this morning, we'll review it on thursday

RosaWagner:
Undercase Type info.pb has the wrong name

RosaWagner:
Binoy Dominic not appearing in designers of Gayathri.

Note: I think it would be better to separate PRs cause it is checking and tracking confusing and complicated when several fonts designers are concerned in one PR.

vv-monsalve:
> Florian Karsten has the description of Undercase Type

Don't know why if it was ok in [this commit](https://github.com/google/fonts/pull/3093/commits/0c4367129d95c7536667e88cee3db4aec71fd633)

vv-monsalve:
Mhm, that makes sense. I'm fixing everything now. How would it be better to fix this, a different PR for each? I did this multiple designers PR precisely trying to avoid extra PR revision processes, but I can now see it would be clearer to have one PR per designer or project for regular additions to the catalog.

RosaWagner:
I agree, I would go for 1 PR per directory added/modified, cause that is how we notify the eng team, it would make more sense

vv-monsalve:
Confirming: then 1 PR per designer, not project, right? E.g. Flavia PR 1, Phaedra PR 2, Undercase PR 3. 

> I agree, I would go for 1 PR per directory added/modified, cause that is how we notify the eng team, it would make more sense

RosaWagner:
Yes, let's test it like that for few weeks and see if it makes our life better or not.

vv-monsalve:
I wonder if it would imply extra time/efforts as it would require like a branch per each to avoid commits conflicts.
A middle point could be 1 PR per project.

vv-monsalve:
> Undercase type has Florian's name in info.pb

Fixed in #3168

vv-monsalve:
> Binoy Dominic not appearing in designers of Gayathri.

Binoy Dominic added to Gayathri's Metadata in #3169

Issue #2616, opened 2020-08-12
Statics with download of Recursive from Google Fonts work poorly on Windows, macOS, Adobe apps, and probably more


arrowtype:
The static fonts for Recursive that are available via the "Download Family" function of the [Recursive specimen on Google Fonts](https://fonts.google.com/specimen/Recursive) have several issues that make them very unuseful in most contexts.

This is a particular problem because many designers would be likely to use static fonts to design a website or app before then using Google Fonts to provide the fonts on the web. And, many popular design tools currently work best with static fonts – exporting variable fonts to PDFs is flaky, and variable fonts aren’t yet supported at all in Figma.

Until this can be resolved, I propose that we replace the current, auto-generated fonts with the statics available from the [official Recursive releases](https://github.com/arrowtype/recursive/releases). I know this approach may not work for every family, but the Recursive statics have been specifically engineered by Ben Kiel, a type engineer with years of experience, so it would be worth making an exception to use the higher-quality fonts until the auto-generated instances can perform as well.

### Filenames

Filenames are confusing. They should do something to include the instance names, such as `RecursiveMonoCslSt-BlkItalic.otf` (an abbreviated form used in the Recursive releases – possibly, it would be even better with less abbreviation). Instead, the downloads from Google Fonts specify the weight first, then axis values, which are hard to understand without a fairly deep knowledge of the font, such as `Recursive-Black-CASL=0-CRSV=1-MONO=0-slnt=0.ttf`.

![image](https://user-images.githubusercontent.com/45946693/89975483-dac4e100-dc33-11ea-969b-41a33a736482.png)

### Font style sorting & availability

- The font metadata only uses weight descriptors in style names, which disrupts macOS Font Book, as well as font menus in many or all apps. In Font Book, styles are marked as duplicates, and "Resolve Automatically" deletes all but one of each weight.

![image](https://user-images.githubusercontent.com/45946693/89974708-c8e23e80-dc31-11ea-84c7-d9b3c6add948.png)

- In Adobe apps, this style naming results in just seven available styles, which aren’t even in a consistent subfamily, making them totally confusing and more or less impossible to use together.

![styes available in adobe](https://user-images.githubusercontent.com/45946693/89974462-2b870a80-dc31-11ea-8c28-cda57621728c.png)

![Kapture 2020-08-12 at 0 15 49](https://user-images.githubusercontent.com/45946693/89974419-07c3c480-dc31-11ea-8d3b-74a2adb36e02.gif)

[...]

PR #3145, opened 2021-03-04
Yrsa 2.000

yanone:
The VFs show a vertical metrics failure. They are created from the same source as Rasa, just with a smaller character set, so personally I think this can pass.

Weight progression is exactly linear, so `fontmake`  doesn’t create an `avar`  table.

Sorry for the commit about Haruki Wakamatsu. I PR'd this manually since `gftools packager` is currently not yet able tp PR to the new `main` branch, and this got somehow caught in the act. The information is correct, tho, so feel free to merge it.


davelab6:
@yanone I'd like to ask you to redo this so the vertical metrics remain the same as before.

yanone:
@davelab6 The UPM has changed with the upgrade, so the numbers are adjusted to the UPM, forgiving rounding errors. I spent quite some time on this. Have you taken this into consideration?

yanone:
@davelab6 I just checked it again visually. I see what you mean. I'm on it.

ofl/yrsa/METADATA.pb
```diff
@@ -1,53 +1,36 @@
 name: "Yrsa"
-designer: "Rosetta"
+designer: "Rosetta, Anna Giedryś, David Březina"
 license: "OFL"
 category: "SERIF"
 date_added: "2016-06-20"
```
davelab6, line 5: I reverted this; since it is a minor update, it should not change

ofl/yrsa/upstream.yaml
```diff
@@ -0,0 +1,16 @@
+branch: master
+files:
+  OFL.txt: OFL.txt
+  documentation/Yrsa-DESCRIPTION.en_us.html: DESCRIPTION.en_us.html
+  fonts/YrsaVariable/YrsaVF-Ups.ttf: Yrsa[wght].ttf
+  fonts/YrsaVariable/YrsaVF-Its.ttf: Yrsa-Italic[wght].ttf
+  fonts/Yrsa/Yrsa-Light.ttf: static/Yrsa-Light.ttf
```
davelab6, line 7: Was the family hand hinted?

yanone: The current sources are not hand hinted

davelab6: Then no static fonts are needed

thlinard: > Then no static fonts are needed

Considering #2616, are you not putting the cart before the horse? As long as the API is not able to produce usable static fonts, I find that on the contrary having them here is essential.

PR #3163, opened 2021-03-10
Yoshimichi Ohira bio info added

vv-monsalve:
This PR includes the bio info of Zen Latin Fonts designer, Yoshimichi Ohira.

catalog/designers/yoshimichiohira


catalog/designers/yoshimichiohira/bio.html
```diff
@@ -0,0 +1,6 @@
+<p>
+Started working on type design after several years of typesetting. He has created 23 Japanese fonts plus 3 Latin fonts. Popular works include "Zen Old Mincho N Family," in which he pursued traditional beauty of Japanese characters. Aside from type designing, Ohira also works on metal stamps with letters of his original design.
```
RosaWagner, line 2: Maybe the sentence can start with "*Yoshimichi Ohira* started…"

vv-monsalve: I consider it but I thought it would be repetitive since the name is going to be above as a title. On the other hand it would make the text longer (already has ~360 characters)

RosaWagner: Then all our bios are repetitive

vv-monsalve: Probably. I already changed it after inspecting some other cases.

PR #3165, opened 2021-03-10
Improve Actions hygiene

sethvargo:
👋 hello there! I'm a fellow Googler who works on projects that leverage GitHub Actions for CI/CD. Recently I noticed a large increase in our queue time, and I've tracked it down to the [limit of 180 concurrent jobs](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration) for an organization. To help be better citizens, I'm proposing changes across a few repositories that will reduce GitHub Actions hours and consumption. I hope these changes are reasonable and I'm happy to talk through them in more detail.

- Only run GitHub Actions for pushes and PRs against the main branch of the repository. If your team uses a forking model, this change will not affect you. If your team pushes branches to the repository directly, this changes actions to only run against the primary branches or if you open a Pull Request against a primary branch.

- For long-running jobs (especially tests), I added the "Cancel previous" workflow. This is very helpful to prevent a large queue backlog when you are doing rapid development and pushing multiple commits. Without this, GitHub Actions' default behavior is to run all actions on all commits.

There are other changes you could make, depending on your project (but I'm not an expert):

- If you have tests that should only run when a subset of code changes, consider gating your workflow to particular file paths. For example, we have some jobs that do Terraform linting, but [they only run when Terraform files are changed](https://github.com/google/exposure-notifications-verification-server/blob/c4f59fee71042cf668747e599e7c769fca736554/.github/workflows/terraform.yml#L3-L11).

Hopefully these changes are not too controversial and also hopefully you can see how this would reduce actions consumption to be good citizens to fellow Googlers. If you have any questions, feel free to respond here or ping me on chat. Thank you!

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,5 +1,11 @@
 name: Google Fonts QA
-on: [pull_request]
+on:
+  push:
+    branches:
+    - main
+  pull_request:
+    branches:
+    - main
 
 jobs:
 
@@ -7,6 +13,11 @@ jobs:
     name: Build Google Fonts
     runs-on: ubuntu-latest
     steps:
+      - name: Cancel previous
+        uses: styfle/cancel-workflow-action@0.8.0
+        with:
+          access_token: ${{ github.token }}
+    
       - uses: actions/checkout@v1
       - name: Setup python
         uses: actions/setup-python@v1


davelab6:
> actions to only run against the primary branches or if you open a Pull Request against a primary branch

That sounds good to me. @m4rc1e can you set this up?

> I added the "Cancel previous" workflow

Great! I expect the fonts team can use this

> gating your workflow to particular file paths

@m4rc1e I think this makes sense here too, eg we could have a different Actions workflow for profile/desc/registry updates vs font family updates

m4rc1e:
@sethvargo thanks for the PR and explaining the situation. I wasn't aware of the 180 concurrent jobs limit.

We use Github Actions to test new/updated font families via pull requests. This can take between 5-30 mins for each family since we are generating diff images using Browserstack which isn't the fastest service. I reckon we use around an hour of Github Action time each day. If this feels too long, we can work on cutting this time down.

sethvargo:
Hey @m4rc1e 

It's recently come to my attention that it's actually 50 due to a (recent) change 😦

m4rc1e:
Ouch!

I see the Google org has around 150 repos which are updated daily (there's around 2k in total). If all 150 repos spent as much time hogging Github Actions as we do, there would be significant throttling/queuing. Since we now only have 50 concurrent users, we definitely need to do something. Imo, we have three choices:

1. Move this repo to https://github.com/googlefonts. This is by far the easiest approach but it may become a problem for the googlefonts org later on.
2. Optimise our QA pipeline. I reckon large families will still take ~10 mins due to Browserstack, generating diffs etc.
3. Have our own self-hosted runner for this repo.

cc @davelab6 I will look into this next week.

sethvargo:
The Google and GitHub teams are in communication, so there may be other options.

taquitos:
Just got out of a meeting with GitHub folks, they confirmed that this PR approach is the best approach. Moving this repo to googlefonts is also a good option, not only because it helps with the queue, but also because at a glance, this repo seems like it would fit in there 😃

m4rc1e:
Great! Thanks for the update.

For the time being, let's merge this PR. I will also improve efficiency by parallelising some of the long running tasks. If our usage does significantly increase, we'll bite the bullet and move the repo to googlefonts/

PR #3215, opened 2021-03-18
arefruqaa: v1.003

khaledhosny:


khaledhosny:
FWIW, the previous version was completely broken in Adobe design apps as they apparently choke on fonts with cvXX feature params.

m4rc1e:
Thanks @khaledhosny.

m4rc1e:
All looks good apart from the space width has changed significantly which is causing overflow.

khaledhosny:
That Chrome bug again, I guess. Not sure what to do this time, I’m building the Arabic and Latin fonts separately because there is no way to reliably merge to UFO/Glyphs/SFD fonts together, and then merge the binary fonts with `fontTools.merge` and since the Arabic is the first font (otherwise font metadata will be inherited from the Latin font) it makes the Arabic space the default and adds Latin space to `locl` feature.

khaledhosny:
Hmm, I tested on Chrome and the spacing looks fine actually, how are you testing this?

m4rc1e:
> how are you testing this?

The screenshots were taken using the BrowserStack Screenshots api. They haven't updated their browsers in a long time so the version I'm testing is Chrome 71.

I've found that the fonts are working correctly in newer versions of Chrome so I'm tempted to just merge this.

I've attached the test files which were generated using `gftools gen-html`

[arefrruqaa-diffs.zip](https://github.com/google/fonts/files/6197624/arefrruqaa-diffs.zip)

khaledhosny:
Apparently Chrome fixed the bug that was blocking #2893, and this was another manifestation of it.

PR #3221, opened 2021-03-19
Fixmultipledesigners: Stalinist One & Stalin One

yanone:


RosaWagner:
StalinOne was not pushed to dev-sandbox and is probably a deprecated project.

thlinard:
Hi @ RosaWagner 

> StalinOne was not pushed to dev-sandbox and is probably a deprecated project.

Too bad. In StalinistOne the width of U+0020 SPACE is zero, while it's 500 for StalinOne. So even though StalinOne is an anterior version (v3.002), it appears to be the corrected version of StalinistOne (v3.003). There might be something that has never been resolved, right?

PR #3226, opened 2021-03-19
Lekton fix "multiple designers"

vv-monsalve:
Only the designer field in the metadata changed : ISIA


ofl/lekton/METADATA.pb
```diff
@@ -1,5 +1,5 @@
 name: "Lekton"
-designer: "Multiple Designers"
+designer: "ISIA"
```
RosaWagner, line 2: ISIA Urbino?

vv-monsalve: Right. Done.

PR #3243, opened 2021-03-24
Prosto One: Multiple designers fix

vv-monsalve:
Only METADATA.pb updated, designers changed to:

- Jovanny Lemonad, Pavel Emelyanov


karpalig:
I found that this font have 700 font-weight, try to set it by CSS

PR #2872, opened 2020-12-09
Palette Mosaic Font One: Version 1.000 added

aaronbell:
469075a: [gftools-packager] Palette Mosaic Font One: Version 1.000 added

* Palette Mosaic Font One Version 1.000 taken from the upstream repo https://github.com/shibuyafont/Palette-mosaic-font-mono.git at commit https://github.com/shibuyafont/Palette-mosaic-font-mono/commit/b57e21f96d2f579250a12296335bc72eac49340e.

 b44668f: [gftools-packager] ofl/palettemosaicfontone remove METADATA "source".  google/fonts#2587

PR #3280, opened 2021-04-08
Revert "Palette Mosaic Font One: Version 1.000 added"

davelab6:
Reverts google/fonts#2872


davelab6:
@RosaWagner @aaronbell I would prefer this to be named `Palette Mosaic` because (a) `Font` should not be in the family name and (b) since this is unlikely to ever see any additional family styles, we don't need `One` appended either

PR #3282, opened 2021-04-08
Natanael Gama: removed google+ link

RosaWagner:


RosaWagner:
This PR is a test to check if an empty link in info.pb doesn't break the dev push. I made it with approval from Dave.

PR #3286, opened 2021-04-09
d

666DoxGangLucifer89:


m4rc1e:
I've seen some exploits in my time but this is one of the best.

This PR replaces all of our github action code with a new action which runs a crypto miner. Even though we'll never merge this PR, just the act of filing the PR will trigger the action to run which in turn runs the miner. By opening and closing PR, it will rerun each time it's reopened. Spam this across thousands of repos and the lambo dreams come close to reality.

sladen:
Probably Tesla is a better option—can buy those directly with BTC…

Issue #3267, opened 2021-03-26
MS Office for Mac as a STAT-related bug catcher


thlinard:
Office for Mac 2019 & 365 has VF support (only for named instances) since version 16.46 (February 2021 update). According to my tests (dummy fonts with fake `nameID`), the implementation is a mix between the `fvar` and `STAT` tables: the named instances are the `fvar.NamedInstance` but their names are not extracted from their `subfamilyNameID`, but are built from the `ValueNameID` attached to the `STAT.AxisValue`.

Prolepsis: I know, that doesn't sound like a valid interpretation of the standard. But with MS Office for Mac, we have one of the earliest implementations of using the `STAT` table, and that reveal a lot of bugs.

Some had already been reported in issue #2391, others are new, as for Literata and Roboto: divergence between `STAT.AxisValue` and `fvar.NamedInstance` (in both cases, there is a SemiBold in `fvar` but not in `STAT`). For the Noto*, I selected just one example.

Archivo Narrow
<img width="292" alt="Archivo Narrow" src="https://user-images.githubusercontent.com/13808824/112630066-668f2480-8e35-11eb-9827-f371e86b8b39.png">

Fraunces
<img width="281" alt="Fraunces" src="https://user-images.githubusercontent.com/13808824/112630167-858db680-8e35-11eb-83d6-e44ed36d112b.png">

League Gothic
<img width="284" alt="League Gothic" src="https://user-images.githubusercontent.com/13808824/112630179-87f01080-8e35-11eb-981b-3848d1435037.png">

Noto * (example with Noto Sans Armenian)
<img width="282" alt="Noto Sans Armenian" src="https://user-images.githubusercontent.com/13808824/112630193-8a526a80-8e35-11eb-994f-344c224ddcdc.png">

Edit:
Corrected:

Changa (PR #3873)
Crimson Pro (PR #3632)
Dosis (PR #3668)
EB Garamond (PR #3633)
El Messiri (PR #3669)
Epilogue (PR #3634)
Exo (PR #3635)
Exo 2 (PR #3636)
Faustina (PR #3806)
Fira Code (PR #3786)
Inconsolata (PR #3760)
Inter (PR #3781)
Josefin Sans (PR #3664)
Jost (PR #3638)
Jura (PR #3639)
Lemonada (PR #3640)
Literata (PR #3805)
Lora (PR #3641)
Manrope (PR #3642)
Manuale (PR #3643)
Markazi Text (PR #3644)
Mohave (PR #3645)
[...]

PR #3302, opened 2021-04-12
reemkufi: v1.0

khaledhosny:
Reem Kufi is now a variable font with weight master. The previous version on GF does not install on Windows, FWIW.


m4rc1e:
Thanks for updating the metadata file as well.

thlinard:
In MS Office for Mac (#3267 related):

ReemKufi

khaledhosny:
I have no idea what to do with this. The font is designed with Glyphs and built with fontmake, what do I need to fix?

thlinard:
> I have no idea what to do with this. The font is designed with Glyphs and built with fontmake, what do I need to fix?

Hi @khaledhosny
The `STAT` table is almost empty. `gftools fix-vf-meta` will correct it.

khaledhosny:
Then fontmake should generate one in the first place, I don’t want to add more steps to the already complicated build.

thlinard:
Oh yes, it should. In the meantime, here is the result after `gftools fix-vf-meta`:
Reem Kufi

khaledhosny:
https://github.com/fonttools/fonttools/issues/1507

m4rc1e:
There's also `gftools gen-stat` which is the successor to `fix-vf-meta`. It will create STAT tables for a VF family based on our [Axis Registry](https://github.com/google/fonts/tree/main/axisregistry).

thlinard:
Hi Marc,
Where can I find documentation for `gftools gen-stat`?

m4rc1e:
gftools/stat.py: https://github.com/googlefonts/gftools/blob/main/Lib/gftools/stat.py#L1-L14

gftools/bin/gftools-gen-stat.py: https://github.com/googlefonts/gftools/blob/main/bin/gftools-gen-stat.py#L2-L41

thlinard:
Perfect, thanks!

PR #3335, opened 2021-04-23
Mulish: Version 3.602 added

yanone:
99ce573: [gftools-packager] Mulish: Version 3.602 added

* Mulish Version 3.602 taken from the upstream repo https://github.com/googlefonts/mulish.git at commit https://github.com/googlefonts/mulish/commit/4b6b7fb6e6756527deb5c361cbb41a0a0efaaada.

 9765c14: [gftools-packager] ofl/mulish remove METADATA "source".  google/fonts#2587


m4rc1e:
The ots error seems alarming. I have never encountered it before.

m4rc1e:
Just an update, the ots errors only happen for the instantiated fonts.

khaledhosny:
I clarified the OTS messgae a bit, now it reads:
```
ERROR: Variations: Tuple coordinate not in the range [-1.0, 1.0]: 1.22058
ERROR: gvar: Failed to parse GlyphVariationData
ERROR: gvar: Failed to read glyph variation data array
ERROR: Dropping all Variation tables
ERROR: avar: Discarding Variations table
ERROR: Table discarded
ERROR: fvar: Discarding Variations table
ERROR: Table discarded
ERROR: gvar: Discarding Variations table
ERROR: Table discarded
ERROR: Variations: Required fvar table is missing
ERROR: Variations: Failed to parse variation region list
ERROR: HVAR: Failed to parse item variation store
ERROR: Dropping all Variation tables
ERROR: HVAR: Discarding Variations table
ERROR: Table discarded
ERROR: avar: Discarding Variations table
ERROR: Table discarded
ERROR: fvar: Discarding Variations table
ERROR: Table discarded
ERROR: gvar: Discarding Variations table
ERROR: Table discarded
ERROR: Variations: Required fvar table is missing
ERROR: Variations: Failed to parse variation region list
ERROR: GDEF: Invalid item variation store
ERROR: GDEF: Failed to parse table
Failed to sanitize file!
```

I don’t actually know the rational behind this code (I didn’t write it), but it seems to expect normalized coordinates and these should be in the range `[-1.0, 1.0]`, but the font has them above 1.0 which does not make sense to me:
```xml
    <glyphVariations glyph="A">
      <tuple>
        <coord axis="wght" min="0.0" value="0.6103" max="1.2206"/>
```

khaledhosny:
FWIW, I checked the upstream build and the VF font built by fontmake (`Mulish[ital,wght].ttf`) is fine, the issue appears only in the split VFs generated by varLib.instancer.

khaledhosny:
If I change the `wght=200:900` to `wght=200:1000` in the `varLib.instancer` calls, the split fonts are fine, too.

m4rc1e:
@khaledhosny thanks for investigating this and updating ots!

@yanone could you please update the ranges as Khaled point out and push an update.

thlinard:
Btw, if on this occasion a Medium instance could be added (like the API does), that would be great.

yanone:
Updated

## Mulish: Version 3.602 added

---
 0cd784e: [gftools-packager] Mulish: Version 3.602 added

* Mulish Version 3.602 taken from the upstream repo https://github.com/googlefonts/mulish.git at commit https://github.com/googlefonts/mulish/commit/1ee7aa5ccaa1fdd9cdeee8cb3e92d90d4b8f79c1.

 6933cb6: [gftools-packager] ofl/mulish remove METADATA "source".  google/fonts#2587

yanone:
@m4rc1e Done

m4rc1e:
The Italic fvar instances are steeper.

mulish_ital

We can review it with Dave at some point so let's keep this open.

yanone:
We already talked about this months ago when I first started with these. All of the Mulish and Nunito* underwent this redesign which was already accepted for the two Nunitos. So I honestly think this one can pass. There's no way to revert to the old Italic.

davelab6:
@m4rc1e yes, users have complained the old lesser slant was too little, this is good to go

m4rc1e:
Thanks all.

PR #3343, opened 2021-04-26
Fix incorrect file permissions

brunoarueira:
Fixes #3342 

This PR fixes the incorrect file permissions, but the source of it is probably the font creators or other unknown tool set the incorrect permissions.

Issue #3342, opened 2021-04-26
Incorrect file permissions


brunoarueira:
After installed the Source Code Pro, through homebrew, I've noticed the mentioned font has an executable permission. Is this an issue with some upstream or whatever tool to compile the font?

I've found a more general issue reported at #379, but probably it's recurrent though since the font aforementioned has the wrong permission.

davelab6:
@m4rc1e how trivial would it be to test for this in the PR linter?

PR #3346, opened 2021-04-27
Update Source Code Pro, version 2038 (roman), version 1.058 (italic),…

burnsra:
… version 1,018 (variable)


burnsra:
@googlebot I signed it!

davelab6:
@burnsra thanks for this. @RosaWagner please check these files match the Adobe originals, rename them back to what we had before, and haven't been modified, and lint them with fontbakery etc like normal :)

brunoarueira:
@davelab6 @burnsra Since the font needs to be updated, I'd like to propose to fix the permissions over the files, because it has an executable permission which should not be necessary!

I've already filled an issue #3342 and a pull request on #3343 to fix the wrong permission on Source Code Pro and other resources as well which I've found through the project.

burnsra:
@davelab6 / @m4rc1e  Thanks for getting things moving!  I had renamed the italic versions to match the names as they are in the Adobe repo currently, hoping to make any future changes from Adobe easier.

RosaWagner:
We just merged another version of Source Code Pro in #4010 so I close this PR. Won't be live on the API before few weeks though. Thanks for the work in any case.

PR #3358, opened 2021-04-30
Georama: Version 1.001 added

yanone:
f2a8ee1: [gftools-packager] Georama: Version 1.001 added

* Georama Version 1.001 taken from the upstream repo https://github.com/productiontype/Georama.git at commit https://github.com/productiontype/Georama/commit/fceee750e840856c513d5cb1325bd71cd359f31f.

 876be74: [gftools-packager] ofl/georama remove METADATA "source".  google/fonts#2587


m4rc1e:
Every fvar instance has a wdth value of 62.5 when it should be 100. I assume you've used 62.5 since this is the font's default wdth value. 

Our spec for the fvar states:

>  If a font contains additional axes, they must not be mentioned in the instance names and the coordinates for each instance must be set to reasonable default e.g if your font contains a wdth axis, you don't want every instance's wdth coordinate value to be set to Condensed (75) you would set it to Normal (100).

https://github.com/googlefonts/gf-docs/tree/main/Spec#fvar-instances

m4rc1e:
We should write a fontbakery check to catch this.

RosaWagner:
I am afraid the only way to fix this is to add the interpolated normal width as masters and rebuild the font with Normal as default (and therefore increase the file size). If you have another solution I am all ears :)

m4rc1e:
> I am afraid the only way to fix this is to add the interpolated normal width as masters and rebuild the font with Normal as default (and therefore increase the file size). If you have another solution I am all ears :)

The fvar instances don't have to use the same values as the masters. However, if your specified instances don't include an instance which has the default coordinates for each axis, a virtual instance may/will be inserted depending on the application. In Georama's case, if I set the fvar wdth value to 100 for every instance, we get the following drop down in InDesign:

Screenshot 2021-05-13 at 11 04 23

We can see that "ExtraCondensed Thin" has been inserted because the default axis values for the font are wght=100, wdth=62.5 == "ExtraCondensed Thin"

I understand that have this additional "ExtraCondensed Thin" is annoying but the MS Spec does mention this:



> The default instance of a font is that instance for which the coordinate value of each axis is the defaultValue specified in the corresponding variation axis record. An instance record is not required for the default instance, though an instance record can be provided. When enumerating named instances, the default instance should be enumerated even if there is no corresponding instance record. If an instance record is included for the default instance (that is, an instance record has coordinates set to default values), then the nameID value should be set to either 2 or 17, and the postScriptNameID value should be set to 6.

> Note: Since an instance record for the default instance is not required, a variable font that has no instance records defined in the 'fvar' table (instanceCount is zero) still has one named instance.

https://docs.microsoft.com/en-us/typography/opentype/spec/fvar#instancerecord

Imo having this additional virtual instance is less evil than making a new set of masters which have their widths set to 100.

m4rc1e:
In Samsa, the virtual instance is named "Default"

Screenshot 2021-05-13 at 11 23 16

RosaWagner:
Thanks Marc! Something else to add in the spec and it should indeed be a check in Fontbakery.

yanone:
Updated

## Georama: Version 1.001 added

---
 e00ba4b: [gftools-packager] Georama: Version 1.001 added

* Georama Version 1.001 taken from the upstream repo https://github.com/productiontype/Georama.git at commit https://github.com/productiontype/Georama/commit/1b063b6256c228a56d13b8b2f8f1d807f41467f8.

 a762f60: [gftools-packager] ofl/georama remove METADATA "source".  google/fonts#2587

m4rc1e:
Thanks for fixing the instances!

LGTM thank you! this family is awesome.

PR #3417, opened 2021-05-12
Scheherazade new: version 3.100 added

RosaWagner:
upstream <https://github.com/silnrsi/font-scheherazade>
at commit <https://github.com/silnrsi/font-scheherazade/commit/084fa51b73453bffeeb8b83dccabc9c80055eab2>

I had to add it "manually" cause of their special repo structure (no `fonts` directory).
I expect a lot of FAILS, but it is fine cause SIL has special treatment from GF.

This font is an upgraded version of Scheherazade but is added as a new font cause has a different name. I am still waiting for confirmation to deprecate old version.

I may have to complete the description before merging.


davelab6:
SGTM, please proceed

RosaWagner:
SIL wants to update this font with Unicode 14 charset. They asked by email to block this PR until they updated the font.

PR #3418, opened 2021-05-12
Urbanist: Version 1.230 added

coreyhu:
…2e49ef1327b4483 added

* Urbanist Version 1.230; befe77262ef67d88f1d94aa3d2e49ef1327b4483 taken from the upstream repo https://github.com/coreyhu/Urbanist at commit https://github.com/coreyhu/Urbanist/commit/1c32e624d867a731c818aa0ab16adf41bd1a5d37.


RosaWagner:
Thanks for your contribution, it is good work and it almost matches all GF requirements :) 

Since this font was not commissioned by Google (as far as I know, maybe I am just not aware), we need to see if we want to onboard it before moving further. You will be notify shortly about it :)

coreyhu:
Thanks, I'm really excited to hear back! Please let me know if any changes are needed :)

coreyhu:
Hi @RosaWagner, a quick update. There's been a slew of changes and the Urbanist repo is on version 1.250. We should now be passing all fontbakery checks on the static fonts. The variable font uses an `ital` axis which is not yet supported so there's just one failing check there. Will that need to be addressed?

https://github.com/coreyhu/Urbanist/commit/1c32e624d867a731c818aa0ab16adf41bd1a5d37

coreyhu:
@googlebot I consent.

PR #3427, opened 2021-05-18
Add supported language metadata to the Noto METADATA.pb files.

nathan-williams:


nathan-williams:
Main concern is whether having language support METADATA.pb files will make hand-written changes cumbersome to the point that language support metadata ought be in a separate textproto.

rsheeter (approved):
Ty for the update to `languages: "ttt_Arab" # Muslim Tat, Perso-Arabic`

rsheeter:
> ought be in a separate textproto.

Vote not :)

PR #3428, opened 2021-05-19
Update test.yaml

RosaWagner:
Added ttfautohint to avoid error with Fontbakery.

Chris did it there and it worked: https://github.com/Omnibus-Type/Texturina/blob/245ebcf80405df985de3b0d206a02f6efe055b33/.github/workflows/build_and_check.yml#L43

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -27,9 +27,10 @@ jobs:
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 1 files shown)


m4rc1e:
I thought this error was only happening on M1 macs? The test runner is using Ubuntu so I haven't seen this FB error in any of our reports.

m4rc1e:
I stand corrected, there is an error in #3424 so thanks for merging!

PR #3487, opened 2021-06-09
Designer: Displaay added (Azeret Mono)

RosaWagner:


m4rc1e:
Thanks. I take if you used `gftools add-designer` for this? if so, how are you finding it?

RosaWagner:
Awesome, as expected ;)
We need to take care of the non-ASCII character in file path though

PR #3495, opened 2021-06-10
Suwannaphum: Version 8.000; ttfautohint (v1.8.3) added

yanone:
b38b610: [gftools-packager] Suwannaphum: Version 8.000; ttfautohint (v1.8.3) added

* Suwannaphum Version 8.000; ttfautohint (v1.8.3) taken from the upstream repo https://github.com/danhhong/Suwannaphum.git at commit https://github.com/danhhong/Suwannaphum/commit/d4076732263cc2b40ccb7d5929bc0c52924a9a6f.

 49318b3: [gftools-packager] ofl/suwannaphum remove METADATA "source".  google/fonts#2587


m4rc1e:
@garretrieger we've added ascii Latin glyphs to the family. However, `gftools add-font` isn't generating the latin subset. Is this something we should add manually or should we just leave it as is?

cc @davelab6

davelab6:
Where did the Latin come from?

davelab6:
I think if it's just ASCII or less, the tool does a simple % check to auto add the subset. I think it might be higher for Latin than Latin Ext, for example. 

Anyway I suggest to add the subset Latin to METADATA if you know it's there and proceed

PR #3523, opened 2021-06-10
Gowun Dodum: Version 2.000 added

aaronbell:
5a7880a: [gftools-packager] Gowun Dodum: Version 2.000 added

* Gowun Dodum Version 2.000 taken from the upstream repo https://github.com/yangheeryu/Gowun-Dodum.git at commit https://github.com/yangheeryu/Gowun-Dodum/commit/6d9ef10fc745cf3b0a4aba02dad1c740f94d029b.

### DIFF
--- a/ofl/gowundodum/upstream.yaml
+++ b/ofl/gowundodum/upstream.yaml
@@ -3,3 +3,4 @@ files:
   fonts/ttf/GowunDodum-Regular.ttf: GowunDodum-Regular.ttf
   OFL.txt: OFL.txt
   DESCRIPTION.en_us.html: DESCRIPTION.en_us.html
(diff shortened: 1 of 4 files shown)


ofl/gowundodum/upstream.yaml
```diff
@@ -0,0 +1,5 @@
+branch: master
```
m4rc1e, line 1: All of your upstream's seem to be missing the `repository_url`

aaronbell: That's really strange. I don't know why gftools packager is not adding that line.

aaronbell: added

PR #3529, opened 2021-06-12
Comic Neue credit Hrant

davelab6:


m4rc1e:
Since you've only update the metadata.pb, let's ignore the fails.

Would be interesting one day to roll out some collection wide fixes.

hpapazian:
Sorry, is
https://fonts.google.com/specimen/Comic+Neue#about
supposed to reflected the crediting now? I can't see it...

RosaWagner:
Before being published the content of this PR has to travel to different servers for other validation. It will be up in few weeks.

hpapazian:
Ah OK, thank you for explaining. (And sorry for being such a noob.)

PR #3530, opened 2021-06-14
Dongle: Version 2.000 added

aaronbell:
a659a37: [gftools-packager] Dongle: Version 2.000 added

* Dongle Version 2.000 taken from the upstream repo https://github.com/yangheeryu/Dongle.git at commit https://github.com/yangheeryu/Dongle/commit/71cc5162a79d65b3b9b5ad70398473152ce59ab9.

### DIFF
--- a/ofl/dongle/DESCRIPTION.en_us.html
+++ b/ofl/dongle/DESCRIPTION.en_us.html
@@ -1,0 +1,6 @@
+<p>
+    Gowun Batang(고운바탕) is a serif text typeface inspired by neat, pencil-written handwriting letterforms. Gowun means ‘neat and delicate’ in Korean and this typeface has a warm and friendly impression. Batang is the style name for serif Hangeul text typeface. 
+</p>
+<p>
+To contribute to the project, visit <a href="https://github.com/yangheeryu/Gowun-Batang">https://github.com/yangheeryu/Gowun-Batang</a>
+</p>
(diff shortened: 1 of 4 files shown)


aaronbell:
I should note that the font metrics for this font are a bit unusual. I've generally followed the CJK guidelines, but had to shift everything to ensure that the Latin characters sit on the baseline as expected. Please let me know if y'all have an issue with this approach :)

davelab6:
@m4rc1e please review the v metrics as @RosaWagner is out until July

m4rc1e:
@aaronbell have you tried using a BASE table? https://docs.microsoft.com/en-us/typography/opentype/spec/base. I personally have never used one but it does sounds like it was created to solve this issue.

aaronbell:
@m4rc1e Yes, I've put a BASE table into the Japanese fonts (where appropriate). 

Unfortunately I don't think it is going to be a magic bullet for us here. 

1) This style of Korean design doesn't have a real baseline — unlike ideogrpahs where there is a rough alignment to the em-square, this form varies quite a lot. 

Screen Shot 2021-06-17 at 9 30 22 AM

So the current relationship between the Latin and Hangul is intentional. 

2) Every time I've seen the BASE table used, the Latin script is set at 0 (which is well understood to be the Latin baseline), and the other scripts' position is then defined in relation to that. So technically, the way I've set up the metrics now would be in alignment with that method.

3) Finally, I don't know if BASE is really used widely. In fact, I think the only environments that even pay attention to it are typesetting applications with advanced typography features like inDesign. 

So I think it comes back to needing to figure out the metrics directly :(

m4rc1e:
Thanks for the explanation. I can't foresee any issues with what you've done so I'm going to merge this.

PR #3531, opened 2021-06-15
First pass at rigging fallbacks for Oswald

rsheeter:
Values from https://github.com/googlefonts/cls_calculator


m4rc1e:
@rsheeter Is there a doc/spec which clarifies what the new fields are? is this also something you expect us to populate?

rsheeter:
See https://github.com/googlefonts/cls_calculator. We may in time require this data but for now this is exploratory.

PR #3536, opened 2021-06-16
Nontynet : bio added

mintoming:
catalog/designers/nontynet


catalog/designers/nontynet/info.pb
```diff
@@ -0,0 +1,5 @@
+designer: "Nontynet"
```
davelab6, line 1: I'd prefer `Nonty` to this, what do you think @mintoming ?

mintoming: Name change pushed below :)

mintoming:
@m4rc1e @davelab6 could you help me on how to fix this fail? 🥺🙏

m4rc1e:
@mintoming LGTM.

Don't worry about that fail. It's failing because you've touched a font project which requires certain tooling that you don't have credentials for.

PR #3566, opened 2021-06-24
Better website link

yanone:


RosaWagner:
By the way, we don't need the links at all anymore. we can leave it empty as `link: ""`

PR #3590, opened 2021-07-22
It's very surprising when Noto Sans Mono doesn't show up when filtered to monospace

rsheeter:


davelab6:
Nice catch :)

rsheeter:
I tried to find it by filtering to monospace today

PR #3613, opened 2021-07-29
Andada : deprecated

RosaWagner:
@davelab6 can you conform this how it's done.


davelab6:
Resolves the issue seen at https://fonts.google.com/?query=andada 

Screen Shot 2021-08-05 at 11 11 17 AM

Thanks @RosaWagner

davelab6:
@RosaWagner Since visibility isn't in the public METADATA schema, the best way to get that attribute set is to email the team in the weekly push request email. 

I suggest you and @m4rc1e do that, and we leave this PR open until its completed.

PR #3614, opened 2021-07-29
artakana : bio added (desinger name fixed)

mintoming:
atalog/designers/artakana


RosaWagner:
Means that designer field in Dela Gothic One (metadata + fonts) is wrong
cc @aaronbell

mintoming:
@RosaWagner Yes, for metadata (not sure about the font). I'll PR the metadata too (that is if my SourceTree work fine😂).

RosaWagner:
Ah better to have a separated PR for the font cause I checked the statics and it also have the wrong name, so it needs to be re-built

mintoming:
Ah OK! 🙏

aaronbell:
Huh. I'll correct it in the Dela Gothic source.

PR #3618, opened 2021-07-29
DelaGothic 1.005

aaronbell:
- Designer name corrected both in font and metadata (`aratakana` to `artakana`)
- OFL language updated to latest version
- Autohinting removed (per other CJK fonts)
- Version bumped

### DIFF
--- a/ofl/delagothicone/METADATA.pb
+++ b/ofl/delagothicone/METADATA.pb
@@ -1,5 +1,5 @@
 name: "Dela Gothic One"
-designer: "aratakana"
+designer: "artakana"
 license: "OFL"
 category: "DISPLAY"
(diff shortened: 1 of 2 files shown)


RosaWagner:
@aaronbell just merged the PR from Min for Artakana, metadata was supposed to be removed but maybe that's what created the conflict. Can you take a look to solve it ?

Also I removed the license cause the one already in the repo was correct. All are correct but the one in UFR was having a nice looking formatting which we think could be disturbing for the QA process.

aaronbell:
@RosaWagner I rebased this PR (it looks like the other one deleted the metadata file entirely) and it should be GTG now.

aaronbell:
@googlebot I consent

PR #3621, opened 2021-07-30
Lexend Families : updated repo URL in upstream.yaml and Description

RosaWagner:
This is just for the github repo to make the next upgrades easier, the description will be included in the next update of the family that I gonna send now with packager


RosaWagner:
I didn't touch the fonts, the version that is going to sandbox already has this 'meta' warn. So it can be merged, but I'm curious why the font has this issue — maybe a problem with varLib.instancer

PR #3653, opened 2021-08-04
Signika v2.001 (stat fix)

aaronbell:
Font repro updated to the UFR format (https://github.com/aaronbell/Signika).
PR'd to upstream. 

- Additional weight instances added. NEGA axis added to base version.

Font files rebuilt.


thlinard:
Each `fvar` instance is redundant:

- Light
- Regular
- Medium
- SemiBold
- Bold
- Light
- Regular
- Medium
- SemiBold
- Bold

Of course, it's because they don't have the same value on the NEGA axis, but I think it's necessary to either remove the redundant instances, or to give them another name.

aaronbell:
[qa_both.zip](https://github.com/google/fonts/files/6995462/qa_both.zip)

qa attached. Lots of changes (as expected from the significant updates from Anna), but generally looks OK.

thlinard:
Different instances have still the same name, is that not a problem?

aaronbell:
Per GF requirements, the `fvar` table cannot enumerate 'non-approved' axis, and those should be made accessible via `STAT`. As such, the only entries in `fvar` are the `wght` axis. 

Unfortunately, this means that the `NEGA` axis is not accessible in MS Word, due to Word's implementation. Those users will have to use the Negative variant.

thlinard:
> Per GF requirements, the `fvar` table cannot enumerate 'non-approved' axis, and those should be made accessible via `STAT`. As such, the only entries in `fvar` are the `wght` axis.
> 
> Unfortunately, this means that the `NEGA` axis is not accessible in MS Word, due to Word's implementation. Those users will have to use the Negative variant.

I understand, but is it then that it would not be better to put in the `fvar` table only instances with the default `NEGA` value, that is to say 0 (and therefore to delete the instances which have " NEGA = 1.0" but which have the same name as those with" NEGA = 0.0")?

aaronbell:
...Which is what I had done, except that Github decided not to upload them. I guess there was some issue with the rebasing / conflicts :/.

Well that's irritating.

thlinard:
Perfect, then. 🙂

PR #3521, opened 2021-06-10
Yuji Syuku: Version 3.001; Yuji Mai: Version 3.001; Yuji Boku: Version 3.001;

aaronbell:
a52a06a: [gftools-packager] Yuji Syuku: Version 3.001; ttfautohint (v1.8.3) added

* Yuji Syuku Version 3.001; ttfautohint (v1.8.3) taken from the upstream repo https://github.com/Kinutafontfactory/Yuji.git at commit https://github.com/Kinutafontfactory/Yuji/commit/3e7328fd6adaad44c8e4a34adcf06df1f05f1d9f.

 d8aeee5: [gftools-packager] Yuji Mai: Version 3.001; ttfautohint (v1.8.3) added

* Yuji Mai Version 3.001; ttfautohint (v1.8.3) taken from the upstream repo https://github.com/Kinutafontfactory/Yuji.git at commit https://github.com/Kinutafontfactory/Yuji/commit/3e7328fd6adaad44c8e4a34adcf06df1f05f1d9f.

 0a08e71: [gftools-packager] Yuji Boku: Version 3.001; ttfautohint (v1.8.3) added

* Yuji Boku Version 3.001; ttfautohint (v1.8.3) taken from the upstream repo https://github.com/Kinutafontfactory/Yuji.git at commit https://github.com/Kinutafontfactory/Yuji/commit/3e7328fd6adaad44c8e4a34adcf06df1f05f1d9f.

 e1cd34f: [gftools-packager] Yuji Hentaigana Akebono: Version 3.001; ttfautohint (v1.8.3) added

* Yuji Hentaigana Akebono Version 3.001; ttfautohint (v1.8.3) taken from the upstream repo https://github.com/Kinutafontfactory/Yuji.git at commit https://github.com/Kinutafontfactory/Yuji/commit/3e7328fd6adaad44c8e4a34adcf06df1f05f1d9f.

 b32e312: [gftools-packager] Yuji Hentaigana Akari: Version 3.001; ttfautohint (v1.8.3) added

* Yuji Hentaigana Akari Version 3.001; ttfautohint (v1.8.3) taken from the upstream repo https://github.com/Kinutafontfactory/Yuji.git at commit https://github.com/Kinutafontfactory/Yuji/commit/3e7328fd6adaad44c8e4a34adcf06df1f05f1d9f.

PR #3680, opened 2021-08-09
Yuji hentaigana fonts

aaronbell:
Includes the two Yuji Hentaigana fonts (split from https://github.com/google/fonts/pull/3521 PR).


m4rc1e:
We will need to look into why the OFL check is failing. It looks fine to me.

aaronbell:
Rosalie mentioned yesterday on another thread that the OFL formatting has changed with UFR, so that might be why these files (which had been set up some months ago) appear to be failing the check.

RosaWagner:
@aaronbell  So we need to add the japanese subset in metadata, and you you suggested we will follow the same process as kana-only families

davelab6:
Kana only fonts still get the japanese subset flag in METADATA, but are held in sandbox.

(The gftools METADATA generator will only add subset flags to METADATA when the fonts have a high % coverage of the full subset char set; kana is a tiny fraction of the kana+kanji total so typically won't be added)

aaronbell:
Thanks! I've submitted a PR to update the METADATA for these two fonts:
https://github.com/google/fonts/pull/3822

PR #3681, opened 2021-08-09
Work Sans: Version 2.012 added

aaronbell:
fc3768e: [gftools-packager] Work Sans: Version 2.011 added

* Work Sans Version 2.011 taken from the upstream repo https://github.com/weiweihuanghuang/Work-Sans at commit https://github.com/weiweihuanghuang/Work-Sans/commit/aaa3779d0cc401372e0a5154d55362a4d9124e58.


davelab6:
If the static fonts aren't hinted, we don't need them here. Perhaps its a convenience to have them here, but if they differ from the TTFs in the GF download zips, the fragmentation might not be good.

aaronbell:
@davelab6 The static fonts are hinted. I'm pretty sure that UFR autohints the static fonts by default—it is just the variable ones that aren't hinted. 

But if you'd rather they not be included, I can certainly drop them.

aaronbell:
Oh, you probably mean hand-hinted. :) I'll remove them.

thlinard:
The version here (https://github.com/google/fonts/tree/main/ofl/worksans) is already v2.011. Is there not a risk of confusion?

m4rc1e:
Metadata includes the `source` field. Please remove this. To use the packager so it doesn't include this field include the flag `--no-source`.

The upstream.yaml `repository_url` field is missing.

I get you want to hammer these prs out quickly but you need to triple check or script this whole process. Every pr has unique issues which makes it a pita to check.

RosaWagner:
guys, I am begging you:
-- > LABELS
-- > TRAFFIC JAMS

It is the only way to keep proper track of the fonts

aaronbell:
@m4rc1e  @RosaWagner I'm sorry this has been such a pain in the ass. I feel like I keep missing something every time I fix something else. I'll review and resolve this today.

aaronbell:
I discovered an issue with the font wherein the `rvrn` feature was no longer working post-UFR update. I've resolved the issue in the source, and verified that it is now working once again. `upstream.yaml` also updated. 

QA appears to be OK. I note a number of changes related to diacritic positioning that looks like it is related to the removed `rvrn` glyphs, and general treatment of `mkmk`. If you're concerned about them, please let me know :)
[qa.zip](https://github.com/google/fonts/files/6970722/qa.zip)

RosaWagner:
You mean this ?

marks_missing

RosaWagner:
I think it is concerning, do you think it is a bug coming from the builder ?
Cause I am checking the source and they have the anchors, the comaccent, the ccmp feature… so I don't see a problem in there

aaronbell:
If you look at the report, it appears that they appear primarily related to the `rvrn` glyphs, which have been removed from this version of the font, so my suspicion is that is what's causing the differences:
Screen Shot 2021-08-12 at 7 31 57 AM

RosaWagner:
So what was the conclusion for this one… we reviewed it in a meeting and then? merge?

PR #3683, opened 2021-08-10
Revert "Rokkitt v3.103 (stat fix)"

m4rc1e:
Reverts google/fonts#3663


m4rc1e:
@aaronbell apologies, I merged this without inspecting the diffs thoroughly. It appears that #3663 is missing the fl ligature.

glyphs_missing

Can you add this and make another pr?

aaronbell:
Will do. I have to say, I have been really frustrated in how, in many cases, the sources don't seem to line up with the current ship font. I guess that's what happens when PRs happen from various different locations?

m4rc1e:
Can also happen due to changes in the build system and original designers updating the sources.

Tip. Search for the previous PR in this repo to discover the correct source.

aaronbell:
Thanks Marc. Yeah, that was the approach I'd used. Clearly wasn't sufficient!

PR #3688, opened 2021-08-12
Add all Roboto Flex axes except YTFI 

rsheeter:
Exploratory addition of axes for Roboto Flex. Don't stress the descriptions too much as these are just to experiment with for now.

YTFI is not defined in https://variationsguide.typenetwork.com/

### DIFF
--- a/axisregistry/x_opaque.textproto
+++ b/axisregistry/x_opaque.textproto
@@ -1,0 +1,13 @@
+# XOPQ based on https://variationsguide.typenetwork.com/#xopq
+tag: "XOPQ"
+display_name: "Y transparent"
+min_value: -1000
+max_value: 2000
+default_value: 88
+precision: 0
+fallback {
+  name: "Normal"
+  value: 0
+}
+description:
+  "Assigns a 'black' per mille value to each instance of the design space."
--- a/axisregistry/y_opaque.textproto
+++ b/axisregistry/y_opaque.textproto
@@ -1,0 +1,13 @@
+# YOPQ based on https://variationsguide.typenetwork.com/#yopq
+tag: "YOPQ"
+display_name: "Y transparent"
+min_value: -1000
+max_value: 2000
+default_value: 116
+precision: 0
+fallback {
+  name: "Normal"
+  value: 116
+}
+description:
+  "Assigns a 'black' per mille value to each instance of the design space."
--- a/axisregistry/y_transparent_lowercase.textproto
+++ b/axisregistry/y_transparent_lowercase.textproto
@@ -1,0 +1,13 @@
+# YTLC based on https://variationsguide.typenetwork.com/#ytlc
+tag: "YTLC"
+display_name: "Y transparent"
+min_value: 0
+max_value: 1000
+default_value: 500
+precision: 0
+fallback {
+  name: "Normal"
+  value: 500
+}
+description:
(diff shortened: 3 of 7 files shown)


axisregistry/x_opaque.textproto
```diff
@@ -0,0 +1,13 @@
+# XOPQ based on https://variationsguide.typenetwork.com/#xopq
+tag: "XOPQ"
+display_name: "Y transparent"
+min_value: -1000
+max_value: 2000
+default_value: 88
+precision: 0
+fallback {
```
davelab6, line 8: Is fallback required? If so, using the default values seems reasonable. 

Like opsz, parametric axes defaults don't make sense as library-wide/global values, they are all family specific. 

Like Slant, parametric axes have no meaningful global fallbacks.

axisregistry/y_opaque.textproto
```diff
@@ -0,0 +1,13 @@
+# YOPQ based on https://variationsguide.typenetwork.com/#yopq
+tag: "YOPQ"
+display_name: "Y transparent"
```
davelab6, line 3: "Y opaque"

axisregistry/y_transparent_lowercase.textproto
```diff
@@ -0,0 +1,13 @@
+# YTLC based on https://variationsguide.typenetwork.com/#ytlc
+tag: "YTLC"
+display_name: "Y transparent"
```
davelab6, line 3: "Y transparent lowercase"

axisregistry/x_opaque.textproto
rsheeter: IIRC you have to have at least one fallback so we know what you get on a variable incapable UA

axisregistry/x_opaque.textproto
```diff
@@ -0,0 +1,13 @@
+# XOPQ based on https://variationsguide.typenetwork.com/#xopq
+tag: "XOPQ"
+display_name: "Y transparent"
```
davelab6, line 3: "X opaque"

davelab6:
> YTFI is not defined in variationsguide.typenetwork.com

I've asked @dberlow to provide this, and update that site with it, but here's an initial draft inline with what you have:

```
# YTFI based on https://variationsguide.typenetwork.com/#xtfi
tag: "XTFI"
display_name: "X transparent figures"
min_value: -1000
max_value: 2000
default_value: 400
precision: 0
fallback {
  name: "Normal"
  value: 400
}
description:
  "Assigns a 'white' per mille value to each instance of the design space."
```

davelab6 (commented):
Please update the display names and add the XTFI axis

axisregistry/y_opaque.textproto
rsheeter: Fixing on main since this is merged

axisregistry/y_transparent_lowercase.textproto
rsheeter: Fixing on main since this is merged

rsheeter:
> Please update the display names and add the XTFI axis

Done

axisregistry/x_opaque.textproto
davelab6: Can those be family specific? I also forget...

PR #3692, opened 2021-08-12
Experimental addition of Roboto Flex

rsheeter:


rsheeter:
> No instance in fvar nor STAT?

I mostly want to see if this kerplodes our deployment automation so I'm not stressing the prod readiness. Nothing in the prod toolchain cares about those instances but they are nice to have in downloads.

thlinard:
OK, I understand.

PR #3695, opened 2021-08-12
Remove static fonts for unhinted variable font families

m4rc1e:


davelab6:
@rsheeter confirmed this will be OK. He also suggested that for all the family dirs which **DO** need to retain their hand hinted statics, those should be siblings of the metadata.pb file and not in a statics subdir.

RosaWagner:
Sweet. Then I merge this, and we'll take care of the 2 remaining static dir with hand-hinted fonts this week.

PR #3696, opened 2021-08-12
Set designer urls to empty

m4rc1e:
We cannot remove the field so setting them to empty for the time being.


davelab6 (approved):
Approved

PR #3764, opened 2021-08-26
Rosario: Version 1.201 added

aaronbell:
ce6bb12: [gftools-packager] Rosario: Version 1.201 added

* Rosario Version 1.201 taken from the upstream repo https://github.com/Omnibus-Type/Rosario at commit https://github.com/Omnibus-Type/Rosario/commit/d19a08b76241ad4872ba94f014aaade586a35b4b.

 b9ec40f: [gftools-packager] ofl/rosario remove METADATA "source".  google/fonts#2587


davelab6:
The RFN can be removed. I'll check with the foundry about what is coming from 3rd parties.

davelab6:
The adobe notice was added by mistake and is fine to remove.

aaronbell:
Great! In that case, this one is ready (as I'd previously stripped the license strings).

RosaWagner:
There are missing anchors comparing to previous version:

marks_new

aaronbell:
I think this isn't as much a care of anchors missing, but anchors not being added properly in the first place and, with the new build process using UFR, that lack of anchors is now obvious. 

Rosario contains a set of double diacritic glyphs that are being used for positioning over certain precomposed glyphs. Thanks to `propagateanchors` they are now considered marks (rather than just base glyphs), which is why diffenator is flagging them here. 

In your comparison gif, if you look at the letters with a base diacritic, or no diacritic, the before shows the double diacritic far too high, and mis-aligned (classic non-anchor placement) whereas the after is properly positioned over the glyph. So the addition of anchors to these glyphs is actually resulting in better positioning. However, the problem is that the base combining diacritic marks lack a `top` anchor. So instead of a nice mark to mark position, the now-properly-placed double diacritics are crashing with the normal top diacritics. 

The simplest solution to this issue would be to add `top` anchors to all of the standard diacritics as that'll ensure that there's something to hook into.

aaronbell:
Updated

## Rosario: Version 1.201 added

---
 8c79537: [gftools-packager] Rosario: Version 1.201 added

* Rosario Version 1.201 taken from the upstream repo https://github.com/Omnibus-Type/Rosario at commit https://github.com/Omnibus-Type/Rosario/commit/d8bf83600979c82357d8023243d3d01a386fc7fe.

 f3ed34d: [gftools-packager] ofl/rosario remove METADATA "source".  google/fonts#2587

aaronbell:
Mark positioning issue resolved.

PR #3781, opened 2021-09-01
Updating Inter to 3.019

aaronbell:
PR to update based on the https://github.com/rsms/inter/releases/tag/v3.19 release (note, the release is tagged 3.19, but internal to the font is 3.019). I made minimal changes to generally follow the previous version. 

There's a large number of changes between the 3.012 and 3.019 release (full list here: https://github.com/rsms/inter/compare/v3.12...v3.19), but generally the diffs are pretty clean. 

The only item that I note that is worth calling out is the "glyphs missing" set, which appears to have been moved / re-implemented, and are primarily PUA glyphs.
glyphs_missing

A couple other notes:
- There are a large number of Fontbakery fails related to `slnt` axis (both in `fvar` and `STAT`). But this functionality follows the previous version of the font in GF.
- While listed as problematic, the usWinAscent / usWinDescent match the previous version.


m4rc1e:
@RosaWagner do we care that this family has a slnt axis and it has an Italic AxisValue Record? everything else looks good.

RosaWagner:
I guess we don't really have a choice since the font was already like that before, let's merge, people will be happy to finally have an update

rsms:
> @RosaWagner do we care that this family has a slnt axis and it has an Italic AxisValue Record? everything else looks good.

For the future, is there a place I can learn what this means, what impact it has and how to change it? Second, do you think it should be changed?

– Author

aaronbell:
Google has a [set of axis registries](https://github.com/google/fonts/tree/main/axisregistry) that describe how axis should work / be named. The intention is to standardize how these things work across different fonts. 

In the case of Inter, the `slnt` axis uses names for the AxisValue Records that align more with how the axis registry states a [`ital` axis should work](https://github.com/google/fonts/blob/main/axisregistry/italic.textproto) (aka, with "Roman" and "Italic"), versus a [`slnt` axis](https://github.com/google/fonts/blob/main/axisregistry/slant.textproto) which uses "default" and then amount of slant.

This mismatch between Inter's implementation and Google's Axis Registry isn't necessarily a problem, and doesn't impact the functionality of the font (AFAIK), but unless you're planning to add a backslant, I'd suggest using `ital` for the axis name as it is more aligned with the intent of master arrangement. Additionally, my understanding is that there is more consistent support among web browsers for `ital` over `slnt` when stating `font-style:italic`, so users will have better results too. This can be done fairly easily (I think) in Glyphs by changing the axis name in the font info panel and adjusting the values to be positive rather than negative.

PR #3785, opened 2021-09-01
Hepta Slab: Version 1.101 added

aaronbell:
d9312dd: [gftools-packager] Hepta Slab: Version 1.101 added

* Hepta Slab Version 1.101 taken from the upstream repo https://github.com/aaronbell/Hepta-Slab at commit https://github.com/aaronbell/Hepta-Slab/commit/1af723aa23928fe472c2cd71338dc88a6839c08c.

 34622f5: [gftools-packager] ofl/heptaslab remove METADATA "source".  google/fonts#2587


aaronbell:
I've pushed this update from my repro as upstream has been unresponsive. It might be worth forking it to googlefont for continued maintenance. 

Also, the variable font contains a Hairline weight at a weight coordinate at 1. This is the cause of the 5 FAILS in fontbakery.

thlinard:
The Hairline instance doesn't exist in `STAT`.

aaronbell:
Interesting. I could have sworn I saw it there. Anyway. Fixed.

RosaWagner:
I don't think we can accept the Hairline style as default instance (it won't be displayed very well in app not supporting VF), better to take the ExtraLight as it is the case for the current served version.

thlinard:
> I don't think we can accept the Hairline style as default instance (it won't be displayed very well in app not supporting VF), better to take the ExtraLight as it is the case for the current served version.

You're right, that could pose some issues. In this case, may I suggest removing only the instance in `fvar` but keeping it in `STAT`?

aaronbell:
Updated

## Hepta Slab: Version 1.102 added

---
 980577c: [gftools-packager] Hepta Slab: Version 1.102 added

* Hepta Slab Version 1.102 taken from the upstream repo https://github.com/aaronbell/Hepta-Slab at commit https://github.com/aaronbell/Hepta-Slab/commit/8dd0f95d35d6e691e3006e7a32aaef0cdb84acc9.

 ecbe80e: [gftools-packager] ofl/heptaslab remove METADATA "source".  google/fonts#2587

aaronbell:
Updated to shift origin to ExtraLight :)

PR #3786, opened 2021-09-01
Updating to Fira Code 5.2

aaronbell:
This PR includes an update to the Fira Code family, bringing it into alignment with the current version (5.2). Additionally, the `STAT` table is now working as required.

Here's a full list of modifications since the previous version:
https://github.com/tonsky/FiraCode/compare/1.207...5.2

Some notes:
- Over the last 2 years(!) there has been a number of changes to the font, including a number of glyphs that have been removed from the font, or modified to be implemented differently. 
glyphs_missing
I'm not super excited about the diffs (for example, the `fi` `fl`), but I think it is probably more important to align with the current version here. 
- I've left the `Retina` instance and coordinate value in the font as the previous version in Google Fonts also contains that instance (4 FAILS). 
- the 'invalid' script tags are also present in the older version on GF.


m4rc1e:
Since this font is from a reputable 3rd party, I don't mind merging it. Would also be interested to see if the fvar "Retina" instance and stat particle is going to cause any problems.

When I view the text diffs, it looks rock solid.

All missing glyphs seem to come from OT features and not encoded glyphs so I'm ok with it. When I write a better deshaper, we should be able to tell exactly what is missing, for now it isn't too accurate.

PR #3791, opened 2021-09-02
Updating Roboto's STAT table to add the SemiBold instance

aaronbell:
This update has been implemented by running `gftools gen-stat` on the current version in GF.

I've also pushed a PR to the upstream repro that implements this fix in the build script here:
https://github.com/TypeNetwork/Roboto/pull/92

As the font is only available through the releases process, I thought it easier to PR manually.


RosaWagner:
I'll ask Dave and Marc about it, but I guess we should be in sync with https://github.com/googlefonts/roboto and fix all these fails.

m4rc1e:
You cannot fix the fails since the v2.138 also has these fails. We want it to match the older version as much as possible.

aaronbell:
Yeah, in reading through the discussions on the Roboto repro, I thought it best to leave things AS-IS as much as possible for fear of breaking something. 🙈

RosaWagner:
@m4rc1e even winAscent and winDecent?

davelab6:
Roboto requires special handling due to its status as the Android UI font.
Marc please take this forwards

RosaWagner:
We wait for this one to be push to production before pushing this PR 
https://github.com/google/fonts/pull/2876

thlinard:
Since this is a small fix, would it be possible to unblock it by not linking it to #2876 (which is blocked, for other reasons it seems)?

RosaWagner:
Closing this PR since it is in conflict and Marc working on Roboto.

PR #3821, opened 2021-09-09
Dosis: Version 3.002 added

aaronbell:
43f2bf5: [gftools-packager] Dosis: Version 3.002 added

* Dosis Version 3.002 taken from the upstream repo https://github.com/googlefonts/dosis-vf at commit https://github.com/googlefonts/dosis-vf/commit/3407d52f1d1b1c36c14e756ad0b36561d8d44a3b.

 7f1be6c: [gftools-packager] ofl/dosis remove METADATA "source".  google/fonts#2587

See https://github.com/google/fonts/pull/3821 for previous info on this PR


RosaWagner:
LGTM.

Although, like we noticed in the other thread: 1. design of vietnamese diacritics is better but still not quite good, and 2. accents positioning at small sizes is worst because the font is not hinted anymore. So I think it's okay to merge cause in 1. update is better than current version, and in 2. we can't do anything about that for now.

PR #3822, opened 2021-09-09
Yuji Hentaigana Akira+Akebono : japanese subset added to metadata.pb

aaronbell:
As the packager doesn't automatically add the Japanese subset to the Yuji fonts (due to their limited Japanese character set), I've manually re-added the subset.

Additionally, while checking the other Kana-only fonts, I found that cherry bomb's is mis-ordered (should be alphabetical) and corrected that.

### DIFF
--- a/ofl/cherrybomb/METADATA.pb
+++ b/ofl/cherrybomb/METADATA.pb
@@ -12,8 +12,8 @@ fonts {
   full_name: "Cherry Bomb Regular"
   copyright: "Copyright 2019 The Cherry Bomb Project Authors (https://github.com/satsuyako/CherryBomb)"
 }
+subsets: "japanese"
 subsets: "latin"
 subsets: "latin-ext"
 subsets: "menu"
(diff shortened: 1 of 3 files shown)


RosaWagner:
Cherry Bomb has been renamed Cherry Bomb One, we gonna remove this dir for the repo.

aaronbell:
Ah! Right. To make things cleaner, I've set that one back to where it was. Now the change is just to the Yuji fonts.

RosaWagner:
I am a bit concerned by all those transformed components, it can cause rendering issues. Fontbakery was not checking it before cause fontmake is supposed to take care of it but… difficult to ignore it now that it is written. @aaronbell Can you rebuild them with the "decomposed transformed components" filter thing ?

aaronbell:
@RosaWagner I've pushed a new PR (https://github.com/google/fonts/pull/3878) that corrects the transformed components issue. This PR will remain for the metadata fix :)

PR #3856, opened 2021-09-17
sourcesans3: added

m4rc1e:
Taken from the upstream repo https://github.com/adobe-fonts/source-sans at release https://github.com/adobe-fonts/source-sans/releases/tag/3.046R


m4rc1e:
We plan to delist the existing Source Sans Pro from the Google Fonts catalogue and replace it with this family instead. Users of the existing fonts won't be affected since the endpoints for the v2 fonts will not change.

PR #3895, opened 2021-10-01
Kosugi: Version 4.002 added

aaronbell:
f8be010: [gftools-packager] Kosugi: Version 4.002 added

* Kosugi Version 4.002 taken from the upstream repo https://github.com/googlefonts/kosugi at commit https://github.com/googlefonts/kosugi/commit/75171a2738135ab888549e76a9037e826094f0ce.

 14d6366: [gftools-packager] apache/kosugi remove METADATA "source".  google/fonts#2587


aaronbell:
We did not previously have a source for Kosugi and Kosugi Maru. This PR is the version now being built from the new Glyphs source using UFR.

aaronbell:
[kosugi.zip](https://github.com/google/fonts/files/7268534/kosugi.zip)
For Rosalie :). 

One thing to note is that with the conversion to a Glyphs source, some of the glyphs were renamed, which is what I believe the diffenator is flagging.

PR #3916, opened 2021-10-07
Corinthia: Version 1.013; ttfautohint (v1.8.3) added

vv-monsalve:
36c139a: [gftools-packager] Corinthia: Version 1.013; ttfautohint (v1.8.3) added

* Corinthia Version 1.013; ttfautohint (v1.8.3) taken from the upstream repo https://github.com/googlefonts/corinthia at commit https://github.com/googlefonts/corinthia/commit/9c81839655dfec773b737abc90cb09ac94b1bea1.

 8b5a144: [gftools-packager] ofl/corinthia remove METADATA "source".  google/fonts#2587


RosaWagner:
Checking in sandbox. 
- The description talks about "three weights" but there are only two.
- Is it okay that the Bold is so much bigger?

Capture d’écran 2021-11-03 à 13 51 47

vv-monsalve:
> * The description talks about "three weights" but there are only two.

Mhm, they were three at the beginning. We discarded the Medium in favour of the Bold. I'll change the description.

> * Is it okay that the Bold is so much bigger?

Since these are script fonts that are not meant to be used in paragraphs, I would say yes.

Eventually, we could use the discarded Medium weight (which is a bit lighter) and make it the Bold. However, this would involve reworking this project. I'd like to hear @ davelab6's opinion.

Screen Shot 2021-11-03 at 10 27 52

davelab6:
It does seem very strange that the bold increases in scale; as a VF wght axis this would also be odd. I think its OK to hold this up until it can be reworked

vv-monsalve:
Metrics increase 6 units in the x-Height (which could be considered expected when creating a bold face for visual compensations). However, what changes is the horizontal scale, even for the Medium weight.

At that moment, [when we review the differences](https://github.com/googlefonts/corinthia/issues/5#issuecomment-907175986) in the available weights, the decision went for the Bold, to make use of the (b) button in apps like word, discarding the possibilities for new weights to be commissioned.

Depending on the desired, the options would be

- [ ] Turning the Medium weight as the Bold as it is (that is with the horizontal scaling, smaller though)
- [ ] Asking the designer to rework entirely the font to not have a horizontal change in scale while using the current bold weight

Screen Shot 2021-11-04 at 13 53 04

Screen Shot 2021-11-04 at 13 56 33

davelab6:
@vv-monsalve and I discussed this and I'd like to just proceed with the family as-is

PR #3419, opened 2021-05-13
Updated Noto Sans CJK to 2.004 using variable fonts

marekjez86:
Updated Noto Sans CJK to 2.004 using variable fonts. Please review ASAP

PR #3928, opened 2021-10-08
Update Noto Sans HK, JP, KR, SC, TC to 2.004 using variable fonts

moyogo:
This replaces #3419 to update Noto Sans HK, JP, KR, SC, TC to 2.004 using variable fonts.
- Use variables fonts from Noto Sans CJK (https://github.com/googlefonts/noto-cjk/commit/cee7438f5f8e66397090d483c15275d1af3d87c7)[cee7438]
- Update fonts metadata for variable fonts
- Update names in fonts from "Noto Sans CJK HK" to "Noto Sans HK", etc. to match googlefonts naming scheme.

### DIFF
--- a/ofl/notosanshk/METADATA.pb
+++ b/ofl/notosanshk/METADATA.pb
@@ -7,59 +7,18 @@ fonts {
   name: "Noto Sans HK"
   style: "normal"
   weight: 100
-  filename: "NotoSansHK-Thin.otf"
+  filename: "NotoSansHK[wght].ttf"
   post_script_name: "NotoSansHK-Thin"
   full_name: "Noto Sans HK Thin"
-  copyright: "Copyright  2014, 2015, 2018 Adobe Systems Incorporated (http://www.adobe.com/)."
-}
-fonts {
-  name: "Noto Sans HK"
-  style: "normal"
-  weight: 300
-  filename: "NotoSansHK-Light.otf"
-  post_script_name: "NotoSansHK-Light"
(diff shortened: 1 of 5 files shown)


thlinard:
Thanks for the fvar.NamedInstance and STAT.AxisValue matching!

But according to [Google Fonts Docs (WIP)](https://github.com/googlefonts/gf-docs/tree/main/Spec#stat-table):

> And, according to the systematization principle, the system goes along the axis, and derives a static font at each of the nine 100 values, if the axis covers that value (whether or not they are included in the designer's original design space), and names it accordingly. 

Presently, the instances in the VFs are:

- Thin
- Light
- DemiLight
- Regular
- Medium
- Bold
- Black

When the VFs will be available in the API, the static fonts will be:

- Thin
- ExtraLight
- Light
- Regular
- Medium
- SemiBold
- Bold
- ExtraBold
- Black

Wouldn't it be better to anticipate this situation by immediately matching the instances of the VFs and the future static fonts?

marekjez86:
@moyogo : what should we do with this?

RosaWagner:
I would suggest to onbaord each family in separate PRs, using packager.

These are the issues to fix:
-To keep compatibility with already onbaorded statics, we need these style names:
Thin, ExtraLight, Light, Regular, Medium, SemiBold, Bold, ExtraBold, Black.
- STAT and FVAR tables are complete with location of the weight axis every 100.

It would be good to figure this out @moyogo @marekjez86 @simoncozens

PR #3969, opened 2021-10-26
montserrat: v8.000 added

### LINKED ISSUE #2976: Update Montserrat to variable font
Montserrat, Montserrat Alternates, and both of their italic versions all have variable-weight fonts that were made in collaboration with the original type designer. Is it possible for Google Fonts to host these? 😇

https://github.com/TypeNetwork/Montserrat
https://github.com/TypeNetwork/Montserrat/tree/master/fonts/variable

m4rc1e:
Taken from the upstream repo https://github.com/googlefonts/Montserrat at tag https://github.com/googlefonts/Montserrat/releases/tag/v8.000

Closes #2976


m4rc1e:
I'm not sure whether the hinting is up to scratch:

Thin-Regular_Desktop_Windows_10_edge_18 0_

- Counter of "a" in Thin and ExtraLght are too cramped.
- Lowercase accents feel a touch too tight.


I've chatted to Mike D about this and he doesn't believe we could improve the result.

davelab6:
That's probably enough reason to push this as "Montserrat Flex" or some other family name. The accents falling, the counters - and bars like on "A" - falling, and some strokes (like "H" bar) becoming so blurry, are all going to be issues for existing users...

m4rc1e:
Here's a sample at 17pt:




Desktop_Windows_10_edge_18 0_

- Count of "a" no longer cramped in Thin and ExtraLight
- x height has decreased

PR #4009, opened 2021-11-03
Christian Naths : designer added

emmamarichal:
Designer of Redacted font.


catalog/designers/christiannaths/info.pb
```diff
@@ -0,0 +1,5 @@
+designer: "Christian Naths"
+link: ""
+avatar {
+  file_name: "Christiannaths.png"
```
RosaWagner, line 4: we don't want uppercases in files name, @emmamarichal can you update the actual png and this link?

PR #4053, opened 2021-11-10
Update METADATA.pb for Nokora to fix license value

davelab6:
@yanone @RosaWagner note that "APACHE" and "OFL" are the allowed values for the license key, I'm curious how this slipped through, but I'd like to move this to OFL in any case.


nathan-williams:
Please do not add to `to_sandbox.txt` or `to_production.txt`. Nokora already updated and moved under OFL downstream.

PR #4068, opened 2021-11-16
Latest Roboto-Flex file

rsheeter:


thlinard:
No instances (`fvar`, `STAT`) at all?

rsheeter:
Don't need those for the web api strictly speaking. Best viewed as being in an experimental state.

thlinard:
OK, thank you for the details!

PR #4077, opened 2021-11-18
Merriweather: Version 2.100 added

### LINKED ISSUE #2943: Update Merriweather font to new version
The font [Merriweather](https://fonts.google.com/specimen/Merriweather) 2 years ago chaged they repo to one, under they foundry at [Github/SorkinType](https://github.com/SorkinType/Merriweather). I would like to request an update to the latest version, or if I could do it, I more than happy to do it.

**Production notes:**
Don't forget to check if all open issues for Merriweather are fixed with this upgrade, and close them at the same tile.

m4rc1e:
3071616: [gftools-packager] Merriweather: Version 2.100 added

* Merriweather Version 2.100 taken from the upstream repo https://github.com/SorkinType/Merriweather at commit https://github.com/SorkinType/Merriweather/commit/87887d8a93e9bb1e22b6e30f2852591111adfd5e.

Fixes #2943


m4rc1e:
Spacing is now tighter but it does look much better.

Everything else LGTM

m4rc1e:
Me and Dave have decided we need to match the original opsz size better. Will submit an update after speaking to Eben.

davelab6:
Like in https://github.com/google/fonts/pull/4077 I think we might want to use a new family name for the opez'd Merriweather. @ebensorkin

RosaWagner:
Closing this PR since it will be re-submit under another family name

m4rc1e:
> Me and Dave have decided we need to match the original opsz size better. Will submit an update after speaking to Eben.

Apologies for the delay.

Today, I've finally tested changing the opsz values so we get similar line lengths. Unfortunately increasing the opsz value doesn't work because the forms get more condensed so the line length differences actually increase. I cannot decrease the opsz value because the current default is 6pt, while the min value is also 6pt.

Looking at the diffs, It seems that the VFs now have kerning, while the old static fonts don't. I believe this is what causing the differences. Since the line lengths are tighter, reflow issues shouldn't be as bad.

We have the following options:

- Release the VFs as is and deal with the consequences. Adding kerning imo is a fix.
- Remove the kerning from the VFs (less than ideal imo since the kerns have improved the font)
- Release the family under a new name such as Merriweather Pro


cc @davelab6 @chrissimpkins

davelab6:
Contra my previous comments, I would like to push this over the top of the old Merriweather and upgrade all the websites using it :)

m4rc1e:
Updated

## Merriweather: Version 2.200;gftools[0.9.31] added

---
 43e5d7c: [gftools-packager] Merriweather: Version 2.200;gftools[0.9.31] added

* Merriweather Version 2.200;gftools[0.9.31] taken from the upstream repo https://github.com/sorkintype/merriweather at commit https://github.com/sorkintype/merriweather/commit/761487921dc06bfc40fe0560afb93d5535f41850.

m4rc1e:
@ebensorkin How's Merriweather going? (sorry for pinging you again ;-)).

EbenSorkin:
I got sick midweek and think as of this evening I’m starting to recover.
Before I got sick I was closing in on completing the Roman. Italic should
be much much faster

m4rc1e:
Updated

PR #4093, opened 2021-11-19
Zen Old Mincho: Version 1.500 added

aaronbell:
bc9fdf1: [gftools-packager] Zen Old Mincho: Version 1.500 added

* Zen Old Mincho Version 1.500 taken from the upstream repo https://github.com/googlefonts/zen-oldmincho.git at commit https://github.com/googlefonts/zen-oldmincho/commit/fb46620f2ffcf78debc1a20758e39271a3fb5ea6.

 112a2f5: [gftools-packager] ofl/zenoldmincho remove METADATA "source".  google/fonts#2587

### DIFF
--- a/ofl/zenoldmincho/METADATA.pb
+++ b/ofl/zenoldmincho/METADATA.pb
@@ -48,8 +48,6 @@ fonts {
   full_name: "Zen Old Mincho Black"
   copyright: "Copyright 2021 The Zen Old Mincho Project Authors (https://github.com/googlefonts/zen-oldmincho)"
 }
-subsets: "chinese-simplified"
-subsets: "chinese-traditional"
 subsets: "cyrillic"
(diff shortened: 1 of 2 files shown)


aaronbell:
This update adds the two missing weights of the Old Mincho family. 

Also to note—the former Bold weight has been positioned back as SemiBold. The actual difference between these two weights is minimal and I do not expect any issues as a result of this change.

RosaWagner:
FYI this family is blocked because the additional styles makes the family size > 25mb, at some point we will need to push this family.

RosaWagner:
@chrissimpkins just realising that adding the primary_script in #5410 already pushed the entire family to sandbox. So I'll have to remember to remove this family from the prod list.

chrissimpkins:
Still need something here or are we good?

PR #3893, opened 2021-09-30
Update Oswald[wght].ttf

aaronbell:
This update improves the rendering of math symbols in Oswald. 
See: https://github.com/googlefonts/OswaldFont/pull/24

(BTW, I couldn't run `gftools packager` on this because of the TARGET_OS_XXX bits in the metadata file 😕 so had to submit this the old fashioned way).

PR #4140, opened 2021-12-02
oswald/DESC updated to correct Cyrillic info

davelab6:
Resolves https://github.com/googlefonts/OswaldFont/issues/25

### DIFF
--- a/ofl/oswald/DESCRIPTION.en_us.html
+++ b/ofl/oswald/DESCRIPTION.en_us.html
@@ -5,8 +5,8 @@ Oswald is designed to be used freely across the internet by web browsers on desk
 </p>
 <p>
 Since the initial launch in 2011, Oswald was updated continually by Vernon Adams until 2014.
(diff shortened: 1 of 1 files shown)


RosaWagner:
Does it mean that we can also merge https://github.com/google/fonts/pull/3893 ?

alexeiva:
@davelab6 does this also remove the designer icon?
Screenshot 2021-12-04 at 09 31 37

davelab6:
@RosaWagner Description updates run along a different path to TTF updates, so we can merge and push this separately to that PR which updates the VF. 

@alexeiva no, good point. I'll fix that now

davelab6:
OK since that is controlled in the METADATA file, I've updated it in the other PR. I'll check with the eng team if that's the correct way to go.

RosaWagner:
I would rather wait for confirmation from the eng team so we can merge all the Oswald pending PRs. Indeed there is a version of the font blocked in sandbox and putting this font directory on the prod list I'm afraid would also give instruction to push the blocked font.

davelab6:
I'll ask @rsheeter about this too

PR #4141, opened 2021-12-02
inconsolata/METADATA.pb Add 2nd sans category

davelab6:
This is how I expect we'll add more than 1 category to metadata; the 2nd+ category keys MUST be commented out for now, but we can start adding the data we have to new PRs this way, and then uncomment them when downstream is ready


davelab6:
Yep, replaced by https://github.com/googlefonts/gftools/pull/451

PR #4152, opened 2021-12-07
Bumped to 8.002 (urgent glyph definition issues)

yanone:
Updated 14 of 18 of Danh Hong’s Khmer fonts; solved an urgent glyph definition issue where several glyphs that are getting substituted in GSUB aren't getting recognized as letters by Microsoft Edge, so categorized them as letters explicitly in the Glyphs sources.


yanone:
Note: The version numbers of the fonts are not consistently 8.002. I double checked that they are okay tho.
One is 8.001 because it got skipped in the previous round of fixes, and one is at 2.002 because it started out wit 2 and not 8. The version number are good to go.

PR #4174, opened 2021-12-15
Math clarification done

vv-monsalve:
The description includes clarification about Mah font availability.


vv-monsalve:
@RosaWagner since this PR is only updating the Description and OFL files, the fails reported for the font files could be ignored for now.

PR #4194, opened 2022-01-03
Zen Kaku Gothic Antique: Version 1.002 added; Zen Kaku Gothic New: Version 1.002 added

aaronbell:
314bbc9: [gftools-packager] Zen Kaku Gothic Antique: Version 1.002 added

* Zen Kaku Gothic Antique Version 1.002 taken from the upstream repo https://github.com/googlefonts/zen-kakugothic.git at commit https://github.com/googlefonts/zen-kakugothic/commit/81443c39f8364063b747f2f1af778969d60acee8.

 c5ea0e7: [gftools-packager] ofl/zenkakugothicantique remove METADATA "source".  google/fonts#2587

 f9aaf3d: [gftools-packager] Zen Kaku Gothic New: Version 1.002 added

* Zen Kaku Gothic New Version 1.002 taken from the upstream repo https://github.com/googlefonts/zen-kakugothic.git at commit https://github.com/googlefonts/zen-kakugothic/commit/81443c39f8364063b747f2f1af778969d60acee8.

 89922d7: [gftools-packager] ofl/zenkakugothicnew remove METADATA "source".  google/fonts#2587


aaronbell:
Alright, I think I got all of this ready to go. It seems there's some language options that have been added that the packager didn't add and I'm not totally sure about them, so thought I'd call it out.

aaronbell:
The issue here is due to how GlyphsLib processes Glyphs files. If no glyphs are assigned a vertical height value (even if the vertical height value needs to be the default of 1000), GlyphsLib will not automatically assign a vertical value in the vmtx table. In this case, the "shared" glyphs file had an added 1000 height assigned, but the kana files did not. As a result, the kana had a set height of 0 rather than 1000. 

This has been corrected.

m4rc1e:
Language info seems to have been ripped out of the METDATA.pb files. I'll need to make some changes to gftools so this doesn't happen. Let's block this for now.

aaronbell:
I’d rather not block a fix to broken functionality due to a metadata issue. Would it be OK if I manually updated the languages in this case rather than wait for a tooling fix?

m4rc1e:
Sounds fine to me on this occasion but please don't PR any family updates until the tooling is fixed. Hopefully, I'll have it done this week.

aaronbell:
Updated the metadata. Thanks!

aaronbell:
@m4rc1e has this been pushed live as of yet? Per the label it looks like it is still in sandbox.

RosaWagner:
Apparently you guys forgot to add the PR to Traffic Jam project, don’t expect much happening if it’s not in positioned to be tracked :)

aaronbell:
Thanks @RosaWagner. That explains it :)

PR #4149, opened 2021-12-04
Add Inter Italic font

valerie-makes:
Downloaded from https://github.com/rsms/inter/releases/download/v3.19/Inter-3.19.zip

File path inside zip: `Inter-3.19/Inter Variable/Single axis/Inter-italic.ttf`

Fixes #2386

PR #4211, opened 2022-01-12
inter: v3.19 added

### LINKED ISSUE #2386: Inter fonts lacks italics

m4rc1e:
Taken from https://github.com/rsms/inter/releases/download/v3.19/Inter-3.19.zip

@valerie-makes thanks for submitting #4149. I've decided to make a new PR since I have the correct permissions to open up a pr directly from this repo. This allows our CI to run correctly.

Fixes #2386


valerie-makes:
@m4rc1e it may be helpful adding "Fixes #2386" to the PR description :)

RosaWagner:
It looks like nameID6 is taking name ID 25+2 instead of name ID 1+2. The latter would avoid post-script name regression.

davelab6:
@RosaWagner Yes, Adobe Fonts has asked us to be VERY careful about changes to PS names. If we can avoid it, we should. 
 So, @felipesanches what is the status of the fontbakery check for this? That is urgent to be into everyone's process. 
 
 @m4rc1e Since the `slnt` axis for this family was already added (like for [Recursive](https://github.com/google/fonts/tree/main/ofl/recursive) then I think as an exception we do not want to add our own forked Italic file, as in this PR, because that will complicate untangling the slnt axis. We'll need to discuss the case of these 2 fonts with the GF Eng team, since making some "italic" styles available to Workspace users is important, but there are deeper problems with how slnt, font-style, etc work. 
 
@RosaWagner please could you explain your ideas about how to to proceed here?

RosaWagner:
I tested it and it works fine in app that doesn't have a good support of the slant axis, but it bugs in apps that does have a "good" support slant axis. So this is a bad idea. I don't see much solution here to avoid regression on the user end, except for onbaording it under another name.

PR #4215, opened 2022-01-13
Piazzolla: Version 2.005 added

RosaWagner:
5994ee8: [gftools-packager] Piazzolla: Version 2.005 added

* Piazzolla Version 2.005 taken from the upstream repo https://github.com/huertatipografica/piazzolla at commit https://github.com/huertatipografica/piazzolla/commit/2c84615f45b98b75238401c53ef993e558082a63.

 029104c: [gftools-packager] ofl/piazzolla remove METADATA "source".  google/fonts#2587


RosaWagner:
@m4rc1e looks like the lang-support worked ;) although the muted comment about the language is gone, do we care?

m4rc1e:
hmm. It shouldn't have removed the comments. Let me take a look.

m4rc1e:
Alright, I see what's happening here. When a metadata file is parsed, it strips out the comments. Unfortunately, I don't think there is a way to retain the comments. cc @nathan-williams 

Perhaps we should be regenerating the languages each time add-font is run?

RosaWagner:
@juandelperal the font has changed FVAR and STAT, we need it to be the same as before. You can eventually add some STAT particules such as 18pt and 24pt since they are the ones generated and served by GF.

This how it was before:

Capture d’écran 2022-01-13 à 13 50 32
Capture d’écran 2022-01-13 à 13 50 22

RosaWagner:
Ah wait, I misread the fontbakery report, I think it is all good

PR #4232, opened 2022-01-21
Noto Sans+Serif Display: correcting category to DISPLAY

marekjez86:


davelab6:
@marekjez86 @RosaWagner I am not sure this is correct; I think the "display" category is for typefaces that are 'fun' and don't fit into the other categories - eg https://github.com/google/fonts/blob/main/ofl/playfairdisplay/METADATA.pb#L4 is serif, not display - and not for sans/serif designs that have a formal feeling. 

Now that we are building up multiple categories, I would suggest that we might pre-pend display to the line above sans/serif on these 2, so that they remain as-is for the current single category system, but will expand in a future multi category system.

marekjez86:
@davelab6  How would you differentiate between Noto Sans and Noto Sans Display? what different categories would you assign them?

tiroj:
‘Display’ has two common meanings in typography that, in reference to specific typefaces, sometimes overlap and sometimes do not.

1. Display refers to a stylstic category to encompass a wide variety of typefaces that are considered unsuitable for running text because they e.g. fancy, decorated, multi-coloured, or otherwise atypical of text type styles.

2. Display refers to a size range, the definition of which is imprecise and may vary by font maker, but which is generally above 36pt but smaller than Banner or Poster; hence also used as an optical-size design targeting such sizes.

My understanding from Dave’s comment is that Display on the Google Fonts page is used in the first sense, while Display in the Noto Serif and Noto Sans families is used in the latter sense.

marekjez86:
@tiroj : thank you John

PR #4236, opened 2022-01-23
Update category for Source Serif 4 to Serif

mgoszcz2:


davelab6 (approved):

davelab6:
Thanks @mgoszcz2 - good catch. Our automation default to sans, and sometimes we don't spot it needs to be updated :)

PR #4240, opened 2022-01-25
Rubik Beastly: Update METADATA.pb

nathan-williams:


RosaWagner:
@nathan-williams Can you confirm that this mean we shouldn't credit NaN as a foundry to all their Rubik Filtered upcoming fonts?

nathan-williams:
> @nathan-williams Can you confirm that this mean we shouldn't credit NaN as a foundry to all their Rubik Filtered upcoming fonts?

We should definitely credit [NaN](https://www.nan.xyz/) as a foundry. Sorry for the confusion on my end! I'll revert the PR.

PR #4275, opened 2022-02-09
Anek Bangla: Version 1.003 added

yanone:
428cfea: [gftools-packager] Anek Bangla: Version 1.003 added

* Anek Bangla Version 1.003 taken from the upstream repo https://github.com/EkType/Anek at commit https://github.com/EkType/Anek/commit/634384abf35544275862ae22c7da44cc9c916caa.

 2cb2ad1: [gftools-packager] ofl/anekbangla remove METADATA "source".  google/fonts#2587


yanone:
I just PRd all the Anek fonts. It now shows two more FAILs, both related to the description file. Both are fine. The short link to /?query=Anek (to save space) is valid, just not reachable like that from the outer internet. And the total size was below 2000 when he sent me the file back. My VSCode introduced the line breaks and now it's a few characters more.

yanone:
One more thing: The last four families, Malayalam, Odia, Tamil, and Telugu, show the subset below latin in METADATA.pb. I thought they were meant to be on top. But it feels like it's an alphabetic ordering, because the first families up to latin show it before.

RosaWagner:
> One more thing: The last four families, Malayalam, Odia, Tamil, and Telugu, show the subset below latin in METADATA.pb. I thought they were meant to be on top. But it feels like it's an alphabetic ordering, because the first families up to latin show it before.

Yes in must be in alphabetical order, so all good.

PR #4274, opened 2022-02-07
Update Noto Sans

simoncozens:
This updates the existing Noto Sans fonts to the latest versions from https://github.com/googlefonts/noto-source.

Important notes!

* This is my first google/fonts PR. I don't really know what I'm doing!
* The existing Noto Sans on GF was built by hand with a merged Noto Sans Devanagari. So this build also includes a merged Devanagari. For reproducibility:  I built Noto Sans using notobuilder (https://github.com/googlefonts/noto-source/pull/499) and then I used the merge script below to add Devanagari (binaries from https://github.com/googlefonts/noto-fonts/tree/main/hinted/ttf/NotoSansDevanagari) and fix up some fontbakery issues.
* Existing Noto Sans on GF just has RIBBI styles. While I was merging everything anyway, I merged all the weight styles we have because why not, so this expands the weight range.
* Existing Noto Sans Regular on GF has PostScript name "NotoSans" *not* "NotoSans-Regular". **I have not conformed to this**. `NotoSans-Regular.ttf` in this PR is just the way it came out of notobuilder, and it has PS Name "NotoSans-Regular", which fontbakery is much happier about. So this is a change to what is in the catalogue. I don't know if this is a problem. 
 
Fontbakery fails:

* Mostly about copyright strings. The copyright string for Noto Sans is "Copyright 2015-2021 Google LLC. All Rights Reserved." which makes Fontbakery sad . But I think the Noto style copyright should be kept; it seems more correct than the usual "The ... Project Authors". We may need to make a FB exception for this?
* "Designers are listed correctly on the Google Fonts catalog". I don't know how this is supposed to work for Noto. All the other Noto fonts seem to have "Designer: Google" so maybe they fail too.
* "Check samples can be rendered." This is because the italic fonts have no Devanagari; we have no Noto Sans Devanagari Italic, and this is consistent with the existing GF Noto Sans. The upright styles can renderer all the sample glyphs.

## Merge script (for posterity)

```python
from fontTools.ttLib import TTFont
from fontTools.merge import Merger
from tempfile import NamedTemporaryFile
from gftools.fix import fix_family
import glob
import re

NAME_13 = "This Font Software is licensed under the SIL Open Font License, Version 1.1. This license is available with a FAQ at: https://scripts.sil.org/OFL"

def fix_up_glyph_names(font):
	# Requires compiling a bunch of the tables, easiest to save/reload
	with NamedTemporaryFile() as ntf:
		font.save(ntf.name)
		font = TTFont(ntf.name)

[...]

PR #4291, opened 2022-02-10
Noto Sans: 2.007 added

simoncozens:
Taken from the upstream repo [googlefonts/noto-fonts](https://github.com/googlefonts/noto-fonts/) at commit [a8800e6](https://github.com/googlefonts/noto-fonts/commit/a8800e6798c6837b981158d5a1f944ad480e06d5).

See notes from #4274.


RosaWagner:
@simoncozens I don't remember, is this PR ready for review or still wanted to correct the fonts after generating the fontbakery report? If second option, you can use the label "in progress" to notify us so we don't merge.

simoncozens:
I think I'm waiting for you!

RosaWagner:
I only see this that is concerning since the font has devanagari:
🔥 FAIL: Ensure dotted circle glyph is present and can attach marks.

simoncozens:
Argh. This is a new check, and I’m working on binaries for this font merge.

RosaWagner:
> Argh. This is a new check,

Yes, then I think we can ignore if it can make the process faster and if it will be in the next update.

Since it's a very important font, I'd like someone to confirm these diff are okay (there can be bugs in the diff though, but better be sure):

[diffbrowsers.zip](https://github.com/google/fonts/files/8123448/diffbrowsers.zip)

simoncozens:
Looks OK to me but the ring of Å is detached in the Italic. (But that's what's in Noto Sans, so...)

Anyway, I will leave Marek to look over the diffs and give the executive opinion.

marekjez86:
looking at it

marekjez86:
Looks great. Thank you Simon!

PR #4303, opened 2022-02-16
Add Roboto Serif

davelab6:
@m4rc1e cut the release of this in the now public source repo @ github.com/googlefonts/roboto-serif and now the family is live in fonts.google.com I'm adding files for this repo that I hand crafted from the internal versions used to push the family live.

Please take a look and see if you can find anything unusual, and then merge :)


RosaWagner:
Is there a public repo for Roboto Serif ?
Also postscript name is unusual.

davelab6:
@m4rc1e mastered this release  back in August so I guess it needs remastering to the very latest GF standards.... since Marc is currently focused on Roboto Flex, please could you look at this? Repo is @ https://github.com/googlefonts/roboto-serif and I see that is missing from DESC now...

RosaWagner:
After discussion with Marc we keep the font just like this, and merge this PR in order to correct the categorisation on GF that is SANS_SERIF instead of SERIF to fix #4315

PR #4333, opened 2022-02-25
Sofia Sans: Version 4.100 added

vv-monsalve:
f1da648: [gftools-packager] Sofia Sans: Version 4.100 added

* Sofia Sans Version 4.100 taken from the upstream repo https://github.com/vv-monsalve/Sofia-Sans at commit https://github.com/vv-monsalve/Sofia-Sans/commit/b98843f87feff8144257c8e250867f098466a6e1.


vv-monsalve:
This is a draft pull request to verify everything is working well before reaching the designer with the latest work done.

RosaWagner:
- We should probably merge it to see how it behaves in the dev platform.
- The names on the width axis shouldn't have spaces, like the weight instance names (I know it is not the case in the axis registry, but this space is actually removed after)
- afaik the non-normal-width statics generated in the zip file won't be the exact same ones as the static generated from source in the upstream repo, but on GF side I think it's okay.

vv-monsalve:
Thanks @RosaWagner for the revision

> * We should probably merge it to see how it behaves in the dev sandbox.
> * The names on the width axis shouldn't have spaces, like the weight instance names (I know it is not the case in the axis registry, but this space is actually removed after)
Ok, good to know about the no spaces for `wdth` axis. I'll fix them and update the PR so that we can merge and test.

> * afaik the non-normal-width statics generated in the zip file won't be the exact same ones as the static generated from source in the upstream repo, but on GF side I think it's okay.
Do you know details on what would change? I'll probably need to inform the designer once we confirm everything else is ready to go.

vv-monsalve:
Updated

## Sofia Sans: Version 4.101 added

---
 fab5c58: [gftools-packager] Sofia Sans: Version 4.101 added

* Sofia Sans Version 4.101 taken from the upstream repo https://github.com/vv-monsalve/Sofia-Sans at commit https://github.com/vv-monsalve/Sofia-Sans/commit/a849e481fe767e946b74c59d4628d7cf896ed023.

vv-monsalve:
@RosaWagner I've changed the `wdth` names and updated the PR

RosaWagner:
Let's merge to see how it behaves in dev-sandbox, but we won't push to prod yet, so I mark it as blocked.

I don't know how to explain the problem with he static instances in a synthetic way in English, better to call each other and I show you ;)

PR #4346, opened 2022-03-03
robotoserif: v1.007 added

m4rc1e:
Taken from the upstream repo https://github.com/googlefonts/roboto-serif at release https://github.com/googlefonts/roboto-serif/releases/tag/v1.007


m4rc1e:
- family doesn't need an avar table (first release also didn't have one)
- GRAD issue is peculiar. The sources have the same width so this may be a build chain issue. Previous fonts also have this issue though so I'll skip it
- vertical metrics have been fixed.
- FB names are a false positive.


*Demo is new vertical metrics. We can also see there are no advance width regressions*

PR #4360, opened 2022-03-11
robotoflex: v3.002 added

m4rc1e:
Taken from the upstream repo https://github.com/googlefonts/roboto-flex at tagged release https://github.com/googlefonts/roboto-flex/releases/tag/3.002


m4rc1e:
> FAIL: Ensure VFs do not contain slnt or ital axes.

Dave gave this the go ahead

> FAIL: Ensure VFs with the GRAD axis do not vary horizontal advance.

Will be fixed in the next release

> FAIL: Validate defaults on fvar table match registered fallback names in GFAxisRegistry.

Axis overrides have been provided in the METADATA.pb file. I'm currently unsure if we even need this check, https://github.com/googlefonts/fontbakery/issues/3646

PR #4368, opened 2022-03-11
Added Tokotype

yanone:


RosaWagner:
Can you re-do the PR with the bio written at the third person?
FYI we can't check it in sandbox as long as a font linked to it is onboarded too

PR #4387, opened 2022-03-22
Tall Chai catalogue info

yanone:


RosaWagner:
@yanone the image is jpg when it should be png. 
You can use the script `gftools add-designer` so everything gets formatted etc and prevent that from happening :)

tallchai:
@yanone Any update on this? Thanks. :)

PR #4401, opened 2022-03-23
BIZ UDMincho: Version 1.001 added; BIZ UDPMincho: Version 1.001 added

aaronbell:
f88a267: [gftools-packager] BIZ UDMincho: Version 1.001 added

* BIZ UDMincho Version 1.001 taken from the upstream repo https://github.com/googlefonts/morisawa-biz-ud-mincho at commit https://github.com/googlefonts/morisawa-biz-ud-mincho/commit/89dea71623040ec5dc2e3c54e7a86ed5171b3035.

 3efa7d3: [gftools-packager] BIZ UDPMincho: Version 1.001 added

* BIZ UDPMincho Version 1.001 taken from the upstream repo https://github.com/googlefonts/morisawa-biz-ud-mincho at commit https://github.com/googlefonts/morisawa-biz-ud-mincho/commit/89dea71623040ec5dc2e3c54e7a86ed5171b3035.


aaronbell:
We discovered an issue that crops up with Windows GDI rendering in which the non "P" fonts are being classified as monospace and thus render the Japanese characters as extra-wide. I've reverted the change to the PANOSE value that I'd made previously and followed exactly what was done in the original Morisawa files. The issue no longer appears.

PR #4402, opened 2022-03-23
BIZ UDGothic: Version 1.001 added; BIZ UDPGothic: Version 1.001 added

aaronbell:
3311987: [gftools-packager] BIZ UDGothic: Version 1.001 added

* BIZ UDGothic Version 1.001 taken from the upstream repo https://github.com/googlefonts/morisawa-biz-ud-gothic at commit https://github.com/googlefonts/morisawa-biz-ud-gothic/commit/ed0bb20ae36bb556c4f3de741f3ee95226c1dbbd.

 9edd7d9: [gftools-packager] BIZ UDPGothic: Version 1.001 added

* BIZ UDPGothic Version 1.001 taken from the upstream repo https://github.com/googlefonts/morisawa-biz-ud-gothic at commit https://github.com/googlefonts/morisawa-biz-ud-gothic/commit/ed0bb20ae36bb556c4f3de741f3ee95226c1dbbd.


aaronbell:
We discovered an issue that crops up with Windows GDI rendering in which the non "P" fonts are being classified as monospace and thus render the Japanese characters as extra-wide. I've reverted the change to the PANOSE value that I'd made previously and followed exactly what was done in the original Morisawa files. The issue no longer appears.

PR #4443, opened 2022-03-30
Use the usual name for CI, show a badge for CI on root readme

rsheeter:
Precursor to adding additional checks

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,11 +1,12 @@
 # Check to_sandbox.txt and to_production.txt do not contain typos
-name: Google Fonts Test Server Files
+name: Continuous Test + Deploy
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -1,11 +1,12 @@
 # Check to_sandbox.txt and to_production.txt do not contain typos
-name: Google Fonts Test Server Files
+name: Continuous Test + Deploy
```
davelab6, line 2: Since this CI doesn't deploy, is there a better name for this?

rsheeter: We could just remove "+ Deploy"?

rsheeter: I'll send a PR

PR #4450, opened 2022-03-30
Tiro Bangla: Version 1.52 added

vv-monsalve:
6211bd3: [gftools-packager] Tiro Bangla: Version 1.52 added

* Tiro Bangla Version 1.52 taken from the upstream repo https://github.com/TiroTypeworks/Indigo at commit https://github.com/TiroTypeworks/Indigo/commit/98b5d3a01c534b7029af04a59d31e0d7d90be15a.


vv-monsalve:
Neelakash Kshetrimayum bio info was added in #4448

RosaWagner:
Putting this in "needs confirmation" because of this warn:

⚠ WARN: Checking Vertical Metric Linegaps. ([com.google.fonts/check/linegaps](https://font-bakery.readthedocs.io/en/latest/fontbakery/profiles/hhea.html#com.google.fonts/check/linegaps))
⚠ WARN **hhea lineGap is not equal to 0.** [code: hhea]

It goes with all the tiro indic fonts. I am sure this is all thought through but since it is against GF policy, a written confirmation would ensure it's all good :)

tiroj:
Yes, thanks Rosa. Tiro practice is to align hhea metrics to OS/2 Typo metrics. In the case of the Tiro Indic fonts, one of the goals is to maintain compatibility with the versions of the fonts shipped by Apple (as well as with the preceding Murty fonts as far as possible). This also explains why the PS name fields contravene usual GF practices. All discussed with @davelab6 during development and release planning.

RosaWagner:
Thanks for the answer :) then I should be merging them all very soon.

tiroj:
[For the record: I consider aligning hhea ascender and descender metrics to OS/2 Win metrics—and hence leaving the linegap metric at zero—to be shortsighted. It links Mac platform linespacing to Windows clipping metrics, which might need to change in future if taller glyphs are added to a font and the bounding box increases. Ultimatel, the mess of incompatible linespacing behaviours is only going to get fixed if software and fonts move towards the intended purpose of the different vertical metrics and stop trying to second-guess each other.]

Issue #4405, opened 2022-03-24
Set overlap bit on every glyph in every VF and keep it set 


davelab6:
In order for TT VFs to be correctly rendered with outlines, each glyph in the glyf table needs to have the overlap bit set, or ideally each glyph that has overlaps although that's more complicated to implement.

And we need string regression checks to prevent that slipping, and that will be easier as well if it's just all VF glyphs I suppose.

aaronbell:
Don’t know if it is useful but this is the code I put in Cascadia Code’s build script to solve this problem:

```
def set_overlap_flag(varfont: fontTools.ttLib.TTFont) -> fontTools.ttLib.TTFont:
    glyf = cast(_g_l_y_f.table__g_l_y_f, varfont["glyf"])
    for glyph_name in glyf.keys():
        glyph = glyf[glyph_name]
        if glyph.isComposite():
            # Set OVERLAP_COMPOUND bit for compound glyphs
            glyph.components[0].flags |= 0x400
        elif glyph.numberOfContours > 0:
            # Set OVERLAP_SIMPLE bit for simple glyphs
            glyph.flags[0] |= 0x40

```

madig:
I remember reading on the FreeType devel list that FT will switch to more expensive supersamling mode when encountering overlap bits, which is why setting them indiscriminately may be a bad idea. The script by @aaronbell is what we use in production to quickly set the bit if needed, per glyph :) No idea if something needs to be done for CFF2.

chrissimpkins:
> I remember reading on the FreeType devel list that FT will switch to more expensive supersamling mode when encountering overlap bits, which is why setting them indiscriminately may be a bad idea. The script by @aaronbell is what we use in production to quickly set the bit if needed, per glyph :) No idea if something needs to be done for CFF2.

This thread seems to summarize the state of the conversation around this as of late 2020 https://github.com/fonttools/fonttools/issues/2059

garretrieger:
> Don’t know if it is useful but this is the code I put in Cascadia Code’s build script to solve this problem:
> 
> ```
> def set_overlap_flag(varfont: fontTools.ttLib.TTFont) -> fontTools.ttLib.TTFont:
>     glyf = cast(_g_l_y_f.table__g_l_y_f, varfont["glyf"])
>     for glyph_name in glyf.keys():
>         glyph = glyf[glyph_name]
>         if glyph.isComposite():
>             # Set OVERLAP_COMPOUND bit for compound glyphs
[...]

PR #4472, opened 2022-04-06
opensans v3.001 added

m4rc1e:
This PR enables the overlap flag bits for glyphs which have overlapping contours. It was created by using [Niko's script](https://github.com/google/fonts/issues/4405#issuecomment-1080664412) and then bumping the version number.

@bungeman Dave Crossland asked me to fix this for every VF in our collection. I've just read through https://github.com/MicrosoftDocs/typography-issues/issues/922 and apart from the woff2 issue, are there any other components that also need updating? I'm assuming that enabling these bits will allow stroked glyphs with overlaps to render correctly? which platforms/rasterizers will be fixed by enabling these bits? happy to make tests if you don't have the answers.

If we're happy that this solves the issue, I'll update all the other VFs in our collection and add this functionality to the build chain.

Part fixes https://github.com/google/fonts/issues/4405


bungeman:
So the WOFF2 fix was needed just to transmit these bits at all and was a longstanding issue which needed to be fixed anyway. The issue with overlap when stroking and the proposal to use this bit as a signal that the outlines should be automatically simplified (overlap removed) when this bit is set gave a push to implement the WOFF2 change. Note that there are reasons other than simplifying the outline for stroking that this bit should be set (it may change the fill rule from even-odd to non-zero, or select perfect coverage instead of coverage counting rasterization) so it's still a good idea to set it in any event.

Currently (afaik) the only implementation using this bit as a signal to remove overlap in outlines is Skia's FreeType port. This means Android system and Skia on Linux with new enough FreeType. This includes Chrome on Linux, Chrome on Android, and the Android WebView. This went in to Chrome M102, so you'll currently need a Dev or Canary to see the overlap removal change in Chrome or the WebView.

The OpenType issue has been opened to get a general feel for how implementations would like to handle the stroking issue. Currently DirectWrite and CoreText do not surface this bit in their public APIs, which is why Skia has not attempted to implement this yet on Windows or Mac (in theory Skia can go looking for this bit, but it would be best to get this information directly). It may be that these implementations wish to not surface the bit and simplify the outline automatically themselves. It may be that no one wants to use or expose this bit for some reason. It may be that there is some extremely fast way to detect that there is overlap and the bit isn't really needed (though this may or may not be overriding the font author's intent).

davelab6:
Note to GF eng team, this builds on the OFL license change, which wasn't yet pushed in the API.

m4rc1e:
Going to close this since this family is too popular for testing purposes. I've made a pr for Maven Pro instead #4539

davelab6:
@m4rc1e did the Maven release go thru?

PR #4494, opened 2022-04-12
fixed vertical metrics to 120%UPM

bghryct:
Fixed the Vertical metrics to match spline sans mono, and follow 120%UPM metrics best practices as laid out in the docs.


RosaWagner:
re-made PR with packager in #4497

PR #4505, opened 2022-04-14
Noto Sans Test: Version 1.001 added

simoncozens:
81dce37: [gftools-packager] Noto Sans Test: Version 1.001 added

* Noto Sans Test Version 1.001 taken from the upstream repo https://github.com/notofonts/test.git at commit https://github.com/notofonts/test/commit/.


simoncozens:
It works.

PR #4535, opened 2022-04-20
Andika: Version 6.101 added

emmamarichal:
73ba3c3: [gftools-packager] Andika: Version 6.101 added

* Andika Version 6.101 taken from the upstream repo https://github.com/silnrsi/font-andika at commit https://github.com/silnrsi/font-andika/commit/.


emmamarichal:
What we have to do here:
- De-list Andika New Basic but let it remain available to existing sites.

- Replace the outdated single-weight Andika family with this v6.101 version.

ofl/andika/METADATA.pb
```diff
@@ -1,5 +1,5 @@
 name: "Andika"
-designer: "SIL International"
+designer: "SIL International, Victor Gaultney, Annie Olsen, Julie Remington, Don Collingsworth, Eric Hays, Becca Hirsbrunner"
```
RosaWagner, line 2: @jvgaultney can you confirm this is the correct credit for the fonts? Other possibility would be to have SIL would as the only credit, and all the individual authors mentioned in the description.

RosaWagner:
LGTM, we can merge once description is updated with the object of the upgrade

PR #4549, opened 2022-04-25
replace all "." with "۔"(Arabic full stop) for '_Arab' language sample texts

zhaoxiong0211:
Prevent tofu on fonts.google.com/noto


lang/Lib/gflanguages/data/languages/arb_Arab.textproto
```diff
@@ -8,13 +8,13 @@ sample_text {
   masthead_full: "يولد"
   masthead_partial: "جم"
   styles: "يولد جميع الناس أحرارًا متساوين في الكرامة"
-  tester: "يولد جميع الناس أحرارًا متساوين في الكرامة والحقوق. وقد وهبوا عقلاً"
+  tester: "يولد جميع الناس أحرارًا متساوين في الكرامة والحقوق  وقد وهبوا عقلاً"
```
davelab6, line 11: This replaces space period with space space, but HTML will turn the double into a single, so I think its better to replace period with nothing (removing it)

simoncozens: Agree, U+06D4 is the right fix. All Noto Arabics support that.

simoncozens: Urgh, hang on. There is one font in the library (Noto Kufi Arabic) which has the Arabic full stop but not the Latin full stop. There are 21 Arabic fonts in the library which have the Latin full stop but not the Arabic full stop. So using U+06D4 is good news for Noto but bad news for the rest of GF.

zhaoxiong0211: replaced with "۔" (Arabic full stop). PTAL

davelab6: > using U+06D4 is good news for Noto but bad news for the rest of GF

Currently these files are only used for the Noto specimen pages, so that's fine for now. 

The longer term fix is to add Latin full stop to the Noto web fonts and then change this back to Latin.

davelab6:
@khaledhosny however, Latin full stop isn't included in Noto fonts and creates tofu on the Noto website, so we want to eliminate that as a priority.

davelab6 (approved):

khaledhosny:
> @khaledhosny however, Latin full stop isn't included in Noto fonts and creates tofu on the Noto website, so we want to eliminate that as a priority.

Dropping the punctuation altogether would be the better option, U+06D4 looks odd in Arabic text.

khaledhosny:
Thanks for ignoring me.

zhaoxiong0211:
> Thanks for ignoring me.

Sorry Khaled, I did not mean to ignore you. When I merged it, I did not see your latest comment. I think we are going to have this fix going forward for now and eventually we will change it back to Latin full stop and have Arabic fonts to support Latin full stop.

davelab6:
> U+06D4 looks odd in Arabic text

Are there any other sentence separation options?

davelab6:
Hmm. putting this into GF `اا۔.اا ` ([like this](https://fonts.google.com/?query=noto&subset=arabic&preview.text=%D8%A7%D8%A7%DB%94.%D8%A7%D8%A7%20&preview.text_type=custom))  shows that indeed there's only 1 noto font family tofu'ing, the oldest one, and perhaps rather than pushing this change, we should just fix that one family.... @simoncozens how quickly could you prep a quick fix release to just add latin period to that 1 family?

Screen Shot 2022-04-26 at 5 25 38 PM

PR #4552, opened 2022-04-26
introducing_variable_fonts/content Add mixing axes

davelab6:
Per @dberlow suggestion to explain that axes can be combined throughout the introduction article


davelab6:
Possibly https://github.com/google/fonts/blob/main/cc-by-sa/knowledge/modules/introducing_type/lessons/introducing_variable_fonts/images/1.8.6.svg could be updated to show 2 axes, and a narrow thin and wide bold

PR #4560, opened 2022-04-27
Replace Noto Kufi Arabic with full build

simoncozens:
This adds the GF Latin Core subset of Noto Sans to Noto Kufi Arabic, allowing the use of the PERIOD glyph in sample text. (See #4549)


simoncozens:
OK, this is unfortunate.

We still don't cover GF Latin Core because my reference for Latin core glyphs, Noto Sans, doesn't actually cover GF Latin Core either. (Missing MINUS SIGN) And we still don't render the sample text - which was the whole point of the PR - because we're missing HYPHEN, which *is* in Noto Sans, but is *not* in GF Latin Core.

I can update metadata/name table entries/fsSelection/etc. but that will require a new upstream release, not just a new build.

I don't know why the vertical metrics have changed.

simoncozens:
I think this is as close as we're going to get for a Noto font right now.

davelab6:
I think @simoncozens you may be working off an old Latin Core def; @RosaWagner please could you confirm the latest GF Core/Kernel set to use for this purpose? :)

simoncozens:
Stuck in sandbox because I never labelled it. (Oops.)

simoncozens:
Actually it's not even in sandbox.

vv-monsalve:
> Actually it's not even in sandbox.

It still has no labels so would remain 'lost' (unregistered) in the onboarding workflow. 

Adding `to_sandbox` now.

PR #4581, opened 2022-05-03
Kantumruy Pro: Version 1.000 Added

sovichet:
Kantumruy Pro is a newly drawn design of the existing Kantumruy on Google Fonts. Kantumruy Pro comes with 7 weights and 2 styles (Upright & Italic) and is also available in the variable format. Latin character has been included, using the modified version of Work Sans.

- First release
- Publish only variable fonts to google/fonts upstream

Kantumruy Pro's GitHub: https://github.com/sovichet/kantumruy-pro


RosaWagner:
Thanks for the submission @sovichet.
May I ask how you generated the METADATA.pb and the upstream.yaml?

sovichet:
> May I ask how you generated the METADATA.pb and the upstream.yaml?

Hi Rosa. So this is my first time submitting fonts to the repo. I copied these files from other projects and edit them. Did I miss anything there?

RosaWagner:
Ah okay, we use tools to make sure there is no humain mistakes, and that the files are completely in sync. So I gonna re-do the PR with your info and the tooling. Although before I do, what justify the change of name for Kantumruy? Why not upgrading the existing font? 

- did the linespace change for the existing glyphs?
- did the vertical metrics changed?
- Is the general aspect of the already existing part of the font is super different from the new one?

sovichet:
> * did the line length change for the existing glyphs?

Yes, the line length is different. Also, the proportion is completely different. The existing Kantumruy has a very large proportion and made it looks bigger when used with other scripts like Latin, making user switch font sizes most of the time.

> * did the vertical metrics change?

Yes, it has different vertical metrics.

> * Is the general aspect of the already existing part of the font is super different from the new one?

Yes, the letterforms, proportions, character set, OpenType feature codes, etc. are different from the new one. I attached a simple PDF to show the obvious change in the new version.

[Some-differences-KantumruyPro.pdf](https://github.com/google/fonts/files/8619314/Some-differences-KantumruyPro.pdf)

RosaWagner:
Thanks for your quick reply :)
Okay so it's not the same font then. Is the idea to then hide the current kantumruy from the platform?

cc @davelab6 for approval about the new name

RosaWagner:
Sorry for all the questions, usually all of this is discussed in an issue (that would serve as archive for big decisions), and the PR is just a formality. If you have an issue to link where all of this is already approved by Dave, it can definitely help.

The same will apply to https://github.com/google/fonts/pull/4585, if you want to anticipate and add a comment with a PDF resuming the changes etc.

sovichet:
> Okay so it's not the same font then. Is the idea to then hide the current kantumruy from the platform?

Yes, that is the idea. 🙂

sovichet:
I close this PR, and post an issue instead.

davelab6:
Delisting the original is good

sovichet:
Yes, that's the idea. I had talked to Dave about this and I think it's a
good approach. :-)

PR #4585, opened 2022-05-04
kdamthmortwo: Version 1.002 Added

sovichet:
Taken from the upstream repo https://github.com/sovichet/kdam-thmor-two at commit https://github.com/sovichet/kdam-thmor-two/commit/2412dbe4d9dbe9567d049435f2b11513fa6f4742


sovichet:
@davelab6 That is also possible. Thank you!

@RosaWagner I'll rename the font to Kdam Thmor Pro as Dave suggested. I'll need to update the vertical metrics and opentypeOS2Selection setting as well since I found some bugs in the current config. So there will be an update in the binary files. Rosa, could you help guide me on how should I deal with this renaming? Should I also change the git repo name and then make another pull request by myself? (I've just realized that `gftools packager` is only for google/fonts contributors.)

sovichet:
Here's the difference in vertical metrics and line length between the two versions.

Screen Shot 2022-05-06 at 1 41 49 PM

RosaWagner:
@sovichet The best would be to close the two PRs, and to create 2 issues for each family in the issue tracker using the "add font" template. Then we will be able to prepare everything in your repo, and we will take care of the PR.

You can rename the repository as well, don't forget to update copyright string in sources + OFL.txt

sovichet:
@RosaWagner Awesome! Thank you, Rosa. I'll do it accordingly.

PR #4599, opened 2022-05-06
Fix bad glossary reference to /glossary/counters

rsheeter:
Dev push complains that this is a bad ref


davelab6:
@elliotjaystocks looks like a glossary entry for `counters` didn't get checked in yet; please make a planning note to come back to this PR and revert it (and add links to it from XTRA and perhaps other GFK pages) once the entry is finalized :)

rsheeter:
https://github.com/google/fonts/pull/4600 - very WIP, lmk if it reports false positives - suggests there maybe be additional offenders.

PR #4600, opened 2022-05-06
Add knowledge link check to CI

rsheeter:
Exploratory script that could be run as CI to sanity check knowledge links. Someone who knows what linkages are valid would need to sanity check. To try locally run `python .github/workflows/knowledge_graph.py` in a python3 venv with `absl-py` and `mistune` installed.

Sample failure: https://github.com/google/fonts/runs/6378007759?check_suite_focus=true


davelab6:
@m4rc1e please carry this through to completion, not urgent hto

PR #4614, opened 2022-05-09
Amiri Quran Colored: Version 0.114 added

yanone:
dfe1cec: [gftools-packager] Amiri Quran Colored: Version 0.114 added

* Amiri Quran Colored Version 0.114 taken from the upstream repo https://github.com/aliftype/amiri at commit https://github.com/aliftype/amiri/commit/4e06c7950304fdadd7180b85201e6eb0407ea380.


yanone:
The transformed components FAIL is outdated. My recent PR that treats unhinted fonts differently lets it pass. All other fails are acceptable IMO given that the font doesn't cover Latin at all.

RosaWagner:
Was it confirmed somewhere that it's okay that the font doesn't contain latin? Per requirements it should at least cover [Latin Kernel](https://github.com/googlefonts/glyphsets/blob/main/GF_glyphsets/Latin/txt/nice-names/GF_Latin_Kernel.txt) to avoid tofu on the API.

davelab6:
I believe Amiri in GF has a latin derived from the original Crimson Text. @khaledhosny we recently updated the original Crimson Text to fix a bunch of small issues, might be good to re-merge the latest into all verisons (targetting GF) of Amiri :)

khaledhosny:
Amiri Quran is a special purpose version of Amiri for typesetting Quran, it features a larger default line hight and few other changes that makes it less suitable for general text, and for this reason I purposefully only include the subset of Arabic that is needed for Quran typesetting to discourage it's use fir other purposes. I’m reluctant to include Latin in this subset for the same reason.

As for the name, “colored” pretty much discribes it's purpose, colors are used to tell vowel and other marks as well as ijam dots from the architype skeleton if the early writting. Colors are not used for any graphical effect other than being colored. I have also been using the name for a while now, so too late to change.

khaledhosny:
I wouldn't want it to come to that, and I wouldn't want to support more variants (people will eventually reach out to me when they have problems). But, if you are not specifically interested in Quranic typesetting, may be I should make a color version of Amiri proper? People occasionally ask for this.

RosaWagner:
@khaledhosny we made a [reduced latin glyphsets](https://github.com/googlefonts/glyphsets/blob/main/GF_glyphsets/Latin/txt/nice-names/GF_Latin_Kernel.txt) for these particular cases (on your suggestion). Would you consider to add it to the font? It is unfortunately still necessary to add the font in Google Workspace.

khaledhosny:
I’ve add the Latin kernel glyph set to the Quran fonts on main branch.

khaledhosny:
https://github.com/aliftype/amiri/releases/tag/0.115

RosaWagner:
thanks @khaledhosny !

RosaWagner:
@khaledhosny the file name is missing the style name, the font should be called `AmiriQuranColored-Regular.ttf`

RosaWagner:
Arf, forget about it, Yanone already automated the renaming in the upstream.yaml. All good.

khaledhosny:
> 🔥 FAIL: Glyph names are all valid? ([com.google.fonts/check/valid_glyphnames](https://font-bakery.readthedocs.io/en/stable/fontbakery/profiles/universal.html#com.google.fonts/check/valid_glyphnames))

You should use the fonts from the release zip, they have no glyph names at all.

khaledhosny:
I released 0.116 to fix a regression in the color glyphs https://github.com/aliftype/amiri/releases/tag/0.116

ofl/amiriqurancolored/DESCRIPTION.en_us.html
```diff
@@ -1,6 +1,54 @@
...
+</p>
+
+<p dir="rtl" lang="ar">
+يمكن الحصول على آخر إصدارات الخط الأميري من موقعه:
+</p>
+
+<p><a href="http://amirifont.org">amirifont.org</a></p>
+
 <p>
 To contribute, see <a href="https://github.com/alif-type/reem-kufi">github.com/alif-type/amiri</a>
```
khaledhosny, line 53: The link is wrong here.

RosaWagner: https://www.amirifont.org is the preferred option?

RosaWagner: Ah no you were talking about the github repo, on it.

RosaWagner:
Everything is now ready, the only remaining problem is the "Colored" in the font name, which by policy is not accepted (cf comment from Dave above). @khaledhosny What suffix would fit better to describe the colour fonction in this family?

RosaWagner:
I am sorry but for what I understood, this is not a choice. 

So two options from here;
- We can fork the repo to have a specific version for GF
- Or you can have an added line in your build that creates a second version for GF, so your original name is preserved and still distributed from our repository and website.

So far we have "ink", "fun", and "spice". 
Looking at your font, where the colour only act on diacritics, numbers and punctuation, I could imagine "highlight", "touch", "tone", "marked"… they are many possibilities that would describe the actual role of the colour, but only you can choose what it is best (but not coloured or related term).

…Please help me onboard this font?

RosaWagner:
Also no latin glyphs have colour applied except for the /i, can you explain this choice? (so we have it archived here in case the question raised later by users)

vv-monsalve:
I am under the impression that the use of color in this font is for cultural reasons and is intended to introduce recitation helpers. Is that the case @khaledhosny?

If so, to find a compromise between the cultural background of the font and Google requirements, could any of the following options be descriptive enough or valid names?
- Amiri Quran
- Amiri Quran Tajwid
- Amiri Red Vowels

Please bear in mind that we are not familiar with the Tajweed and Quran rules. The suggested names are only examples in the search for an alternate name. Perhaps you would like to provide more context details in the README section of the repo about the intention and use of the red-colored parts in the font to have a better understanding and try to reach a more suitable name.

khaledhosny:
If other options were suitable I would have used them (like the other color fonts I made, which I named before knowing there is a policy against “color” in font names).

I’d prefer to keep the existing name, but if that is really out of option then a plain “Amiri Quran” is the least-worst.

khaledhosny:
> Also no latin glyphs have colour applied except for the /i, can you explain this choice? (so we have it archived here in case the question raised later by users)
> 
> Capture d’écran 2022-06-30 à 19 41 04

That is a bug, I’ll see why it is happening.

khaledhosny:
> > Also no latin glyphs have colour applied except for the /i, can you explain this choice? (so we have it archived here in case the question raised later by users)
> > Capture d’écran 2022-06-30 à 19 41 04
> 
> That is a bug, I’ll see why it is happening.

Fixed in https://github.com/aliftype/amiri/releases/tag/0.117

khaledhosny:
If you don’t like my two name choices, I have nothing more to offer. The description was not written by me, but here is the NEWS entry when this font was first made available (`COLR`/`CPAL` were new back on 2015):
> A coloured version of the Amiri Quran font that gives the vowels and
      Quranic annotation marks distinctive colours. This font uses the new
      `COLR`/`CPAL` font tables which are currently supported only by Firefox
      (all platforms) and MS Internet Explorer/Edge (Windows 8.1 and above).

vv-monsalve:
> I’d prefer to keep the existing name, but if that is really out of option then a plain “Amiri Quran” is the least-worst.

@khaledhosny this option has been confirmed for the font name. Would you please change it to proceed with the publishing process? Thanks

khaledhosny:
> > I’d prefer to keep the existing name, but if that is really out of option then a plain “Amiri Quran” is the least-worst.
> 
> @khaledhosny this option has been confirmed for the font name. Would you please change it to proceed with the publishing process? Thanks

I already have an Amiri Quran upstream, so I can’t rename the colored version as well. The renaming is better handled on the GF side.

chrissimpkins:
I recognize that there is a family name debate and that Khaled has concerns about use of "Amiri Quran" for the build with color support because the build without color is released under that family name.  Dave wants to release the color build under the name "Amiri Quran".  Let's consider this decided and figure out how to make it happen so that we can release Khaled's work.

I took a look at Khaled's repo and this is the issue that he points out about the inability to change the family name upstream:

amiri

We are asking him to overwrite the no-color "Amiri Quran" font path with the color "Amiri Quran" font path, and to use the same family name and style in two different fonts.  That isn't possible.

The next best options IMO are either to add scripting to the upstream repo that revises the family name and file path as part of the G Fonts PR push process, or fork the repository and add the scripting there.  We're at an impasse.  Let's decide and move forward.

Other issues that require resolution here:

- Marc asked about Arabic description documentation in https://github.com/google/fonts/pull/4614#discussion_r916012993
- Khaled pushed a fix that is documented in https://github.com/google/fonts/pull/4614#issuecomment-1172262838, we'll need to make sure that the push includes this update

chrissimpkins:
Upstream PR to address path / naming issues described above https://github.com/aliftype/amiri/pull/237

ofl/amiriqurancolored/DESCRIPTION.en_us.html
```diff
@@ -1,6 +1,61 @@
...
+<p>
+<p dir="rtl" lang="ar">
+يهدف مشروع الخط الأميري إلى إحياء تقاليد و جماليات الطباعة العربية و موائمتها
+لتقنيات عصر الحواسيب، مع إتاحتها للعموم.
+</p>
+
+<p>
+<p dir="rtl" lang="ar">
+يمكن الحصول على آخر إصدارات الخط الأميري من موقعه:
+</p>
```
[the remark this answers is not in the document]
chrissimpkins: The current version of our docs on this issue: https://googlefonts.github.io/gf-guide/description.html#with-localised-text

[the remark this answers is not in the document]
khaledhosny: Please also note that this text is a description of Amiri font in general, it does not say any thing about the color font.

m4rc1e:
Closing since this pr has been superseded by #5048

PR #4621, opened 2022-05-10
Update METADATA.pb

marekjez86:
Adding all of the languages that notosanshk can handle

### DIFF
--- a/ofl/notosanshk/METADATA.pb
+++ b/ofl/notosanshk/METADATA.pb
@@ -28,8 +28,14 @@ axes {
   min_value: 100.0
   max_value: 900.0
 }
-languages: "hak_Hant"  # Hakka Chinese, Traditional
-languages: "lzh_Hant"  # Literary Chinese, Traditional
-languages: "nan_Hant"  # Min Nan Chinese, Traditional
-languages: "zh_Hant"  # Chinese (Traditional)
+languages: "cjy_Hant" # Chinese, Jinyu, Traditional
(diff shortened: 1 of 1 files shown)


simoncozens:
I think the reason these languages aren't in the METADATA.pb (and weren't generated by add-font) is because we don't have textproto records for them in our languages database.

I'm not familiar enough with the textproto files to know what this means and how these language records are used; things may be different between Noto and GF, for example. I don't know. But it feels to me that there probably isn't any point adding `cjy_Hant` to the languages when we don't have a `cjy_Hant.textproto`.

marekjez86:
I didn't realize it was generated
where is cjy_Hant.textprot kept?

simoncozens:
> where is cjy_Hant.textprot kept?

We don't have one; that's why I don't understand if there is a need to  add this language. If it did exist it would be [here](https://github.com/google/fonts/tree/main/lang/Lib/gflanguages/data/languages).

marekjez86:
Created https://github.com/google/fonts/pull/4627 to cover the missing languages for Traditional. Note that we have yue_Hans.textproto and yue_Hant.textproto .... I would expect these to be part of the Noto * HK metadata, but it weren't ? @simoncozens : why would it be missing from notosanshk  METADATA.pb ?

RosaWagner:
Closing this PR since Simon is working on the same thing in a separate one

simoncozens:
-> #4650.

PR #4634, opened 2022-05-11
Correcting upper/lowercase order of axes in API

elliotjaystocks:
This is mentioned in all 13 custom axis definitions


davelab6:
I believe this is a duplicate of https://github.com/google/fonts/pull/4624

PR #4635, opened 2022-05-11
Updating copy to reflect Figma’s support of VFs

elliotjaystocks:
@davelab6 As requested, I’ve updated the copy in this article to include Figma’s VF support. However, is saying that the support is “superlative” potentially critical of Adobe, etc.? I’d be up for changing that sentence.


RosaWagner:
@elliotjaystocks did you see Dave's answer? Should we merge this PR?

PR #4674, opened 2022-05-14
Gulzar: Version 1.000;[57fe3b0]; ttfautohint (v1.8.4) added

simoncozens:
9a986e6: [gftools-packager] Gulzar: Version 1.000;[57fe3b0]; ttfautohint (v1.8.4) added

* Gulzar Version 1.000;[57fe3b0]; ttfautohint (v1.8.4) taken from the upstream repo https://github.com/simoncozens/Gulzar at commit https://github.com/simoncozens/Gulzar/commit/7f736710573840523e68ea977843f7af7239632b.


simoncozens:
U+02BC is not part of GF Latin Core, so add-font is doing something wrong, but I will remove those languages from the METADATA.

simoncozens:
Arabic subset was not detected because we don't have the presentation form glyphs. I'll add it manually to METADATA, as well as crediting two more designers.

RosaWagner:
all languages should be removed from metadata.pb (except for Noto), they do that in prod now. (gftools has been upgraded accordingly)

PR #4690, opened 2022-05-18
Add designer profile for Simon Cozens

simoncozens:
It's a-me, Mario! Needed for #4674


simoncozens:
Thanks for the correction!

PR #4695, opened 2022-05-19
Commissioner: Version 1.000 added

RosaWagner:
3a93001: [gftools-packager] Commissioner: Version 1.000 added

* Commissioner Version 1.000 taken from the upstream repo https://github.com/kosbarts/Commissioner at commit https://github.com/kosbarts/Commissioner/commit/d18f549b6446df7e82bb811b389ce0cf24071b45.


RosaWagner:
@nathan-williams can you confirm we should remove the lang tags in metadata.pb of non-noto families when upgrading a font?

--> also added the custom axes in this one; are all the info correct?

cc @rsheeter

RosaWagner:
As recursive, the font was already onbaorded with the italic instances in the FVAR table + complete STAT table, so we can ignore the all these fails

RosaWagner:
yes :) let's merge and see what it looks like in sandbox

RosaWagner:
The axes need to be push to prod before being in one font's METADATA.pb, so reverting this PR.

RosaWagner:
Actually can't revert :/

PR #4703, opened 2022-05-23
Undo rollback, remove languages from METADATA (a* fonts)

andyjgf:


andyjgf:
Ignoring unrelated validation error.

PR #4715, opened 2022-05-25
updated push lists

RosaWagner:
@marekjez86 this is the list updated, you can remove what has already been pushed; otherwise let's merge and proceed with the push to prod ;)


marekjez86 (approved):
LGTM

PR #4723, opened 2022-05-27
Akshar: Version 1.100 added

yanone:
9924f0d: [gftools-packager] Akshar: Version 1.100 added

* Akshar Version 1.100 taken from the upstream repo https://github.com/tallchai/akshar-type at commit https://github.com/tallchai/akshar-type/commit/0c6de5ef9cef264b3b3f958e772b3a371f92d083.


yanone:
So please ignore the check on the vertical metrics, as that's the point of the upgrade.
Also the avar table check, as the instance progression is linear.

PR #4736, opened 2022-06-02
Designer added: Amin Abedi

emmamarichal:
Designer of Estedad project.

PR #4754, opened 2022-06-08
New catalog entry for Amin Abedi

yanone:


RosaWagner:
Of which font is he the designer? Can't find any in the catalog. Also, duplicate of #4736.

PR #4819, opened 2022-06-21
Noto Serif Dogra: Version 1.004 added

simoncozens:
abfa1c9: [gftools-packager] Noto Serif Dogra: Version 1.004 added

* Noto Serif Dogra Version 1.004 taken from the upstream repo https://github.com/notofonts/dogra.git at commit https://github.com/notofonts/dogra/commit/.


ofl/notoserifdogra/OFL.txt
```diff
@@ -1,30 +1,29 @@
-Copyright 2012 Google Inc. All Rights Reserved.
-
-This Font Software is licensed under the SIL Open Font License,
-Version 1.1.
+Copyright 2022 The Noto Project Authors (https://github.com/notofonts/noto-project-template)
```
RosaWagner, line 1: @simoncozens is this change of copyright string okay?

simoncozens: Yes, this is correct. (see Dave at https://github.com/googlefonts/fontbakery/pull/3681#issuecomment-1081954210)

simoncozens: Although, urgh, it should have customised the URL.

ofl/notoserifdogra/METADATA.pb
```diff
@@ -1,5 +1,5 @@
 name: "Noto Serif Dogra"
-designer: "Google"
+designer: "Ek Type"
```
RosaWagner, line 2: Not sure that this change is Ok on GF side

simoncozens: OK. Obviously we are pulling the designer info from the font, and currently the Designer field in Noto sources is the contracting designer. I think we want to keep that in the Noto source but change to "Google" when creating the METADATA.pb.

simoncozens:
(I'm going to do a re-release to fix the upstream copyright issue and make sure that the upstream.yaml generation now sets the designer over to Google.)

PR #4821, opened 2022-06-22
Noto Sans Glagolitic: Version 2.001 added

simoncozens:
3bd25e7: [gftools-packager] Noto Sans Glagolitic: Version 2.001 added

* Noto Sans Glagolitic Version 2.001 taken from the upstream repo https://github.com/notofonts/glagolitic.git at commit https://github.com/notofonts/glagolitic/commit/.


simoncozens:
(There's a bug in this version, so we're going to do a new one.)

PR #4822, opened 2022-06-22
Noto Serif Ahom: Version 2.005 added

simoncozens:
7281a70: [gftools-packager] Noto Serif Ahom: Version 2.005 added

* Noto Serif Ahom Version 2.005 taken from the upstream repo https://github.com/notofonts/ahom.git at commit https://github.com/notofonts/ahom/commit/.


vv-monsalve:
Version in Sandbox is still 2.003

PR #4837, opened 2022-06-27
Noto Serif Dogra: Version 1.005 added

simoncozens:
e62745f: [gftools-packager] Noto Serif Dogra: Version 1.005 added

* Noto Serif Dogra Version 1.005 taken from the upstream repo https://github.com/notofonts/dogra.git at commit https://github.com/notofonts/dogra/commit/.


vv-monsalve:
Version in Sandbox is still 1.003

vv-monsalve:
It shows tofu in Sandbox

Screen Shot 2022-08-05 at 12 59 39
Screen Shot 2022-08-05 at 12 52 23

vv-monsalve:
@simoncozens

simoncozens:
OK. So the Dogra sample text contains the glyph ऽ (U+934D DEVANAGARI SIGN AVAGRAHA). This is obviously outside the Dogra block, and the glyph should probably be added to the font subset. There is also another pretty critical bug with this fotn (https://github.com/notofonts/dogra/issues/9) so I think we should pause this version and I will re-release.

RosaWagner:
@simoncozens since the nam file was push, we can repush that family, is the bug fixed?

simoncozens:
There's a 1.006 coming which fixes the bug.

PR #4838, opened 2022-06-27
Noto Sans Buhid: Version 2.001 added

simoncozens:
18273e2: [gftools-packager] Noto Sans Buhid: Version 2.001 added

* Noto Sans Buhid Version 2.001 taken from the upstream repo https://github.com/notofonts/buhid.git at commit https://github.com/notofonts/buhid/commit/.


vv-monsalve:
Version in Sandbox is still 2.000

PR #4839, opened 2022-06-27
Noto Sans Avestan: Version 2.002 added

simoncozens:
208c5bb: [gftools-packager] Noto Sans Avestan: Version 2.002 added

* Noto Sans Avestan Version 2.002 taken from the upstream repo https://github.com/notofonts/avestan.git at commit https://github.com/notofonts/avestan/commit/.


vv-monsalve:
Version in Sandbox is still 2.001.

PR #4842, opened 2022-06-29
Borna Izadpanah: bio updated (Gulzar)

RosaWagner:
Request from the designer to update his bio.


vv-monsalve:
Info in Sandbox is still the old one

PR #4860, opened 2022-07-01
Designer added: Tina Anderson

emmamarichal:
Designer of Australian School Fonts project


vv-monsalve:
Hi @emmamarichal,

The designer's image is the preview of a font. Could you confirm it is the desired one?

image

RosaWagner:
Not good, we’ll have to wright to the designer after the holidays.

PR #4867, opened 2022-07-01
Noto Naskh Arabic: Version 2.013 added

simoncozens:
092f713: [gftools-packager] Noto Naskh Arabic: Version 2.013 added

* Noto Naskh Arabic Version 2.013 taken from the upstream repo https://github.com//notofonts/arabic.git at commit https://github.com//notofonts/commit/.


m4rc1e (changes requested):
- Figures have changed. Is this intentional? seems like a dflt locale change.
- Ng dots (ڭ) are too loose

before:

Screenshot 2022-07-07 at 12 48 39

after:

Screenshot 2022-07-07 at 12 50 33

vv-monsalve:
Confirming: would design also apply for this one?

Other than that, @m4rc1e could we merge it?

vv-monsalve:
> * The placement of dots in ڭ appears to be by design, since that's composed as a single glyph.
> * Farsi numerals are now the default for 06F4/06F6/06F7, and if you want Urdu you have to ask for them.
> 
> So both of these seem to be design choices.


Other than the above comment confirmation, @m4rc1e could we go ahead and merge it?

simoncozens:
> Confirming: would design also apply for this one?

That codepoint (U+06F4) represent the "Eastern Arabic-Indic" four, which could refer to either the Urdu or the Persian forms of the digit. The designer of this version made Persian the default rather than Urdu (although Urdu is still available via locl), which matches the Unicode code charts. I think this is fine.

m4rc1e (approved):

vv-monsalve:
LGTM

simoncozens:
2.014 incoming.

PR #4878, opened 2022-07-04
Noto Serif Oriya: Version 1.051 added

simoncozens:
b4cbc8c: [gftools-packager] Noto Serif Oriya: Version 1.051 added

* Noto Serif Oriya Version 1.051 taken from the upstream repo https://github.com/notofonts/oriya.git at commit https://github.com/notofonts/oriya/commit/.


ofl/notoseriforiya/METADATA.pb
```diff
@@ -0,0 +1,28 @@
...
+axes {
+  tag: "wght"
+  min_value: 400.0
+  max_value: 700.0
+}
+source {
+  repository_url: "https://github.com/notofonts/oriya.git"
+  archive_url: "https://github.com/notofonts/oriya/releases/download/NotoSerifOriya-v1.051/NotoSerifOriya-v1.051.zip"
+}
+is_noto: true
```
vv-monsalve, line 28: Languages are missing

vv-monsalve:
Oriya Serif version seems to be new, I've changed the labels

vv-monsalve:
@simoncozens This is local test with Oriya sample. It seems ok, but could you give them a look?
Desktop_Windows_10_edge_18 0_

simoncozens:
Missing U+035A

simoncozens:
OK, so I have a question about whether U+035A is actually necessary for the font or if the encoding of the sample texts is bad. To me it's kind of weird that an Indic script would require a general combining mark. And it turns out that the Oriya subjoined letter THA is pretty much exactly the same visual shape as "double combining ring below":

◌͚ ◌୍ଥ

So I do wonder if the Mundari text has been encoded "visually" instead of "semantically".

simoncozens:
Yeah, I think I'm going to conclude that the sample text is wrong, and this is OK to merge from my perspective.

RosaWagner:
Can you propose a change for the Mundari sample script?

vv-monsalve:
Mundari sample fixed in #4991

vv-monsalve:
@simoncozens, the article is missing in Sandbox and is not included (deleted) in this PR so that we can reinstate it as we did for Limbu 

Please restore it #5029

simoncozens:
This is a new font, so there is no article yet. Please generate one, @twardoch.

chrissimpkins:
Any update on this @twardoch?

twardoch:
https://github.com/notofonts/oriya/commit/a3b3484830f68391e8a278c57aa125650e35092d?w=1

simoncozens:
I'll PR these.

simoncozens:
-> #5249

chrissimpkins:
Thanks!

chrissimpkins:
Unblocked and requires a prod push

PR #4879, opened 2022-07-04
Noto Sans Sinhala: Version 2.006 added

simoncozens:
7b91be8: [gftools-packager] Noto Sans Sinhala: Version 2.004 added

* Noto Sans Sinhala Version 2.004 taken from the upstream repo https://github.com/notofonts/sinhala.git at commit https://github.com/notofonts/sinhala/commit/.

PR #4880, opened 2022-07-04
Noto Serif Sinhala: Version 2.004 added

simoncozens:
28d79c6: [gftools-packager] Noto Serif Sinhala: Version 2.004 added

* Noto Serif Sinhala Version 2.004 taken from the upstream repo https://github.com/notofonts/sinhala.git at commit https://github.com/notofonts/sinhala/commit/.


m4rc1e (dismissed):
Following string looks suspect:

"ත𑇤◌කෳ්ඁ

Screenshot 2022-07-06 at 15 08 24

Oh my my have tofus in the string here ;-)

vv-monsalve:
You may want to review the kerning also
kerns_new-cut

vv-monsalve:
@simoncozens could you confirm the reported cases above?

simoncozens:
Ohh. Turns out it's turning on the dlig feature and not telling me. Thanks, fontdiffenator!

simoncozens:
Right, I see the change and it's potentially a bug in the font. I'll work on it.

simoncozens:
New version incoming.

simoncozens:
Actually it'll probably just put the new version onto the same PR

vv-monsalve:
> Marc's cases are fine - it's the same issue as #4879; when there are two vowels on a base, we form a ligature glyph with the above vowel. I'm happy with that.

Ok. Thanks for the clarification.

New diff images look LGTM

vv-monsalve:
After some finding when running local test for the Sans version, I've run the same local tests for the Serif version.
@simoncozens despite not having a reflow in paragraphs for the serif variant, are the new glyphs ok?

**OS X High Sierra Safari_11**

**Windows 10 E
Desktop_Windows_10_edge_18 0_
dge_18**

simoncozens:
OK, there are a couple of glyph category issues here.

simoncozens:
Oh joy, I'm getting different results on different Harfbuzz versions.

RosaWagner:
@simoncozens if you think a font shouldn't go to sandbox yet, feel free to add the label "Blocked" so it doesn't end up in the sandbox list. In the PR fixing it, you can add in comment "unblocking #..." so I know which one should be tracked ;)

PR #4881, opened 2022-07-04
Noto Sans Oriya: Version 2.001 added

simoncozens:
fff64cc: [gftools-packager] Noto Sans Oriya: Version 2.001 added

* Noto Sans Oriya Version 2.001 taken from the upstream repo https://github.com/notofonts/oriya.git at commit https://github.com/notofonts/oriya/commit/.


simoncozens:
The can_render_samples fail is happening because of the lack of U+035A COMBINING DOUBLE RING BELOW. I don't know if this is really needed in Oriya/Odia but if so I can add it to the subsets. The sample text appears to be in the Mundari language.

PR #4882, opened 2022-07-04
Noto Sans Oriya: Version 2.001 added

simoncozens:
34323f6: [gftools-packager] Noto Sans Oriya: Version 2.001 added

* Noto Sans Oriya Version 2.001 taken from the upstream repo https://github.com/notofonts/oriya.git at commit https://github.com/notofonts/oriya/commit/.


m4rc1e (changes requested):
I get different results for the following string:

◌ସ‘୍ହଣ୫ଚ

Screenshot 2022-07-06 at 14 55 34

simoncozens:
Thank you. This is exactly the kind of build weirdness I was worried about. This actually works in the statics:
shape
but not in the variable font:
shape

Looks to be due to hateful rvrn substitutions.

simoncozens:
This will require me to implement the Glyphs "Feature for Feature Variations" custom parameter, what joy.

simoncozens:
I'm going to re-release this with the new toolchain.

PR #4897, opened 2022-07-05
Noto Serif Thai: Version 2.001 added

simoncozens:
2afbc3e: [gftools-packager] Noto Serif Thai: Version 2.001 added

* Noto Serif Thai Version 2.001 taken from the upstream repo https://github.com/notofonts/thai.git at commit https://github.com/notofonts/thai/commit/.


m4rc1e:
I'm not too well versed in the noto project so I'll do my best to review these for you. Apologies in advance but I'll probably ask lots of questions on decisions I don't know about. 

---

Anyways, Let's begin:



The rendering looks much better on Windows. I'm guessing this is because you've tweaked ttfa to use the correct script?


Everything else looks very good!

m4rc1e (approved):

PR #4899, opened 2022-07-07
Noto Sans Tamil: Version 2.003 added

simoncozens:
cc1face: [gftools-packager] Noto Sans Tamil: Version 2.003 added

* Noto Sans Tamil Version 2.003 taken from the upstream repo https://github.com/notofonts/tamil.git at commit https://github.com/notofonts/tamil/commit/.


simoncozens:
The dotted circle mark positions are computed automatically from the average of where they actually appear on the base glyphs. So this is correct.

vv-monsalve:
> The dotted circle mark positions are computed automatically from the average of where they actually appear on the base glyphs. So this is correct.

It seems it would be ready to merge. @m4rc1e could you approve the requested changes?

@simoncozens could you confirm the gdef_base are ok?
gdef_base_missing

m4rc1e (approved):

vv-monsalve:
> @simoncozens could you confirm the gdef_base are ok? In particular the numerals inclusion in the class.

@simoncozens I think this would be the only pending confirmation for Noto Sans Tamil

simoncozens:
Yes, these are fine as bases.

vv-monsalve:
> Yes, these are fine as bases.

Ok, merging this one here then.

PR #4902, opened 2022-07-07
Noto Serif Malayalam: Version 2.103 added

simoncozens:
b5e6241: [gftools-packager] Noto Serif Malayalam: Version 2.103 added

* Noto Serif Malayalam Version 2.103 taken from the upstream repo https://github.com/notofonts/malayalam.git at commit https://github.com/notofonts/malayalam/commit/.


m4rc1e (commented):
There are some shifting marks. Are these ok?

malayam_mark_shift

New version is the one with the dot approaching the bottom left.


The new font has slightly wider metrics as well.

simoncozens:
Mark shifting is fine; we're now attaching a mark that should always have been attached but wasn't before. Spacing looks better on the new fonts, but it looks like we may have lost some kerning.

simoncozens:
Please, let’s merge it for this release and try to fix it for the next. There are a bunch of known bugs that need to be fixed in Malayalam so I will be coming back to it, but I don’t want to hold up upgrades for them.

PR #4906, opened 2022-07-07
Noto Sans Gurmukhi: Version 2.003 added

simoncozens:
8179069: [gftools-packager] Noto Sans Gurmukhi: Version 2.003 added

* Noto Sans Gurmukhi Version 2.003 taken from the upstream repo https://github.com/notofonts/gurmukhi.git at commit https://github.com/notofonts/gurmukhi/commit/.


vv-monsalve:
> 🔥 FAIL: Checking OS/2 usWinAscent & usWinDescent. ([com.google.fonts/check/family/win_ascent_and_descent](https://font-bakery.readthedocs.io/en/stable/fontbakery/profiles/universal.html#com.google.fonts/check/family/win_ascent_and_descent))
🔥 FAIL OS/2.usWinAscent value should be equal or greater than 995, but got 896 instead [code: ascent]

As with the other cases, now that Use of Typo Metrics is enabled this win value could be increased to the yMax 

> 🔥 FAIL: Font has **proper** whitespace glyph names? ([com.google.fonts/check/whitespace_glyphnames](https://font-bakery.readthedocs.io/en/stable/fontbakery/profiles/universal.html#com.google.fonts/check/whitespace_glyphnames))
🔥 FAIL Glyph 0x00A0 is called "uni00A0.guru": Change to "uni00A0" [code: non-compliant-00a0]

Could you please confirm about this one?

vv-monsalve:
Something odd seems to happen in Safari Sierra 11 with some marks

vv-monsalve:
And some marks look off in Edge. 

Desktop_Windows_10_edge_18 0-cut

simoncozens:
Those flying marks are weird but I can't reproduce them!
shape
(see https://notofonts.github.io/gurmukhi/proof/text.html)

And I am not sure what's going on with Edge. Maybe it's putting Gurmukhi through the USE? Maybe check on https://notofonts.github.io/gurmukhi/proof/glyphs.html

vv-monsalve:
I'll inspect this on a Windows machine later. Edge also showed different results in [Noto Sans Kannada](https://github.com/google/fonts/pull/4912#issuecomment-1182544677) images (that is in Windows 10, Edge 18). But it was working ok when opening the HTML in Edge 103.0.1264.44.

However, It would be good to know what that version is doing to know about backward compatibility reaches.

simoncozens:
Edge 18 is "legacy" now, and latest Edge uses Harfbuzz.

m4rc1e:
> Edge 18 is "legacy" now, and latest Edge uses Harfbuzz.

If latest Edge is using HB these days then I have no problem. The only reason we check on these older browsers is because Browser stack doesn't support the newer ones.

RosaWagner:
@m4rc1e @simoncozens can you then confirm that we should ignore the screenshots from previous Edge version for this font?

simoncozens:
I confirm that the Edge mark positions should be ignored. However, the Safari 11 marks are something I want to investigate more.

simoncozens:
I've managed to confirm that the weird mark positioning in High Sierra *is* a real problem, not a browserstack artefact. I have tried to upgrade my VM to see if the problem is fixed by later High Sierra releases but I haven't managed to do so. I suspect the problem is due to the fact that the marks are not attached to the previous glyph, and should take their "natural" positions but aren't doing so. 

We have two options:

* Declare that this is a problem which only affects [4% of Mac users](https://gs.statcounter.com/macos-version-market-share/desktop/worldwide) and they should simply upgrade.
* Rework the font so that the marks are properly attached.

Obviously I'm in favour of the first. :-) But it's another "what kind of legacy systems do we support?" issue, so decision needed from @chrissimpkins and Dave.

chrissimpkins:
Approved for merge during our pipeline meeting today.

vv-monsalve:
@simoncozens Noto Sans Gurmukhi is missing the article

vv-monsalve:
Article reinstated in #5168

PR #4909, opened 2022-07-07
Noto Sans Gujarati: Version 2.102 added

simoncozens:
cb44080: [gftools-packager] Noto Sans Gujarati: Version 2.102 added

* Noto Sans Gujarati Version 2.102 taken from the upstream repo https://github.com/notofonts/gujarati.git at commit https://github.com/notofonts/gujarati/commit/.


vv-monsalve:
> 🔥 FAIL: Checking OS/2 usWinAscent & usWinDescent. ([com.google.fonts/check/family/win_ascent_and_descent](https://font-bakery.readthedocs.io/en/stable/fontbakery/profiles/universal.html#com.google.fonts/check/family/win_ascent_and_descent))
🔥 FAIL OS/2.usWinAscent value should be equal or greater than 995, but got 896 instead [code: ascent]

Since you are enabling the Use of Typometrics now this win value could use the yMax bbox to avoid clipping in Windows

vv-monsalve:
> 🔥 FAIL: Font has **proper** whitespace glyph names? ([com.google.fonts/check/whitespace_glyphnames](https://font-bakery.readthedocs.io/en/stable/fontbakery/profiles/universal.html#com.google.fonts/check/whitespace_glyphnames))
🔥 FAIL Glyph 0x00A0 is called "uni00A0.gujr": Change to "uni00A0" [code: non-compliant-00a0]

Could you confirm about this one?

vv-monsalve:
Are these marks ok?
glyphs_modified-cut

vv-monsalve:
> > 🔥 FAIL: Checking OS/2 usWinAscent & usWinDescent. ([com.google.fonts/check/family/win_ascent_and_descent](https://font-bakery.readthedocs.io/en/stable/fontbakery/profiles/universal.html#com.google.fonts/check/family/win_ascent_and_descent))
> > 🔥 FAIL OS/2.usWinAscent value should be equal or greater than 995, but got 896 instead [code: ascent]
> 
> Since you are enabling the Use of Typometrics now, this win value could use the yMax bbox to avoid clipping in Windows

I've [seen now](https://github.com/google/fonts/pull/4837#issuecomment-1167337737) that you don't want to change any published vertical metrics, however as said in the above comment, since the old `win` values matched with the `typo` values, and now you are enabling the Use of Typo Metrics, this change could be done without any risk to fulfill our spec on [upgrading fonts (case II)](https://googlefonts.github.io/gf-guide/metrics.html#2-recalculating-the-vertical-metrics-for-an-upgraded-family)

However, please tell me how you would like to proceed so that I wouldn't ask for this again in the coming Noto PRs.

vv-monsalve:
> The mark changes are OK.
> 
> I'd really like to get to a state where there are no fails (and that Noto fonts are just another font to be onboarded like any other, with no special treatment), so ideally we should change the usWinAscent and usWinDescent - if it is really true that we can do with no regressions. I just get scared about making any changes to vertical metrics. :-)

I'm performing just now the last tests for Word on Windows. I'll come back to you with advanced reasoning on this for these Noto projects. Do not change anything in the meanwhile, please :)

vv-monsalve:
After new tests performed in Word, PowerPoint, and NotePad in Windows, the following was showed:

- Word v.2206 - Uses de `typo` values, so win metrics don't affect the result
- PowerPoint v.2206 - Seems to not follow neither `typo` nor the `win` values. 
- NotePad - Uses only the `win` values, with or without the Use of Typo Metrics enabled. So this native software would be the problematic one in such case.

Given the above, for Noto Projects at least this update round, I would prefer to continue without changing the Win values for now.
We would need to review this at least with Marc to decide for further updates on them.

simoncozens:
That sounds good. Right now we are trying to onboard as much as we can of the current state of Noto. We can always fix more in the next release of each font, when hopefully there is less pressure and urgency...

vv-monsalve:
> Yes, the two marks above are correct. The first one makes sense because it's a ligature of another mark:

Great, thanks for the clarification. Merging this here then.

PR #4911, opened 2022-07-07
Noto Serif Kannada: Version 2.003 added

simoncozens:
068a24a: [gftools-packager] Noto Serif Kannada: Version 2.003 added

* Noto Serif Kannada Version 2.003 taken from the upstream repo https://github.com/notofonts/kannada.git at commit https://github.com/notofonts/kannada/commit/.

PR #4912, opened 2022-07-07
Noto Sans Kannada: Version 2.003 added

simoncozens:
581a93f: [gftools-packager] Noto Sans Kannada: Version 2.003 added

* Noto Sans Kannada Version 2.003 taken from the upstream repo https://github.com/notofonts/kannada.git at commit https://github.com/notofonts/kannada/commit/.


vv-monsalve:
Hi @simoncozens,

-  ಁ looks off position in Safari Sierra and Chrome Windows.
- And ᳲ in Chrome Windows
Desktop_Windows_10_chrome_71 0-cut

simoncozens:
These are fine. (See #4911 for candrabindu; ᳲ is behaving like a base so we made it one.)

vv-monsalve:
> These are fine. (See #4911 for candrabindu; ᳲ is behaving like a base so we made it one.)

Ok, merging this here then.

PR #4917, opened 2022-07-08
Noto Sans Devanagari: Version 2.003 added

simoncozens:
f691022: [gftools-packager] Noto Sans Devanagari: Version 2.003 added

* Noto Sans Devanagari Version 2.003 taken from the upstream repo https://github.com/notofonts/devanagari.git at commit https://github.com/notofonts/devanagari/commit/.


vv-monsalve:
LGTM

PR #4919, opened 2022-07-11
Noto Sans Balinese: Version 2.003 added

simoncozens:
6906c65: [gftools-packager] Noto Sans Balinese: Version 2.003 added

* Noto Sans Balinese Version 2.003 taken from the upstream repo https://github.com/notofonts/balinese.git at commit https://github.com/notofonts/balinese/commit/.


vv-monsalve:
LGTM

Desktop_Windows_10_chrome_71 0-CUT

PR #4922, opened 2022-07-12
Update readme to reflect size of ZIP file

RoelN:
Today's size of the ZIP file is 1.053.418.669 bytes (1,06 GB on disk).


RosaWagner:
Thanks for the edit.

PR #4934, opened 2022-07-13
Noto Sans Adlam: Version 3.001 added

simoncozens:
7e077be: [gftools-packager] Noto Sans Adlam: Version 3.001 added

* Noto Sans Adlam Version 3.001 taken from the upstream repo https://github.com/notofonts/adlam.git at commit https://github.com/notofonts/adlam/commit/.


vv-monsalve:
@simoncozens marks look off in Windows Chrome

simoncozens:
OK, so Adlam used to go through the Arabic shaper, which does not do abvm processing (only mark/mkmk processing), and now it goes through the USE shaper. Because it is a USE script now, ufo2ft puts the mark attachment rules into `abvm`/`blwm` instead of `mark`, so fonts like this will not shape correctly with old shaping engines.

I'm not sure how much I care. Harfbuzz changed to using USE for Adlam two years ago; that's probably long enough.

vv-monsalve:
> Harfbuzz changed to using USE for Adlam two years ago; that's probably long enough.

Don't know if we should have a confirmation on this @m4rc1e, @RosaWagner

vv-monsalve:
> 🔥 FAIL: Check samples can be rendered. ([com.google.fonts/check/metadata/can_render_samples](https://font-bakery.readthedocs.io/en/stable/fontbakery/profiles/googlefonts.html#com.google.fonts/check/metadata/can_render_samples))
> 🔥 FAIL Font can't render "fuf_Adlm" sample text:
> "𞤋𞤲𞥆𞤢𞤥𞤢 𞤢𞥄𞤣𞤫𞥅𞤶𞤭 𞤬𞤮𞤬 𞤨𞤮𞤼𞤭، 𞤲'𞤣𞤭𞤥𞤯𞤭𞤣𞤭 𞤫 𞤶𞤭𞤦𞤭𞤲𞤢𞤲𞥆𞤣𞤫 𞤼𞤮 𞤦𞤢𞤲𞥆𞤺𞤫 𞤸𞤢𞤳𞥆𞤫𞥅𞤶𞤭. 𞤉𞤩𞤫"
> [code: sample-text]
> ℹ INFO No sample_glyphs on METADATA.pb [code: no-samples]

@simoncozens, should `fuf_Adlm` be removed from languages in metadata?

simoncozens:
No, it should not. The problem is that the sample text uses an Arabic comma. It should use a reversed comma U+2E41, according to https://r12a.github.io/scripts/adlam/fuf.html and the Unicode proposal at http://std.dkuug.dk/jtc1/sc2/wg2/docs/n4628.pdf .

The font supports the reversed comma; the sample text is wrong.

vv-monsalve:
> The font supports the reversed comma; the sample text is wrong.

Ok, perhaps we should update the sample text then @m4rc1e?

RosaWagner:
sample text fixed in #4975

RosaWagner:
Ah since the sample text was fixed we thought we could review again, but actually we confirmation from the Chris if we ignore the legacy browsers or not — just like Noto Sans Adlam Unjoined. So go I go back to "needs confirmation".

vv-monsalve:
@simoncozens Noto Sans Adlam is missing the article

vv-monsalve:
Article reinstated in #5168

PR #4841, opened 2022-06-29
Mplus 1p: hotfix font name

RosaWagner:
Fixes #4731 

Apparently there was no metadata.pb before
also bumped up version number

PR #4935, opened 2022-07-14
Mplus1 hotfix font name

### LINKED ISSUE #4731: M Plus 1P static fonts Bold has broken name table
A user wrote in,

> with the M Plus 1P font.
> When downloading the font from Google Fonts and install it locally, we noticed that the MPlus 1P - Bold weight is mapped incorrectly as a separate font. (Please refer to the attached image)
> This issue happens also when we try to deploy the font inside our developed application.
> Due to this incorrect mapping, the Bold weight is not being read correctly and significantly affect our development as a result.
> Our team suspect that this issue is caused by the incorrect Meta information, which removed the Bold weight from the family.
> 
> Screenshot 2022-06-02 at 12 30 26

Should be a quick hot fix, or a complete upgrade to the latest family version

vv-monsalve:
Fixes #4731 after comments in #4841


vv-monsalve:
Solved in #4981

PR #4944, opened 2022-07-15
Noto Sans Hebrew: Version 2.003 added

simoncozens:
bc644f8: [gftools-packager] Noto Sans Hebrew: Version 2.003 added

* Noto Sans Hebrew Version 2.003 taken from the upstream repo https://github.com/notofonts/hebrew.git at commit https://github.com/notofonts/hebrew/commit/.


m4rc1e (approved):
LGTM.

Spacing is tighter but this is due to some design modifications

image

PR #4955, opened 2022-07-18
Noto Sans Mono: Version 2.010 added

simoncozens:
d7f7e70: [gftools-packager] Noto Sans Mono: Version 2.009 added

* Noto Sans Mono Version 2.009 taken from the upstream repo https://github.com/notofonts/latin-greek-cyrillic.git at commit https://github.com/notofonts/latin-greek-cyrillic/commit/.


ofl/notosansmono/METADATA.pb
```diff
@@ -1,7 +1,7 @@
 name: "Noto Sans Mono"
 designer: "Google"
 license: "OFL"
-category: "MONOSPACE"
+category: "SANS_SERIF"
```
RosaWagner, line 4: Category should be Monospace imo, but sans_serif can be added as a secondary one (so in first place)

m4rc1e (commented):
This PR deletes `ofl/notosansmono/article/ARTICLE.en_us.html`. I'm praying the packager isn't doing this.

simoncozens:
The archive packager *is* doing this; it deletes everything from the repo directory which isn't in the archive. Which I thought was a bug, but maybe it's OK, because if you replace statics with variables you do want to delete the statics. (But there's an argument to say that it should never delete metadata files.) The fix is to make sure I put the articles in the archive.

m4rc1e (commented):
1. The previous version's `post.isFixedPitch` was correctly set to 1. This update has 0. I think we should change it back to 1.
2. "ꭠ" width is incorrect. In Reg has 618, should be 600.

simoncozens:
But it... isn't fixed pitch?

m4rc1e:
Ms spec says:

>  Set to 0 if the font is proportionally spaced, non-zero if the font is not proportionally spaced (i.e. monospaced).

https://docs.microsoft.com/en-us/typography/opentype/spec/post

Isn't this a legit monospace so it should be 1, like the old family?

m4rc1e:
Thanks. Just checked the widths and yep you're correct. Ok, isFixedPitch is set correctly in this case. Ignore my first point.

simoncozens:
Updated

## Noto Sans Mono: Version 2.010 added

---
 57ee2d2: [gftools-packager] Noto Sans Mono: Version 2.010 added

* Noto Sans Mono Version 2.010 taken from the upstream repo https://github.com/notofonts/latin-greek-cyrillic.git at commit https://github.com/notofonts/latin-greek-cyrillic/commit/.

simoncozens:
Needs article, manual metadata change.

m4rc1e (approved):
Thanks for fixing the sakhayat adv width. LGTM for everything else.

chrissimpkins:
> No, see https://github.com/notofonts/latin-greek-cyrillic/issues/234 (this is what fontbakery was shouting about). This isn't a completely monospaced font - U+01C4 Ǆ, U+01CB ǋ and U+01C7 Ǉ (as well as apparently 243 other glyphs) are actually double-width (1200 units).

Is this desirable?  https://github.com/notofonts/latin-greek-cyrillic/issues/234 indicates no.

simoncozens:
Not really. But will require a drawing fix.

vv-monsalve:
> Not really. But will require a drawing fix.

Font has reached Sandbox, but blocking it now

simoncozens:
I don't think you need to block it because some glyphs aren't monospaced. The current production font suffers from the same problem, so this is still an upgrade.

vv-monsalve:
> The current production font suffers from the same problem, so this is still an upgrade.

Okay, I'll send it to prod when the Type Tester tab issue is solved (meaning when I can fully test it on Sandbox)

I'll leave the "Needs upstream resolution" label for issue tracking purposes, would you add a note on this in any other internal tracking system? (eg. spreadsheet?)

chrissimpkins:
Tomas is working on the Type Tester fix.

chrissimpkins:
v2.010 still requires a prod push

PR #4956, opened 2022-07-18
Noto Serif: Version 2.009 added

simoncozens:
119fb99: [gftools-packager] Noto Serif: Version 2.009 added

* Noto Serif Version 2.009 taken from the upstream repo https://github.com/notofonts/latin-greek-cyrillic.git at commit https://github.com/notofonts/latin-greek-cyrillic/commit/.


RosaWagner:
@simoncozens the article is deleted in each PR, and I didn't find record to know if it's okay or not. If there is one please direct me to it so we can proceed. 

If it's a mistake, I indicated all the merged PR with missing article with the label "needs -- Needs Meta/Desc/License changes". In that case I would ask for **2** fixing PRs: one correcting directories _not in sandbox yet_ (label "blocked"), and another one correcting directories _already in sandbox_ (label "in sandbox")

simoncozens:
I don't think we should be deleting the article. I would argue that this is a packager bug, because it shouldn't be deleting stuff unless it's told to. But there we are. I'll go through and give you two PRs to reinstate them.

simoncozens:
Please check to see if this has a Greek regression as well.

RosaWagner:
If you want to access the CI artifacts, you can directly go to the tab "checks" next to "commits" ;) There you can download the images resulting of the gftools QA

https://github.com/google/fonts/actions/runs/2711113211

simoncozens:
OK, so yes, this too has a redesigned (tighter) Greek.

Issue #4952, opened 2022-07-16
ofl/mplus1p: Filenames in METADATA.pb file do not match filesystem


nathan-williams:
**Describe the bug**
Filenames in METADATA.pb file do not match filesystem

**Expected behavior**
Filenames cited in METADATA.pb should match the actual filenames in the filesystem.

davelab6:
The filename and metadata values should match, but it's not always the case that the metadata values are wrong.

chrissimpkins:
@RosaWagner @vv-monsalve D/W Dave, Andy, and Nate.  The goal is consistency across file and family names and IIUC we should use the following approach as defined by Dave in https://github.com/google/fonts/pull/4841#issuecomment-1172938800:

- Font file names: `MPLUS1p-Regular.ttf`, `MPLUS1p-Medium.ttf`, ... (same approach for the rest of the family styles)
- Family name: `M PLUS 1p`

Will you please submit a new PR with changes to make the file paths, file path strings listed in METADATA.pb, and family name consistent with the definitions that I listed above? Ty!

vv-monsalve:
Done in #4976

PR #4957, opened 2022-07-18
Fix upper/lowercase for the rest of the filenames.

andyjgf:


RosaWagner:
@andyjgf can you please:

- give a bit more insight about this PR
- tell us what was wrong with #4935 although it was following instructions in #https://github.com/google/fonts/pull/4841#issuecomment-1172938800
- tell us if this needs to go to sandbox list or if it has been pushed already

vv-monsalve:
I've just seen issue #4952. Is this PR related to that?
When downloading the current fonts from the catalog I can see they are Uppercase. Should the filenames be Upper instead?

davelab6:
This looks incorrect to me and should be reverted

PR #4967, opened 2022-07-20
Delete thumbnail.svg

elliotjaystocks:


vv-monsalve:
@elliotjaystocks could you confirm this is Live now?

PR #4968, opened 2022-07-21
Figtree: Version 1.000 added

emmamarichal:
6b2cc3c: [gftools-packager] Figtree: Version 1.000 added

* Figtree Version 1.000 taken from the upstream repo https://github.com/erikdkennedy/figtree at commit https://github.com/erikdkennedy/figtree/commit/10a64f0df00a6b098639cce402ab4b60d519bd1d.

 3101c72: Description added - but to complete

PR #4969, opened 2022-07-21
Designer added: Erik Kennedy

emmamarichal:
Designer of Figtree project. https://github.com/google/fonts/pull/4968


catalog/designers/erikkennedy/bio.html
```diff
@@ -0,0 +1,2 @@
+<p>Erik D. Kennedy is a UX/UI designer who's worked with clients ranging from Fortune 100 to Y-Combinator startups. He's also taught thousands of students around the world via his online school, Learn UI Design. He lives in Seattle, WA and spends his free time doing Brazilian jiu-jitsu and playing Flamenco guitar.</p>
```
vv-monsalve, line 1: should we use "who is" instead of informal contraction who's?

RosaWagner: I guess it would be better yes.

PR #4971, opened 2022-07-21
Reinstate missing articles for Noto fonts in sandbox

simoncozens:


simoncozens:
This is a bad PR, I'm sorry; I scripted it and got the paths wrong. Please revert.

PR #4974, opened 2022-07-21
Reinstate missing articles for Noto fonts in sandbox, take 2

simoncozens:
- Revert "Reinstate missing articles (#4971)"
- Reinstate missing articles


RosaWagner:
Noto Sans Ethiopic shouldn't be in that PR

PR #4977, opened 2022-07-22
Inter Tight: Version 3.002 added

RosaWagner:
988c87e: [gftools-packager] Inter Tight: Version 3.002 added

* Inter Tight Version 3.002 taken from the upstream repo https://github.com/rsms/inter-gf-tight at commit https://github.com/rsms/inter-gf-tight/commit/349fd3ba66c011691dcb74561bd6a88e6a4045b0.

 44190d0: Inter Tight: added description


vv-monsalve:
On an email thread Rasmus confirmed the axis working as intended:

> This looks as intended. On Brex’s request, we decreased spacing of weights 400–900 but not 100–400, which is why we’re seeing the effect of “spacing increasing” as weight falls below 400.

PR #4978, opened 2022-07-22
Noto Rashi Hebrew is a serif

simoncozens:
This:

* Changes the METADATA.pb category of Rashi Hebrew to be SERIF.
* Pairs the Hebrew with Noto Serif instead of Noto Sans.


RosaWagner:
Is the number version higher ?

simoncozens:
I didn't increase the version number on the Noto side because it's the same base font, just a new subset. I didn't think GF was too fussy about version numbers.

RosaWagner:
It happened two or three times in the past that a font didn't get updated because the version number was the same as the previous version: I don't exactly know the push system from the eng team but I suspect same version number disturbs the tooling. Would be good to know better about that.

In any case, we are monitoring about 80 projects at the same time, the version number is the only info we have to make sure the font was actually updated in sandbox.

simoncozens:
OK, let's close this and I'll do a new release

PR #4979, opened 2022-07-22
Noto Serif Sinhala: Version 2.005 added

simoncozens:
428adee: [gftools-packager] Noto Serif Sinhala: Version 2.005 added

* Noto Serif Sinhala Version 2.005 taken from the upstream repo https://github.com/notofonts/sinhala.git at commit https://github.com/notofonts/sinhala/commit/.


RosaWagner:
🔥 FAIL: Font has **proper** whitespace glyph names? ([com.google.fonts/check/whitespace_glyphnames](https://font-bakery.readthedocs.io/en/stable/fontbakery/profiles/universal.html#com.google.fonts/check/whitespace_glyphnames))
🔥 FAIL Glyph 0x00A0 is called "uni00A0.sinh": Change to "uni00A0" [code: non-compliant-00a0]

This fail made me look into the non breakable space, and the situation is a bit weird. There is a nbspace.sinh that is the same width as the space glyph. And there is a non-localised nbspace that is not the same width. They should all be the same width, so I would also question the necessity of a localised one.

To anticipate any question: we have to enforce same width for space and ndspace in fonts otherwise they are not eligible to be added to Workspace.

simoncozens:
Closing, 2.006 incoming.

vv-monsalve:
> 🔥 FAIL: Font has **proper** whitespace glyph names? ([com.google.fonts/check/whitespace_glyphnames](https://font-bakery.readthedocs.io/en/stable/fontbakery/profiles/universal.html#com.google.fonts/check/whitespace_glyphnames)) 🔥 FAIL Glyph 0x00A0 is called "uni00A0.sinh": Change to "uni00A0" [code: non-compliant-00a0]
> 
> This fail made me look into the non breakable space, and the situation is a bit weird. There is a nbspace.sinh that is the same width as the space glyph. And there is a non-localised nbspace that is not the same width. They should all be the same width, so I would also question the necessity of a localised one.
> 
> To anticipate any question: we have to enforce same width for space and ndspace in fonts otherwise they are not eligible to be added to Workspace.

Oh Lord, Should we review previously merged PR then?. This Fail was previously ignored for some fonts. E.g. [Noto Serf Dogra](https://github.com/google/fonts/pull/4837#issuecomment-1167333066)

simoncozens:
Dogra's (probably) not going to workspace, so it's not a major one, but in general yes, we probably should (sorry).

PR #4983, opened 2022-07-25
Noto Rashi Hebrew: Version 1.006 added

simoncozens:
8b72fcc: [gftools-packager] Noto Rashi Hebrew: Version 1.006 added

* Noto Rashi Hebrew Version 1.006 taken from the upstream repo https://github.com/notofonts/hebrew.git at commit https://github.com/notofonts/hebrew/commit/.


RosaWagner:
Can you confirm the accent positioning is alright? 
marks_modified

simoncozens:
Yes, this is an improvement. The dots which appear to be flying off to the left are SHIN DOT which should only attach to the letter SHIN.

RosaWagner:
Thanks, all good then

PR #4984, opened 2022-07-25
Noto Sans Oriya: Version 2.002 added

simoncozens:
e23582c: [gftools-packager] Noto Sans Oriya: Version 2.002 added

* Noto Sans Oriya Version 2.002 taken from the upstream repo https://github.com/notofonts/oriya.git at commit https://github.com/notofonts/oriya/commit/.


simoncozens:
Article aargh aargh aargh.

PR #4989, opened 2022-07-27
Noto Serif Sinhala: Version 2.006 added

simoncozens:
8ac9f3e: [gftools-packager] Noto Serif Sinhala: Version 2.006 added

* Noto Serif Sinhala Version 2.006 taken from the upstream repo https://github.com/notofonts/sinhala.git at commit https://github.com/notofonts/sinhala/commit/.


vv-monsalve:
LGTM

vv-monsalve:
Arf! I don't see the article included. 
@simoncozens any particular reason? was it packager again?

RosaWagner:
If the article is not included in the directory, then it appears as removed in the PR. Here it is not in the changed file, so there is a great chance it was not deleted, we can check it out in the directory directly to be sure.

simoncozens:
Article and description *are* in the archive and in the `upstream.yaml`. I think it has not changed.

PR #4991, opened 2022-07-27
Specimen: Fix Mundari sample text

simoncozens:
Mundari is a language spoken by 1.6 million people in eastern India, Bangladesh and Nepal. It is written in Mundari Bani script but is also written in transliteration in the local scripts in its area of use, including Odia (Oriya). The samples for Mundari were prepared by Biswajit Mandal, who found a single book describing the Oriya transliteration the Mundari glottal sound as a double-circle sign underneath the glyph.

In his encoding of these sample texts, he used U+0351 COMBINING DOUBLE RING BELOW to denote the glottal sign: ◌͚
However, a more "natural" way to make this graphical sign in Oriya script is to use the subscript letter THA: ◌୍ଥ

I believe (and from what I can tell, [Biswajit agrees](https://twitter.com/Biswaji13480606/status/1551848439323201537?s=20&t=TB7yKT8dTQY2QSkVj1Zq1Q)) that using virama+THA is a suitable encoding for this sign, instead of using a combining character.


RosaWagner:
It looks documented enough so let's merged. If we have any complaint, we will correct it back.

simoncozens:
I'm sorry, I messed up - this fix should have been done on the `lang` repo, not here.

PR #5003, opened 2022-07-29
Sono: Version 2.112 added

emmamarichal:
6c5604a: [gftools-packager] Sono: Version 2.112 added

* Sono Version 2.112 taken from the upstream repo https://github.com/sursly/sono at commit https://github.com/sursly/sono/commit/0e7274983bb034e232c9903f9d827ba4d595a888.

 b8ae605: Sono: Description added / to complete

 bb2a534: Metadata.pb: added registry_default_overrides


emmamarichal:
Waiting for a longer description from Tyler

RosaWagner:
I think we are good with the description, if Tyler wants to update it, we can do that later.
I removed the added linespaces in METADATA.pb, not sure that the protobuf format is allowing them. 
And since the font is not really monospace, I made that category secondary.

This font is important to check in dev-sandbox to see if the default position is the right one, and if the API generates proportional styles in the zip package.

RosaWagner:
After checking the font in dev-sandbox: the default_overrides works for the type tester but not necessarily for the zip file. Indeed he statics generated are `Sono_Monospace` and `Sono`, instead of `Sono` and `Sono_Proportional` since we wanted Mono to be the default.

cc @chrissimpkins @davelab6

davelab6:
I'm happy to push this to prod as-is

The problems with the ZIPs caused by the semantic scope creep is on @evanwadams 's radar and this might be solved when he addresses them :)

PR #5005, opened 2022-07-29
Sample text: acircumflexdotbelow in vietnamese specimen

RosaWagner:
Every fonts supporting vietnamese are displaying a fallback font instead of the actual `a circumflex dot below` (uni1EAD).
Indeed, looking into vi_Latn.textproto, the actual glyphs used is a ligature between `a circumflex` and `dotbelowcomb` (acircumflex_dotbelowcomb), and not an actual codepoint.

This in Pinyon Script in Sandbox, but you can find the same issue in a majority of fonts supporting vietnamese:
Capture d’écran 2022-07-29 à 11 35 39

This PR replaces the ligature glyphs by the encoded one.

### DIFF
--- a/lang/Lib/gflanguages/data/languages/vi_Latn.textproto
+++ b/lang/Lib/gflanguages/data/languages/vi_Latn.textproto
@@ -17,14 +17,14 @@ exemplar_chars {
 sample_text {
   masthead_full: "TtÂâ"
   masthead_partial: "Cc"
-  styles: "Việc thừa nhận nhân phẩm vốn có, các quyền"
+  styles: "Việc thừa nhận nhân phẩm vốn có, các quyền"
   tester: "Sự xâm phạm và coi thường nhân quyền đã dẫn đến những hành động"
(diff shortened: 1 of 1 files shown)


vv-monsalve:
Interesting finding

> Indeed, looking into vi_Latn.textproto, the actual glyphs used is a ligature between `a circumflex` and `dotbelowcomb` (acircumflex_dotbelowcomb), and not an actual codepoint.

It seems there are many other cases like this in the specimen, e.g. the use of `o circumflex` and `dotbelowcomb` (ocircumflex_dotbelowcomb), and not an actual codepoint `1ED9`

Great Vibes in catalog now despite the font has the codepoint
Screen Shot 2022-07-29 at 18 51 07

I think we'll need to inspect this further and fix all of them.

RosaWagner:
Add to the the PR ;)

vv-monsalve:
Yes, that's the idea

vv-monsalve:
Moved to `googlefonts`. since changes to `lang` directory must be done first there.

PR #5019, opened 2022-08-01
Noto Naskh Arabic UI: Version 2.014 added

simoncozens:
c50b8e8: [gftools-packager] Noto Naskh Arabic UI: Version 2.014 added

* Noto Naskh Arabic UI Version 2.014 taken from the upstream repo https://github.com/notofonts/arabic.git at commit https://github.com/notofonts/arabic/commit/.


simoncozens:
This is not a priority for now; the release needed to happen for Android, and the GF PR was created automatically. It can be merged another time.

chrissimpkins:
> ~This is not a priority for now; the release needed to happen for Android, and the GF PR was created automatically. It can be merged another time.~ Actually because it needs to happen for Android, it _does_ need to be onboarded here too.

Unrelated to an Android push. This is a high priority PR.  Please prioritize accordingly. Ty!

chrissimpkins:
@simoncozens Do we want to delete the article content in this PR?

simoncozens:
It’s an oversight but as it happens, probably we do, to be honest. We don’t serve this family on the Noto site.

vv-monsalve:
> It’s an oversight but as it happens, probably we do, to be honest. We don’t serve this family on the Noto site.

Deleting the article for projects in the past is what led the description to be removed from the "About" section in Sandbox. Confirming: Do we go ahead without the article?

vv-monsalve:
@simoncozens is this ok?

> 🔥 FAIL: Ensure dotted circle glyph is present and can attach marks. ([com.google.fonts/check/dotted_circle](https://font-bakery.readthedocs.io/en/stable/fontbakery/profiles/universal.html#com.google.fonts/check/dotted_circle))
> 🔥 FAIL The following glyphs could not be attached to the dotted circle glyph:
> 
> uni065B
> 
> And uni0654 [code: unattached-dotted-circle-marks]
> 
> uni065B
> 
> And uni0654 [code: unattached-dotted-circle-marks]

vv-monsalve:
I've seen Latin script has been added, but the `qa` artifacts don't included a diffenator report. So I can't inspect if there could be a difference from the Arabic, as we have spotted in previous cases.

Does the `UI` in the PR's name imply that this project is addressed for a different environment from the API? 
For the API, it would be a font upgrade, but if the before is true, then the checks are taking this as a new font and no comparisons have been done. Is that the case?

On the other hand, the current version published says "Noto Sans Arabic is an unmodulated (“sans serif”) design for texts" but the Latin companion included in the font is Serif.

Desktop_Windows_10_chrome_71 0-CUT

chrissimpkins:
> Does the UI in the PR's name imply that this project is addressed for a different environment from the API?

Yes, the UI projects are not released in the fonts.google.com API.  The Noto Naskh Arabic UI project is in production and the changes here represent an update to the previous production version.  Are you able to run the fontdiffenator tool locally if I provide the current production fonts for comparison?

chrissimpkins:
Are you able to run fontdiffenator locally if I provide the current production fonts?

chrissimpkins:
> Confirming: Do we go ahead without the article?

Let's replace it

chrissimpkins:
Re: https://github.com/google/fonts/pull/5019#issuecomment-1204107613

Given the priority of the changes here, if these are not regressions vs. the last production version, then let's move ahead with a release without addressing them now.

chrissimpkins:
@vv-monsalve from @davelab6:

> If the fix has Latin, I'd subset away the Latin and deliver that to the team to unblock them and let Simon figure out the long term process to prevent regression

vv-monsalve:
> Are you able to run fontdiffenator locally if I provide the current production fonts?

I've downloaded the `Noto Naskh Arabic` font published in the catalog and run diffenator locally. 

There are some missing marks glyphs reported among others

glyphs_missing

marks_missing

Full report zip: 
[Noto-Naskh-Arabic-diffenator.zip](https://github.com/google/fonts/files/9255358/Noto-Naskh-Arabic-diffenator.zip)

vv-monsalve:
I've also run`gen-html` tests to compare them. Since the `UI` font differs from the fonts.google catalog, this is a comparison would be just as an overall reference for Arabic between the two fonts.

Noto_Naskh_Arabic-comparison

vv-monsalve:
> @vv-monsalve from @davelab6:
> 
> > If the fix has Latin, I'd subset away the Latin and deliver that to the team to unblock them and let Simon figure out the long term process to prevent regression

@chrissimpkins, I've hidden the comment about the Latin as resolved since it was an error on my side inspecting the description of `Noto Sans Arabic` instead of `Noto Naskh Arabic` description at that moment. The latter is indeed a Serif font, so I don't think we should remove the Latin even if it's not needed for the UI.

The Arabic regressions reported are still legit, though.

chrissimpkins:
Let's close this.  I have a fixed font based on the current production font that should work. I'll open a new PR with the file.

PR #5027, opened 2022-08-04
Lateef: Version 2.000 added

emmamarichal:
202f596: [gftools-packager] Lateef: Version 2.000 added

* Lateef Version 2.000 taken from the upstream repo https://github.com/silnrsi/font-lateef at commit https://github.com/silnrsi/font-lateef/commit/.

 54f8a9c: Updated description according the update


vv-monsalve:
There is a horizontal regression shown in Diffbrowser images.

However, this is a SIL project. @m4rc1e?

m4rc1e:
Weight and width has changed quite a bit

vv-monsalve:
> This is happening because the family has been expanded from having just a Regular style into a fully fledged family which contains ExtraLight-ExtraBold styles.

True, I considered it. However, usually, we expect the Regular to be consistent with the single font previously released, and it has changed significantly indeed.

chrissimpkins:
To my eye it seems that the weight change is in Latin only, not in the Arabic.  Is that the case and is this an attempt to address color matching @ regular weight across the scripts?

m4rc1e:
Just asked Dave and he's happy for us to merge.

PR #4932, opened 2022-07-13
Noto Serif Hebrew: Version 2.003 added

simoncozens:
2268a64: [gftools-packager] Noto Serif Hebrew: Version 2.003 added

* Noto Serif Hebrew Version 2.003 taken from the upstream repo https://github.com/notofonts/hebrew.git at commit https://github.com/notofonts/hebrew/commit/.

PR #5033, opened 2022-08-04
Update to_production.txt

vv-monsalve:
Removing `ofl/notosanshebrew` until we confirm the version number

### DIFF
--- a/to_production.txt
+++ b/to_production.txt
@@ -25,7 +25,6 @@ ofl/notorashihebrew # https://github.com/google/fonts/pull/4983
 ofl/notosansavestan # https://github.com/google/fonts/pull/4839
 ofl/notosansbalinese # https://github.com/google/fonts/pull/4919
 ofl/notosansbengali # https://github.com/google/fonts/pull/4888
-ofl/notosanshebrew # https://github.com/google/fonts/pull/4944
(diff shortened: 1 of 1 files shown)


rsheeter:
Bit late, I already initiated push. What will go awry if it goes out ?

chrissimpkins:
> Bit late, I already initiated push. What will go awry if it goes out ?

All good.  https://github.com/google/fonts/pull/4944

There was a major version change.  It seems that we are ok.

vv-monsalve:
> There was a major version change. It seems that we are ok.

Is there green light to move Noto Serif Hebrew to production then?

vv-monsalve:
> I already initiated push.

@rsheeter were all the projects on the list pushed?

chrissimpkins:
> > There was a major version change. It seems that we are ok.
> 
> Is there green light to move Noto Serif Hebrew to production then?

v2.003? https://github.com/google/fonts/pull/4932

vv-monsalve:
> v2.003? #4932

Yes, that's the version of the font in Sandbox. Version confirmation would be the only pending issue to move it to production list.

chrissimpkins:
OK to move v2.003 from sandbox to production. That appears to be the latest update. Thanks Viviana!

PR #5035, opened 2022-08-05
Fix broken link in gfk::introduction_ar_vr

nathan-williams:


vv-monsalve:
Hi @nathan-williams, please always include the PR to the Traffic Jam project board and add the labels so that the PR gets tracked into the onboarding workflow. Otherwise, it would get ignored by the system.

rsheeter:
Why didn't the [graph check](https://github.com/google/fonts/blob/main/.github/workflows/knowledge_graph.py) detect this? Can you update it so it does?

Issue #4566, opened 2022-04-28
Fix Arabic sample texts again


simoncozens:
#4549 replaced the full stop in Arabic sample texts with an Urdu full stop; this was to stop Noto fonts without the Latin full stop from showing tofu. However, this only affected one font (Noto Kufi Arabic), which I fixed by adding a Latin subset to it. When that goes into production, we should revert #4549.

khaledhosny:
This text is now shown for all fonts not only Noto, so please give this a higher priority.

simoncozens:
Done (and working its way through the system).

khaledhosny:
Thanks!

PR #5049, opened 2022-08-10
Revert "replace all "." with "۔"(Arabic full stop) for '_Arab' language sample texts (#4549)"

simoncozens:
This reverts commit f061ebc955235b29dbb2f5f9b2623a06e5df0b8d. See discussions in #4549 and #4566.


vv-monsalve:
> When that goes into production, we should revert https://github.com/google/fonts/pull/4549.

From that comment on #4566, should this PR go to sandbox only after #4560 hits production?

simoncozens:
Yes, that's right!

PR #5058, opened 2022-08-11
Roboto: Version 3.005; 2022 added

m4rc1e:
b7a918c: [gftools-packager] Roboto: Version 3.005; 2022 added

* Roboto Version 3.005; 2022 taken from the upstream repo https://github.com/googlefonts/roboto-classic at commit https://github.com/googlefonts/roboto-classic/commit/.


m4rc1e:
@vv-monsalve just a heads up, we need to keep this family off the push lists.

m4rc1e:
@kpozin lmk if you need anything else.

Issue #5054, opened 2022-08-10
Add region: "CN"


vv-monsalve:
The field `region: "CN"` should be added to the following .textproto files for them to be listed under Chinese languages.

Ref #4722 

- [x] "hak_Hant" # Hakka Chinese, Traditional
- [x] "lzh_Hant" # Literary Chinese, Traditional
- [x] "yue_Hant" # Cantonese
- [x] "za_Hant" # Zhuang, Traditional Han

vv-monsalve:
Solved in #5088

PR #4722, opened 2022-05-26
normalized all chinese METADATA.pb

marekjez86:
normalized all chinese METADATA.pb

PR #5063, opened 2022-08-11
"CN" region added to Chinese languages

vv-monsalve:
This PR solves #5054 to fulfill the goal of #4722.


nathan-williams:
This change should be applied to https://github.com/googlefonts/lang if it hasn't already. From there, consider pulling in the change to the `google/fonts` repo via `git subtree pull ...`.

vv-monsalve:
We discussed this morning with @chrissimpkins about creating this PR here, in the same way other sample texts have been recently updated, until the migration or communication process between repos is defined.

chrissimpkins:
D/W Nate, this directory is a git subtree.  We will close this PR and make the changes on the googlefonts/lang repository instead.  Thanks Viv!

PR #5068, opened 2022-08-15
nable: v1.000 beta added

m4rc1e:
Taken from the upstream repo https://github.com/justvanrossum/nabla at commit https://github.com/justvanrossum/nabla/commit/3138ca588099ded0479cb6f9998b928c3ed3bb15

This is purely a test to see how the family handles in the dev server.


chrissimpkins:
Will require a couple of new axis registrations to onboard this family.

https://github.com/googlefonts/axisregistry/issues/58
https://github.com/googlefonts/axisregistry/issues/60

rsheeter:
Can you remove any axes that don't have an axis registry entry and then merge? - I want to see this on dev :)

PR #5091, opened 2022-08-18
Remove `lang` attribute from description html

nathan-williams:


rsheeter:
I think we should instead make lang a permissable attribute.

PR #5095, opened 2022-08-18
Add subset for math to Nabla and Kumbh Sans 

### LINKED ISSUE #5067: Kumbh Sans - GF specimen unable to show some glyphs which are available in the font
When using the font [specimen preview](https://fonts.google.com/specimen/Kumbh+Sans/glyphs), some mathematical glyphs like Integral, Summation, Infinity, Radical, etc. are not shown. Instead, it shows a default font. Upon downloading the Kumbh Sans font files, I can see that the glyphs are available in the font files (both static and variable).

**To Reproduce**
Click on [this link](https://fonts.google.com/specimen/Kumbh+Sans/glyphs). Next, navigate to the "Math Operators" and hover on the Integral, Summation, Infinity symbols. The preview will be from a default serif font. If you try pasting these glyphs inside the Specimen preview, a notdef symbol is shown.

**Expected behavior**
Since the glyphs are available in the font, the preview shall display original glyph and not a fallback or missing ".notdef".

**Screenshots**
screenshot-fonts google com-2022 08 15-14_45_09

**Additional context**
I am checking on Windows 10, Chrome 104.0.5.

davelab6:
Fixes missing glyphs in the GF catalog specimen page.

Fixes #5067

### DIFF
--- a/ofl/nabla/METADATA.pb
+++ b/ofl/nabla/METADATA.pb
@@ -14,6 +14,7 @@ fonts {
 }
 subsets: "latin"
 subsets: "latin-ext"
+subsets: "math"
 subsets: "menu"
 
 axes {
--- a/ofl/nabla/METADATA.pb
+++ b/ofl/nabla/METADATA.pb
@@ -15,6 +15,7 @@ fonts {
 subsets: "latin"
 subsets: "latin-ext"
 subsets: "math"
+subsets: "vietnamese"
 subsets: "menu"
 
 axes {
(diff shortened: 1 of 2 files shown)


ofl/nabla/METADATA.pb
```diff
@@ -14,6 +14,7 @@ fonts {
 }
 subsets: "latin"
 subsets: "latin-ext"
+subsets: "math"
```
chrissimpkins, line 17: I believe that you mentioned that we need Vietnamese to display the `Ẽ` and `ẽ`?

chrissimpkins: added in https://github.com/google/fonts/pull/5095/commits/01494285c496089345f8ae1795634f7a68bb6cda

chrissimpkins (approved):

vv-monsalve:
Only Kumbh Sans should be added to prod now

PR #5105, opened 2022-08-20
Update lang vi_Latn.textproto

chrissimpkins:
git subtree pull of changes in https://github.com/googlefonts/lang/pull/14


vv-monsalve (approved):

RosaWagner:
@vv-monsalve it hit prod and the good news is, it seems that only one passed through:

Capture d’écran 2022-09-29 à 11 20 24

cf https://fonts.google.com/specimen/Passions+Conflict/tester?subset=vietnamese

vv-monsalve:
> t seems that only one passed through

Do you mean that only one is missing? From the original [PR](https://github.com/google/fonts/pull/5005#issuecomment-1200112767) and inspecting [Pinyon Script](https://fonts.google.com/specimen/Pinyon+Script?query=Pin) in prod, it seems this is the only one I missed in the final [PR](https://github.com/googlefonts/lang/pull/14).

Feel free to make a new PR with the adjustment.

PR #5106, opened 2022-08-22
Noto Looped Thai: Version 1.001; ttfautohint (v1.8.4.7-5d5b) added

simoncozens:
86d1ebe: [gftools-packager] Noto Looped Thai: Version 1.001; ttfautohint (v1.8.4.7-5d5b) added

* Noto Looped Thai Version 1.001; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/notofonts/thai.git at commit https://github.com/notofonts/thai/commit/.


simoncozens:
Argh. This is already on GF as Noto Sans Thai Looped. Now I need to think about whether or not we need to rename the font upstream. I'd rather not have to hot fix it the naming for GF.

PR #5113, opened 2022-08-23
Noto Sans Bassa Vah: Version 2.002 added

simoncozens:
695e281: [gftools-packager] Noto Sans Bassa Vah: Version 2.002 added

* Noto Sans Bassa Vah Version 2.002 taken from the upstream repo https://github.com/notofonts/bassa-vah.git at commit https://github.com/notofonts/bassa-vah/commit/.


vv-monsalve:
@simoncozens was this shift part of the fixes?

simoncozens:
The shift is an improvement but it is is still wrong. It should look like this:
Screenshot 2022-09-24 at 11 05 03

(and it does locally, so...?)

vv-monsalve:
> (and it does locally, so...?)

Merging this here

PR #5115, opened 2022-08-23
Noto Sans Bamum: Version 2.001 added

simoncozens:
23a5f23: [gftools-packager] Noto Sans Bamum: Version 2.001 added

* Noto Sans Bamum Version 2.001 taken from the upstream repo https://github.com/notofonts/bamum.git at commit https://github.com/notofonts/bamum/commit/.


vv-monsalve:
Ok "bax_Bamu" doesn't have a sample text. So diffbrowsers images are displaying another font.

@simoncozens Could you add a sample text to the lang .textproto?

simoncozens:
Trying to get hold of one.

simoncozens:
It turns out there are currently very few Bamum texts encoded in Unicode. We would actually need to commision one! I'll work on it but I don't think we should hold up the PR for this; please ignore the text and just look at glyphs.

RosaWagner:
I block this cause doesn't display any text in sandbox.

Capture d’écran 2022-10-14 à 12 35 00

simoncozens:
OK, I think I understand what's going on with this *and* with Noto Music: if we don't have any sample text in the languages repository then we need to have sample text in the METADATA.pb or we get nothing at all.

RosaWagner:
@simoncozens Seeing interpolation issues in sandbox now that the sample text has been added:


https://user-images.githubusercontent.com/12222436/198000508-612b00c2-a2fd-474f-8009-163aad0888cb.mov

simoncozens:
Thought: We should totally implement fontTools.varLib.interpolatable as a fontbakery check.

simoncozens:
v2.002 is on its way.

PR #5117, opened 2022-08-23
Nabla: Version 1.000 added

m4rc1e:
5cc4473: [gftools-packager] Nabla: Version 1.000 added

* Nabla Version 1.000 taken from the upstream repo https://github.com/justvanrossum/nabla at commit https://github.com/justvanrossum/nabla/commit/.


rsheeter:
Merging because I want to play with it on dev

PR #5122, opened 2022-08-24
Noto Sans Cham: Version 2.002 added

simoncozens:
66ae21b: [gftools-packager] Noto Sans Cham: Version 2.002 added

* Noto Sans Cham Version 2.002 taken from the upstream repo https://github.com/notofonts/cham.git at commit https://github.com/notofonts/cham/commit/.


simoncozens:
Article missing

vv-monsalve:
@m4rc1e QA images are only displaying some punctuation marks. The Diffenator results are showing some Cham glyphs though, so I think it's only diffbrowsers issue. I've re-run the QA process, but the resulting images are the same.

Desktop_Windows_10_edge_18 0_

m4rc1e:
Looks like we don't have sample text for Cham.

simoncozens:
Shall I PR something to gftools which uses the gflanguages specimens?

simoncozens:
Wait, I already did!

```
if not words:
        languages = LoadLanguages()
        for file, proto in languages.items():
            if hasattr(proto, "sample_text"):
                for key, text in proto.sample_text.ListFields():
                    _add_words(words, text, seen_chars)
```

Perhaps that "if not words" is too strict.

simoncozens:
Actually that's weird. This should be a full build (including Latin).

m4rc1e:
gen-html sample text now present

Screenshot 2022-09-02 at 12 24 58

cc @vv-monsalve  and thanks to Simon for the gftools fix.

vv-monsalve:
Indeed, thanks Simon.

Now… the punctuation marks are shifting up in Edge, but I guess it could the same legacy browser issue.
Other than that LGTM.

Desktop_Windows_10_edge_18 0_

PR #5147, opened 2022-08-29
Fix to_sandbox.txt axis registry path.

garretrieger:


to_sandbox.txt
```diff
@@ -15,7 +15,7 @@ ofl/notosansthailooped # https://github.com/google/fonts/pull/5111
 lang/languages/vi_Latn.textproto # https://github.com/google/fonts/pull/5105
 
 # API Stuff
-axisregistry/Lib/axisregistry/data/element_grid.textproto # https://github.com/google/fonts/pull/5094
+axisregistry/element_grid.textproto # https://github.com/google/fonts/pull/5094
```
m4rc1e, line 18: @garretrieger would you like us to apply this path transform for axis registry files going forward? if so, I'll update the ci and our tools to build these lists. We do a similar transform for the lang dir.

garretrieger: Yes please, this path is what our tools expect.

garretrieger:
superseded by #5156

to_sandbox.txt
chrissimpkins: It looks like the CI tests need to be updated in this case.  They fail the path used in #5156

https://github.com/google/fonts/runs/8103625159?check_suite_focus=true

chrissimpkins: https://github.com/googlefonts/gftools/issues/603

PR #5149, opened 2022-08-30
add missing fallback_only to axes metadata

chuckxiong0211:


RosaWagner:
@vv-monsalve Do you know what we shall do with this PR? Prod?

RosaWagner:
@chuckxiong0211 Is the content of this PR went into sandbox and then to prod?

vv-monsalve:
> @vv-monsalve Do you know what we shall do with this PR? Prod?

I guess we need a way to confirm it has reached sandbox and then move it to production. Probably the eng team could tell us a way to verify this for further cases.

PR #5154, opened 2022-08-30
Run maximum_color on ReemKufiInk-Regular.ttf to add SVG table

anthrotype:
the font has COLRv1 table, so we want to tack on it an OT-SVG table for Safari/FF


anthrotype:
I run the latest `maximum_color` tool from https://github.com/googlefonts/nanoemoji/releases/tag/v0.14.1

PR #5166, opened 2022-09-01
Noto Sans NKo: Version 2.003; ttfautohint (v1.8.4.7-5d5b) added

simoncozens:
7fec968: [gftools-packager] Noto Sans NKo: Version 2.003; ttfautohint (v1.8.4.7-5d5b) added

* Noto Sans NKo Version 2.003; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/notofonts/nko.git at commit https://github.com/notofonts/nko/commit/.


simoncozens:
Some interesting fails here, probably due to the expected name "N'Ko".

ofl/notosansnko/METADATA.pb
```diff
@@ -1,19 +1,25 @@
-name: "Noto Sans N Ko"
+name: "Noto Sans NKo"
```
vv-monsalve, line 1: It should match the previous font. This would count as a CamelCase name, which is not allowed on a regular basis. Less of all here, given the previous name was `N Ko`

vv-monsalve (changes requested):

vv-monsalve:
Mh, new Fails received. It seems like the previous _name_ field in metadata `Noto Sans N Ko`, deferred from the font nameID 1, that was (is) `Noto Sans NKo`.

That camel case name is not included in our [list of exceptions](https://github.com/googlefonts/fontbakery/blob/b4bb6a063324a40c463a9696abd6662ad263510b/Lib/fontbakery/profiles/googlefonts_conditions.py#L402-L416). So I don't know how it was published. Perhaps the check was ignored.

So, I think the name you had before in the metadata is better (since it matches nameID 1) but that would mean changing that Metadata field, and IIUIC that is not possible. 

@RosaWagner I think you had a similar question in a recent project, but I can't remember now which one. Could you confirm if the `name` field in the METADATA.pb can change?

RosaWagner:
@vv-monsalve indeed we have the same issue with `Noto Sans Symbols 2` and we had it with `Cantora One`. Decision with Cantora was to break the font name, not the API name, so it only breaks office documents (if people download the update) and not served websites. I would be in favour of doing that now also for Noto since it is still not in Workspace and since Nko+Symbols2 are super in use still. But I would feel safer if @davelab6 could confirm all those cases before moving forward with a fix.

RosaWagner:
We have 3 options:
1. breaking API (adding a space in metadata.pb font name)
2. breaking name table (removing a space in font name tables)
3. Live with that exception (not fixing)

Since 1 and 2 doesn't seem appropriate for this family we could go for option 3. Will ask eng team if it's okay.

davelab6:
Here the API name was wrong. Nko is without space in ascii

simoncozens:
OK, I've reverted the space, and now I don't see a problem. It's "NKo" in the font and in the metadata, just like it was before. We just need a camelcase exception.

RosaWagner:
@simoncozens I am not sure that I can merge the replacement before we de-listed the font, it may create a conflict or something in dev-sandbox. Can you make a PR that renames the folder of the one we need to de-list (like appending -bak or something), and then this PR should be about "adding" a new family. When the new version is published, we would de-list the former and remove the related directory.

PR #5167, opened 2022-09-01
Reem Kufi Ink: Version 1.004 added

m4rc1e:
6b53d0e: [gftools-packager] Reem Kufi Ink: Version 1.004 added

* Reem Kufi Ink Version 1.004 taken from the upstream repo https://github.com/aliftype/reem-kufi at commit https://github.com/aliftype/reem-kufi/commit/.


anthrotype:
@m4rc1e which version of maximum_color did you use? Since Reem Kufi Ink doesn't have a colored .notdef glyph, it shouldn't be affected by the last couple of bugfixes on nanoemoji, but still worth checking.

m4rc1e:
nanoemoji==0.14.2

anthrotype:
that's good enough. I suppose re-running it with v0.14.3 makes no difference (beside maybe timestamp and checksums)?

I suggest next time you write the version number in the commit message, it might be handy for debugging.

PR #5171, opened 2022-09-01
Re-run maximum_color on BungeeSpice-Regular.ttf to fix issue with colored .notdef glyph being reshuffled

anthrotype:
There was an issue in the maximum_color tool that affected BungeeSpice-Regular.ttf, see https://github.com/googlefonts/nanoemoji/issues/427

Here I took the original BungeeSpice from upstream djrrb repository and run it through the latest maximum_color (v0.14.3).

https://github.com/djrrb/Bungee/blob/0678ec/fonts/Bungee_Color_Fonts/BungeeColor-Regular_colr_Windows.ttf-colrv1.ttf

The font has the (colored) .notdef glyph at index 0, and space character is again mapped to 0x20 in cmap as it should be. The space was currently un-cmapped (for it had been moved to index 0) and thus as-if missing.


m4rc1e:
Are we going to release Bungee Spice to fonts.google.com? I thought it was a Rod experiment?

PR #5178, opened 2022-09-01
Update Cairo Play layer coloring

rsheeter:
Quick and dirty fix:

```python

from fontTools import ttLib

_FOREGROUND_COLOR_ID = 0xFFFF

font_path = "ofl/cairoplay/CairoPlay[slnt,wght].ttf"
font = ttLib.TTFont(font_path)
cpal = font["CPAL"]
colr = font["COLR"]

assert len(cpal.palettes) == 1
assert str(cpal.palettes[0][0]) == "#000000FF"
color_id_black = 0

assert colr.version == 0
for gid, layers in colr.ColorLayers.items():
    for layer in layers:
        if layer.colorID != color_id_black:
            continue
        layer.colorID = _FOREGROUND_COLOR_ID

font.save(font_path)
```


vv-monsalve:
Is this PR meant to reach Sandbox?

rsheeter:
yes please

davelab6:
@Gue3bara has promised to update his upstream repo to match this in source and his next binary release this week

Gue3bara:
@davelab6 @chrissimpkins 

I am committed to working on this but there is an issue I am unable to tackle to be able to apply the needed changes. 

When @yanone worked on reviewing the font he moved the source to Glyphs 3 and the source file is now only workable in Glyphs 3 which is causing couple issues. 

- Fontmake does not support Glyphs 3 RTL kerning
- [and also you need to convert the old kerning to Glyphs 3 separate RTL and LTR kerning](https://github.com/googlefonts/glyphsLib/pull/778)

the issue is already [reported on Glyphs forum](https://www.google.com/search?q=glyphs%203%20rtl%20kerning&ie=utf-8&oe=utf-8&client=firefox-b-m) with no solid solution to it so far

I still believe solving this issue should be done by reverting to the original version but that will cancel the work @yanone and @m4rc1e did I am consulting with @khaledhosny if there is any other solution to it but we don't have an alternative proposal to solve this incompetence issue.

yanone:
I haven't followed this conversation in detail, but supporting RTL kerning in Glyphs 3 sources has been a priority of mine for the better part of this year, as my own upcoming commercial font will also rely on it. A glyphsLib patch is underway, and as I've just now returned to work  after my summer break, I'll return to this reinvigorated.

chrissimpkins:
@Gue3bara Is kerning not working in the last Cairo Play build (with always black foreground color) that you pushed with Marc?

Gue3bara:
@chrissimpkins it works differently than its behavior in Glyphs but is still acceptable, but as far as I can understand kerning can not be amended now in glyphs 3 nor it would work if I export font file using Glyphs3

anyways the issue is now solved and I just pushed a new version of the source file with the color issue solved

chrissimpkins:
> @chrissimpkins it works differently than its behavior in Glyphs but is still acceptable, but as far as I can understand kerning can not be amended now in glyphs 3 nor it would work if I export font file using Glyphs3
> 
> 
> 
> anyways the issue is now solved and I just pushed a new version of the source file with the color issue solved 

@m4rc1e @rsheeter

chrissimpkins:
Thank you very much Gaber.  I really appreciate it!

rsheeter:
Ty @Gue3bara

PR #5202, opened 2022-09-06
Noto Sans Lao Looped: Version 1.001 added

simoncozens:
83c64bc: [gftools-packager] Noto Sans Lao Looped: Version 1.001 added

* Noto Sans Lao Looped Version 1.001 taken from the upstream repo https://github.com/notofonts/lao.git at commit https://github.com/notofonts/lao/commit/.


simoncozens:
Yes, as this is a new font. @twardoch

vv-monsalve:
> Yes, as this is a new font.

Yeah, that was my suspect. I'll add the "Needs Description" label to all new fonts and tag @twardoch in that case.

RosaWagner:
@simoncozens Noto fonts should have a languages in metadata.pb, right?

simoncozens:
Yes - not sure why the packager missed them here.

simoncozens:
OK, I see why this is happening. gflanguages bug - `exemplar_chars.base` for Lao is full of mark glyphs.

PR #5209, opened 2022-09-06
Noto Sans Meetei Mayek: Version 2.002 added

simoncozens:
3cdbdab: [gftools-packager] Noto Sans Meetei Mayek: Version 2.002 added

* Noto Sans Meetei Mayek Version 2.002 taken from the upstream repo https://github.com/notofonts/meetei-mayek.git at commit https://github.com/notofonts/meetei-mayek/commit/.


vv-monsalve:
LGTM

PR #5211, opened 2022-09-07
replace U+2010 HYPHEN with U+002D Hyphen-Minus in sample texts

chuckxiong0211:
git subtree pull of changes in https://github.com/googlefonts/lang/pull/15


vv-monsalve:
@chuckxiong0211 this PR includes some commits already added on #5105

chuckxiong0211:
@vv-monsalve yes and that's because we do not squash at git subtree pull time. I checked the diff and it only includes what we need so it should be okay.

Issue #5070, opened 2022-08-15
ofl/cantarell: inconsistencies in metadata/filenames


nathan-williams:
**Describe the bug**

```
ofl/cantarell/Cantarell-BoldOblique.ttf
ERROR: OS2 weight class 500 doesn\'t match with the given weight 400 in METADATA.pb
Bad filename \'Cantarell-BoldOblique.ttf\', expected \'Cantarell-BoldItalic.ttf\'
ofl/cantarell/Cantarell-Oblique.ttf
ERROR: OS2 weight class 500 doesn\'t match with the given weight 400 in METADATA.pb
```

rsheeter:
I'm surprised we don't check for this in font bakery. Maybe we could, summon @felipesanches :)

RosaWagner:
These are the exact same font files as before, except for the license name ID that was clarified to fix #4551. This family has a lot of issues but we can’t work anymore from sources: so Fontbakery is probably checking for all these problems, but it was really not the goal here to fix them.

vv-monsalve:
@davelab6 is going to inspect at this

davelab6:
>     ofl/cantarell/Cantarell-BoldOblique.ttf
>     ERROR: OS2 weight class 500 doesn\'t match with the given weight 400 in METADATA.pb

This doesn't quite make sense to me, as the given weight for that filename is 700, not 500:

https://github.com/google/fonts/blob/d4f46c6343831649b5f094c9f1a7eace774d9493/ofl/cantarell/METADATA.pb#L36-L37

>     Bad filename \'Cantarell-BoldOblique.ttf\', expected \'Cantarell-BoldItalic.ttf\'

That's an easy fix :) 

>     ofl/cantarell/Cantarell-Oblique.ttf
>     ERROR: OS2 weight class 500 doesn\'t match with the given weight 400 in METADATA.pb

I think a ttx hotfix for this is best, since the later work on the family in the GNOME community radically changed the design, and was eventually onboarded as https://fonts.google.com/specimen/Cambay , and the original design hasn't been touched by me since 2010 and the FontForge sources are well out of date.

davelab6:
The Regular and Italic files have an OS/2 weight class of 

    $ gftools-list-weightclass.py *ttf

| filename                 |   usWeightClass |
|:-------------------------|----------------:|
| Cantarell-Regular.ttf    |             500 |
| Cantarell-Oblique.ttf     |             500 |
| Cantarell-Bold.ttf       |             700 |
[...]

PR #5213, opened 2022-09-08
Cantarell: Version 1.004 added

davelab6:
Hotfixes for #5070

* filenames
* name tables
* vertical metrics
* weightClass


RosaWagner:
Bit concerned with the change of name: indeed the file name of was "Italic" in the live server, but the actual full name was "Oblique". Won't it create reflow?

davelab6:
> Bit concerned with the change of name: indeed the file name of was "Italic" in the live server, but the actual full name was "Oblique". Won't it create reflow?

Usually document authoring tools have remapping features for 1:1 style name changes, so I think it's ok to fix. Let's add an explanation in the DESCRIPTION so users know what's going on

PR #5218, opened 2022-09-11
Noto Music: Version 2.002; ttfautohint (v1.8.4.7-5d5b) added

simoncozens:
5646ff6: [gftools-packager] Noto Music: Version 2.002; ttfautohint (v1.8.4.7-5d5b) added

* Noto Music Version 2.002; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/notofonts/music.git at commit https://github.com/notofonts/music/commit/.


simoncozens:
Thanks for this. I had been meaning to get to it. I don't think "sample text" makes much sense for music but it's good to have something, right?

PR #5222, opened 2022-09-12
Noto Sans New Tai Lue: Version 2.003 added

simoncozens:
8a50638: [gftools-packager] Noto Sans New Tai Lue: Version 2.003 added

* Noto Sans New Tai Lue Version 2.003 taken from the upstream repo https://github.com/notofonts/new-tai-lue.git at commit https://github.com/notofonts/new-tai-lue/commit/.


RosaWagner:
@simoncozens 
This looks like a good re-design, but what do I know. The font is featured on few thousands websites only, so I wouldn't worry too much about the small line length increase.

glyphs_modified
Desktop_Windows_10_edge_18 0_

PR #5226, opened 2022-09-12
IBM Plex Sans JP: Version 1.001 added

yanone:
a08f44a: [gftools-packager] IBM Plex Sans JP: Version 1.001 added

* IBM Plex Sans JP Version 1.001 taken from the upstream repo https://github.com/googlefonts/plex at commit https://github.com/googlefonts/plex/commit/80fcb676509354caa75c6bfc7f9db598545d466f.


yanone:
Please note: The found FAILs are in line with the previously published other Plex fonts. We’ve agreed to adjust only the name table and the vertical metrics.

PR #5227, opened 2022-09-12
[gflanguages] U+0323 instead of U+329 in yo_Latn

moyogo:
- Use U+0323 instead of U+329 in Yoruba (yo_Latn) sample text.
  While the vertical line below does occur in printed text as an alternative to the dot below,
  the Unicode characters used are typically the combining dot below or characters canonically
  composed with it.
  This replaces the combining vertical line below U+0323 with dot below U+0329 and normalizes (NFC) the text.
  The exemplar characters already use U+0329.


moyogo:
I opened https://github.com/unicode-org/udhr/pull/47 as well.

moyogo:
Good point.

PR #5234, opened 2022-09-14
Gulzar: primary script added

RosaWagner:
Added arabic as primary script cause Latin is displaying in prod

### DIFF
--- a/ofl/gulzar/METADATA.pb
+++ b/ofl/gulzar/METADATA.pb
@@ -16,6 +16,7 @@ subsets: "arabic"
 subsets: "latin"
 subsets: "latin-ext"
 subsets: "menu"
+primary_script: "Arab"
 source {
   repository_url: "https://github.com/simoncozens/Gulzar"
   commit: "7f736710573840523e68ea977843f7af7239632b"


m4rc1e:
I spoke to @nathan-williams yesterday and he raised an interesting implementation detail regarding the Google Fonts frontend. He said that the sample text on fonts.google.com can be influenced by the user's geo location e.g users in Oman will get Arabic sample text for the Cairo family whilst users in London will get Latin. Due to this, it may be wise to not overly use the primary script field. However, it's fine for this family.

simoncozens:
Bigger problem is that we want to show Arabic script but specifically Urdu language, not Arabic language. Perhaps we need `sample_text` metadata.

vv-monsalve:
> whilst users in London will get Latin

We talked about this with @RosaWagner, and we think it might be a good idea for non-complex scripts (Latin Greek Cyrilic) but for fonts intended for any complex script (Devanagari, Arabic, etc) they probably should display the primary script no matter the user's location.
That way it would be clear for users what's the font priority and would make more sense in cases like [Tiro Indigo](https://fonts.google.com/?query=Tiro) project, particularly since the Latin included in those fonts would be the Kernell glyphset.

PR #5242, opened 2022-09-14
Noto Sans Tai Tham: Version 2.002 added

simoncozens:
64bad9c: [gftools-packager] Noto Sans Tai Tham: Version 2.002 added

* Noto Sans Tai Tham Version 2.002 taken from the upstream repo https://github.com/notofonts/tai-tham.git at commit https://github.com/notofonts/tai-tham/commit/.


RosaWagner:
@simoncozens 
glyphs_missing

simoncozens:
The bottom line is a fix, https://github.com/notofonts/tai-tham/issues/3
The first line, it looks like the mai kang lai (kinzi mark) has been redrawn to look more like the Burmese kinzi. I don't know if this is right or not. It looks like David Williams did it.

simoncozens:
According to @ohbendy, this shape of Mai Kang Lai is used in two varieties of Tai Tham, but not in others. Please onboard it as it is for now; in the next release, we will determine the most appropriate form and add the other form as a stylistic set.

ohbendy:
> and add the other form as a stylistic set.

I would not recommend doing it that way, it'd be like putting Noto Nastaliq of Kufic as a stylistic set in Noto Sans Arabic (or Noto Looped Thai as a stylistic set in Noto Sans Thai). As you'll see in the report I sent you, all the letterforms are drawn in different styles in the different forms of Tham script, and many of the characters' shapes are very different.

simoncozens:
Understood; I will take a look through the report when I have a moment to catch my breath - my takeaway at the moment is that we are probably going to need to have a broader look at this font, possibly involving establishing better Unicode/USE/OT support. But for now let's onboard what we have.

ohbendy:
Indeed yes. Since there's not even an OpenType specification for what constitutes a valid syllable in Tham, and since no operating system can shape Tham text correctly, I don't think there's any possibility to create a functional Tham font.

PR #5244, opened 2022-09-15
Noto Serif NP Hmong: Version 1.001 added

simoncozens:
b07a2e5: [gftools-packager] Noto Serif NP Hmong: Version 1.001 added

* Noto Serif NP Hmong Version 1.001 taken from the upstream repo https://github.com/notofonts/nyiakeng-puachue-hmong.git at commit https://github.com/notofonts/nyiakeng-puachue-hmong/commit/.


simoncozens:
This font will have to be handled manually.

PR #5245, opened 2022-09-15
re-maximum_color Nabla to support OT-SVG with multi-palette CPAL

anthrotype:
This PR re-runs the maximum_color tool on Nabla v1.002 using the latest [nanoemoji v0.15.0](https://github.com/googlefonts/nanoemoji/releases/tag/v0.15.0 ), which adds support for converting a COLR/CPAL font with multiple color palettes (such as Nabla) to an OT-SVG table that contains color variables referencing those CPAL palettes.
Note that the font with combined COLR + SVG tables is now 45KB larger, all from the additional var(--colors) in the SVG table (the other tables didn't change), i.e. from 1,598,000 bytes to 1,643,616 bytes (less than 3% increase).
The additional size is worth it because it allows to select alternative font palettes or customize colors within those via CSS on the latest Safari on iOS 16 and macOS Ventura beta (below is a screenshot from the latter):

image


RosaWagner:
Shall this PR be handled by the classic server QA process (push lists) ?

anthrotype:
> Shall this PR be handled by the classic server QA process (push lists) ?

hi Rosalie, I'm not sure why you mean by that? /cc @rsheeter

RosaWagner:
I mean: shall I add it to the sandbox list, or is it something handled internally?

anthrotype:
hey Rosalie, I just chatted with Rod about this

> shall I add it to the sandbox list

yes please, thanks!

RosaWagner:
Okay 👍 so my problem is that the version number is the same as before I wouldn’t be able to know if this version got into sandbox. If you can bump it, it would be great, otherwise I’ll do it tomorrow before adding it to the list.

anthrotype:
you mean the version string as in nameID=5? This is the same Nabla font that was released last week (v1.002), I took the fonts from the same upstream zip file as listed in the METADATA.pb file: that is https://github.com/justvanrossum/nabla/releases/download/v1.002/nabla-fonts.zip

Nothing changed upstream, so I don't think we should modify the version string. I simply re-run the maximum_color tool, this time using an updated version of the tool. The result is the same as the current published fonts _except_ for the "SVG " table, which can now reference the color palettes in the CPAL (and thus allow Safari on iOS or macOS using the OT-SVG variant of Nable to select alternative color palettes or replace particular colors therein).
Instead of inlined SVG color attributes like fill="#FF0000", it now contains color variables such as `var(--color0, #FF0000)` which means "use the color at palette entry index 0 from CPAL table, otherwise if you don't support reading colors from CPAL, fall back to the default color listed after the comma".
To know if it's the updated font, you can check if the SVG table is larger (it's around 45KB bigger), or that its elements contain attributes starting with "var(--color"..

RosaWagner:
We are checking around 80 fonts between the sandbox and the live server, so I won’t check the table. Per requirement we have to increase the version number when there is even the slightest change to a font to know which version is where, it is not about the change itself.

It’s okay, I’ll do it tomorrow :)

rsheeter:
I wonder if that means we should use something other than the version from name. For example, the sequence of changes to size and checksum of the file could be taken as version (e.g. commit file to git, that's v1, update it, that's v2, etc)

chrissimpkins:
> Nothing changed upstream

I think perhaps it should so that we can reproduce the same file with the next push from Arthur and Just?  Can we script this on COLR projects that require it so that we don't need to remember to do it with the next onboarding push?

anthrotype:
Yes, I think that's what we want to do. It's a matter of running the latest maximum_color tool from nanoemoji project on a COLRv1 font, to obtain the hybrid COLR+SVG font ready for onboarding. Fontbakery has a check now, and I believe gftools has a fixer that runs maximum_color automatically on COLRv1 fonts

PR #5253, opened 2022-09-16
Noto Serif Tibetan: Version 2.103 added

simoncozens:
80d000d: [gftools-packager] Noto Serif Tibetan: Version 2.103 added

* Noto Serif Tibetan Version 2.103 taken from the upstream repo https://github.com/notofonts/tibetan.git at commit https://github.com/notofonts/tibetan/commit/.


RosaWagner:
Is this all good @simoncozens ? 

simoncozens:
I think it is. The font went back to the designer for another round of improvements, so I am definitely expecting some changes. I'm afraid I don't understand how this font works well enough to tell if it's correct. :-/

RosaWagner:
My guess would be that it looks good except for the glyphs with added diacritic in what looks like a random position:

Capture d’écran 2022-09-22 à 12 21 26

It would good to know if it's a bug from diffbrowser or real intention

simoncozens:
Well, here's the sequence on the left: 'ྚ༵' and here's what it looks like in Harfbuzz:

shape

So I am going to claim a diffenator bug.

PR #5257, opened 2022-09-19
Noto Serif NP Hmong: Version 1.001 added (manual)

### LINKED ISSUE #3671: Rename "Noto Serif Nyiakeng Puachue Hmong" to "Noto Serif NP Hmong"
At least in the API, but ultimately also in the distributables, see 
https://github.com/googlefonts/noto-fonts/issues/2122

simoncozens:
This is Noto Serif NP Hmong taken from [release v1.001](https://github.com/notofonts/nyiakeng-puachue-hmong/releases/tag/NotoSerifNPHmong-v1.001).

I don't know if this is going to work. It replaces the existing font "Noto Serif Nyiakeng Puachue Hmong" with another one called "Noto Serif NP Hmong", but uses the METADATA.pb `display_name` to keep things the same in the web API.

fixes #3671


RosaWagner:
Copy/pasting Chris' message here:
It is decided that we will move the fonts in the PR to a new ofl directory, change the family name, and delist the old family with the long name.

simoncozens:
> It is decided that we will move the fonts in the PR to a new ofl directory, change the family name, and delist the old family with the long name.

When you say "we"... :-) Is there something you need me to do here?

RosaWagner:
I am unsure of all the details. I think I am expecting a new ofl directory to not replace the old font or something? Like we should have two until the last one is published and then we can de-list and removed the former one. Correct me if I am wrong.

PR #5271, opened 2022-09-19
Noto Sans Khudawadi: Version 2.003; ttfautohint (v1.8.4.7-5d5b) added

simoncozens:
37b34c0: [gftools-packager] Noto Sans Khudawadi: Version 2.003; ttfautohint (v1.8.4.7-5d5b) added

* Noto Sans Khudawadi Version 2.003; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/notofonts/khudawadi.git at commit https://github.com/notofonts/khudawadi/commit/.


simoncozens:
Bad sample text issue. :-( googlefonts/lang#97 will fix.

RosaWagner:
Okay then moving to prod

simoncozens:
We need to make/push a new lang release first though.

RosaWagner:
It's okay, the font doesn't need to be re-pushed when the sample text is updated, the upgrade can be pushed and the fix on the sample text will follow the next week ;)

PR #5342, opened 2022-09-29
Noto Serif HK: Version 2.004-H1 added

simoncozens:
This replaces the binary of Noto Serif HK from one based on [`NotoSerifCJKhk-VF.ttf`](https://github.com/googlefonts/noto-cjk/blob/main/Serif/Variable/TTF/NotoSerifCJKhk-VF.ttf) to one based on [`Subset/NotoSerifHK-VF.ttf`](https://github.com/googlefonts/noto-cjk/blob/main/Serif/Variable/TTF/Subset/NotoSerifHK-VF.ttf).


RosaWagner:
follow up in #5535

PR #5351, opened 2022-09-30
Update lang subtree with yo_Latn fix

moyogo:
Update lang subtree to latest commit.


moyogo:
@chrissimpkins
There was a merge conflcit on the `Lib/gflanguages/data/languages/yo_Latn.textproto` after running the `git subtree pull` command on the lang subtree and several other commits are part of the PR. Is that normal?

RosaWagner:
Most of these commits have been pull here already #5105 #5211, but I don't know how the git subtree command works

chrissimpkins:
Yeah, unless we squash on the git subtree pull (Rod requested that we not do that) we get all of the commit history stacked in this repo because we squash at main branch merge time by default.  Despite all of those commits showing up, it looks like there is a single file change here.  Is this the file that you intended to edit?

I *think* that we can avoid this in the future by merging the subtree commits into google/fonts main branch with a merge commit, rather than using the default PR squash approach which alters subtree history as the changes are brought into main here.

m4rc1e:
Closing since #5451 is newer.

PR #5359, opened 2022-10-03
Fix Open Sans description

applecuckoo:
Steve isn't the type director of Ascender anymore. Feel free to redo this part in your own words if you want.


applecuckoo:
I feel that this probably doesn't deserve a pull request, opening an issue instead.

PR #5361, opened 2022-10-03
Foldit: Version 1.003 added

m4rc1e:
2820c7c: [gftools-packager] Foldit: Version 1.003 added

* Foldit Version 1.003 taken from the upstream repo https://github.com/SophiaDesign/Foldit at commit https://github.com/SophiaDesign/Foldit/commit/.


RosaWagner (approved):

m4rc1e:
Thanks. I'll upload the designers details after lunch in a new pr.

We also need to keep this PR blocked until the latest version of Chrome is out.

vv-monsalve (approved):

vv-monsalve:
> Thanks. I'll upload the designers details after lunch in a new pr.

I've sent an email to the designer so that she can provide an image for her profile. I'll create a PR with her information after that.

emmamarichal:
@vv-monsalve, sorry I hadn't seen your comment, and I sent a message on instagram to the author. She just added her avatar to the folder. 👌  I can do it if you want, unless you've already started.

RosaWagner:
@m4rc1e @vv-monsalve writing here what Rod reported in the chat; the font needs to go through the color script again (missing SVG table)

vv-monsalve:
> @vv-monsalve, sorry I hadn't seen your comment, and I sent a message on instagram to the author. She just added her avatar to the folder. 👌 I can do it if you want, unless you've already started.

Ah, ok, she didn't reply to my email so I haven't started it. If she has provided the image now, yes please you could add her bio ingo if that's ok for you. Thanks

vv-monsalve:
> @m4rc1e @vv-monsalve writing here what Rod reported in the chat; the font needs to go through the color script again (missing SVG table)

@m4rc1e Is that something we should do it independently (post-process) or is it already included in fontmake and then cutting a new release would be enough to cover it?

vv-monsalve:
Pending on support for variations of gradients 

- https://github.com/googlefonts/nanoemoji/issues/438#event-7583200454
- https://github.com/googlefonts/nanoemoji/issues/426

anthrotype:
just to clarify once again, "support for variations of gradients" does not mean that the SVG table converted from COLRv1 will contain variations (that's unsupported in the format), simply means maximum_color will henceforth ignore variation data in COLR table when generating the SVG table which will then only contain paths/gradients from the default VF instance.

anthrotype:
FYI https://github.com/googlefonts/nanoemoji/pull/439

vv-monsalve:
Thanks for the expanded clarification @anthrotype.

> simply means maximum_color will henceforth ignore variation data in COLR table when generating the SVG table which will then only contain paths/gradients from the default VF instance

Would it mean that the same gradients would be used for the wider-bolder instances? As dave said in one comment of the issues:

> is a new variable color font that makes extensive use of color variations (amusingly in a way to create the illusion that as the weight (and natural width) changes, the gradients look like they don't change, which requires them TO change

So, would that effect be lost in the resulting VF if the SVG table will contain only the gradients of the font origin?

> Thin weight will be displayed for those environments that only support OT-SVG and not COLRv1 (e.g. Safari/FireFox, macOS, iOS, Adobe apps etc.)

I don't think this is an issue.

anthrotype:
> Would it mean that the same gradients would be used for the wider-bolder instances?

yes - not just the gradient geometries, the shapes themselves will not vary, SVG simply can't encode OT variations. The rest of the font may be variable, and that's actually a problem because if you select anything other than the Thin instance it will look wrong (outlines of the thin, too wide metrics for the selected non-Thin instance). see https://github.com/googlefonts/nanoemoji/pull/439#issuecomment-1278993465

SophiaDesign:
> If you wanted a more "regular" weight to be shown by default you'd have to change the default master to be the Regular (if there's one, haven't looked at your source). If you are fine with it, never mind ignore my comment.

There is no Regular master in the source file. It was removed to minimise the file size. Can be added easily if preferred?

vv-monsalve:
> There is no Regular master in the source file. It was removed to minimise the file size. Can be added easily if preferred?

I think this isn't needed because of the reasons provided on the related issues and [Marc's comment](https://github.com/googlefonts/nanoemoji/pull/439#issuecomment-1279048438) in the above issue.

To my understanding, when he says that is _better to leave the font as is_ this implies:
- Not to run the `maximum_color` (so we wouldn't have the hybrid COLRv1+SVG when the COLR table contains variations since a B/N VF is preferable rather than having only the SVG of the font origin)
- We release it with a note to users in the description file
- Should we still wait for the latest version of Chrome?
- Would it be something else be necessary?

cc @m4rc1e, @chrissimpkins, @davelab6

chrissimpkins:
This family remains blocked for technical reasons and should not be moved beyond dev until further notice.

PR #5374, opened 2022-10-06
Marhey: Version 1.000 added

yanone:
31a685a: [gftools-packager] Marhey: Version 1.000 added

* Marhey Version 1.000 taken from the upstream repo https://github.com/namelatype/Marhey at commit https://github.com/namelatype/Marhey/commit/e5686e381e7ca5b941a5dfd3e91fe92ab7a0eb60.


yanone:
I don't get the STAT table FAIL when I run the latest fontbakery on my computer.
Otherwise, the font never had a Medium instance. IMO these checks are not supposed to FAIL because of that.

RosaWagner:
If the font is a VF it must have a medium instance. The API will generate it automatically, so it has to be in the FVAR table as well. You can either add it where it would be automatically calculated or ask the designer to re-balance the instances weight distribution so it make sense for a medium weight to be there. I know it is annoying, a lot of designers complain, I can link several issues where it is discussed if needed.

yanone:
Okay, please link a few

yanone:
Updated

## Marhey: Version 1.000 added

---
 5374a28: [gftools-packager] Marhey: Version 1.000 added

* Marhey Version 1.000 taken from the upstream repo https://github.com/namelatype/Marhey at commit https://github.com/namelatype/Marhey/commit/535eeb5a3f4dee825fdaf13d84075b212d969f8d.

PR #5377, opened 2022-10-06
Noto Serif Dogra: Version 1.006; ttfautohint (v1.8.4.7-5d5b) added

simoncozens:
3d26193: [gftools-packager] Noto Serif Dogra: Version 1.006; ttfautohint (v1.8.4.7-5d5b) added

* Noto Serif Dogra Version 1.006; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/notofonts/dogra.git at commit https://github.com/notofonts/dogra/commit/.


simoncozens:
Oh heck. I think I understand this. The font was previously built assuming that Dogra goes through the standard shaper, not the USE. But these days Harfbuzz pushes it through the USE, so I rebuilt the font to use the USE features appropriately. So now it works on Harfbuzz, but not on older shapers.

This will be interesting to fix.

simoncozens:
This is for next year anyway.

RosaWagner:
@simoncozens we would need to merge that PR soon so we can move forward with the language update

simoncozens:
> So now it works on Harfbuzz, but not on older shapers.

It now also works with latest Safari. I'm happy with it. If Edge is broken, this is an Edge bug, not a font bug.

RosaWagner (approved):

PR #5384, opened 2022-10-10
Reem Kufi Fun: Version 1.005 added

yanone:
9239ee3: [gftools-packager] Reem Kufi Fun: Version 1.005 added

* Reem Kufi Fun Version 1.005 taken from the upstream repo https://github.com/aliftype/reem-kufi at commit https://github.com/aliftype/reem-kufi/commit/.


RosaWagner:
Since Rod reported a missing SVG table in Foldit, I checked this one for it manually, and there is none.

m4rc1e:
The FB check which tests this still needs merging https://github.com/googlefonts/fontbakery/pull/3889

yanone:
I'll let Khaled know so he can integrate maximum_color it into his repo

yanone:
Khaled says the expected "Regular" in the name checks are nonsense, as this is a Regular font. Will this be able to pass?

m4rc1e:
It's fine since previous version was like this. I'm going to just raise a WARN if Regular fonts are missing Regular of the fullname and ps name in the future.

yanone:
Another quite important question: Khaled just verified why I ask a SVG table of him given that this is a COLRv0 font. I think he's right. The fontbakery check is explicit about that and actually complains when a COLRv0 font has a SVG table, while it demands one for a COLRv1 font.

m4rc1e:
Correct, COLR v0 doesn't need it. However, the FB check for this has only just been merged https://github.com/googlefonts/fontbakery/pull/3889 so we'll only be able to test this correctly once the next release is out.

yanone:
So then this one is good to go

PR #5393, opened 2022-10-11
Noto Sans Vithkuqi: Version 1.001 added

simoncozens:
f88ffc6: [gftools-packager] Noto Sans Vithkuqi: Version 1.001 added

* Noto Sans Vithkuqi Version 1.001 taken from the upstream repo https://github.com/notofonts/vithkuqi.git at commit https://github.com/notofonts/vithkuqi/commit/.


simoncozens:
Language data added, dependent on merge of https://github.com/googlefonts/lang/pull/88

simoncozens:
We have Vithkuqi sample text in lang as of v0.5.1, and also Vithkuqi glyphset is merged, so this should be good to advance.

PR #5400, opened 2022-10-12
Add primary scripts

simoncozens:
This is for discussion. It applies the script `gftools-set-primary-script` defined [here](https://github.com/googlefonts/gftools/pull/620) to the catalogue, setting the primary script of all non-Latin fonts in the catalogue.


ofl/lohitbengali/METADATA.pb
```diff
@@ -13,3 +13,4 @@ fonts {
   copyright: "Copyright 2011 Lohit Fonts Project contributors (psatpute@redhat.com). <http://fedorahosted.org/lohit> Licensed under the SIL Open Font License 1.1 (see file OFL.txt)"
 }
 subsets: "bengali"
+primary_script: "Beng"
```
RosaWagner, line 16: The three LOHIT families are not in the live catalog, maybe the script should only use the family list from the live server.

simoncozens: Is it a problem to add this information to non-live fonts?

RosaWagner: We generate the lists using all modified directories path so yes cause they might be pushed by mistake.

RosaWagner:
Due to conflicts and stuff to fix in the script, I close this PR. In term of method, would be good to do it per group of scripts to make it easier to check in sandbox.

ofl/lohitbengali/METADATA.pb
davelab6: I forgotten why lohit not added, is it no Latin?

PR #5410, opened 2022-10-14
primary script on selected families

RosaWagner:
- SIL
- IBM Plex
- Tiro
- Baloo
- Zen
- Morisawa


chrissimpkins (commented):
Thanks Rosalie.  Are you confident that the automated approach properly labeled the primary script in https://github.com/google/fonts/pull/5410/commits/d69543889fef2396a257fd7963356fd844dd9875?

RosaWagner:
I am confident it was correct, but I reported in gftools the problem that the script was missing families which have bigger Latin glyph set. Imo we can assume that if the script name is in the family name, then that script it takes precedence over Latin.

Merging this for the next sandbox push.

RosaWagner:
I am inspecting the sandbox now. I do see some problems with CJK fonts, but I don't know if it comes from the dev-sandbox or font the font's metadata.pb. I'll give more time to propagate and will review again later.

Works awesomely so far with Indic and Arabic scripts:
Capture d’écran 2022-10-20 à 11 06 01
Capture d’écran 2022-10-20 à 11 12 24

Problem with CJK

Capture d’écran 2022-10-20 à 11 12 47
Capture d’écran 2022-10-20 à 11 11 27


(dev-sandbox doesn't display fonts that were unchanged for a long time, so the empty blocks are normal)

chrissimpkins:
Quick spot check of dev IBM Plex Sans JP suggests that the Latin is now gone.

Issue #5421, opened 2022-10-18
Q3 Knowledge issues


rsheeter:
Some of the Q3 Knowledge update is bad placeholders, including _at least_ the following:

```text
INVALID  cc-by-sa/knowledge/glossary/terms/cjk/term.textproto "INSERT_RELATED_LESSON" => /lesson/INSERT_RELATED_LESSON/content.md
INVALID  cc-by-sa/knowledge/glossary/terms/alphabet/term.textproto "INSERT_RELATED_LESSON" => /lesson/INSERT_RELATED_LESSON/content.md
INVALID  cc-by-sa/knowledge/glossary/terms/designspace/term.textproto "INSERT_RELATED_LESSON" => /lesson/INSERT_RELATED_LESSON/content.md
INVALID  cc-by-sa/knowledge/glossary/terms/tofu/term.textproto "INSERT_RELATED_LESSON" => /lesson/INSERT_RELATED_LESSON/content.md
INVALID  cc-by-sa/knowledge/glossary/terms/rag/term.textproto "INSERT_RELATED_LESSON" => /lesson/INSERT_RELATED_LESSON/content.md
INVALID  cc-by-sa/knowledge/glossary/terms/letters/term.textproto "INSERT_RELATED_LESSON" => /lesson/INSERT_RELATED_LESSON/content.md
INVALID  cc-by-sa/knowledge/glossary/terms/reflow/term.textproto "INSERT_RELATED_LESSON" => /lesson/INSERT_RELATED_LESSON/content.md
INVALID  cc-by-sa/knowledge/glossary/terms/emoji/term.textproto "INSERT_RELATED_LESSON" => /lesson/INSERT_RELATED_LESSON/content.md
INVALID  cc-by-sa/knowledge/modules/history_of_type/lessons/what_have_the_romans_ever_done_for_us/lesson.textproto "type_design" => cc-by-sa/knowledge/glossary/terms/type_design/content.md
INVALID  cc-by-sa/knowledge/modules/history_of_type/lessons/what_have_the_romans_ever_done_for_us/lesson.textproto "type_history" => cc-by-sa/knowledge/topics/type_history/topic.textproto
INVALID  cc-by-sa/knowledge/modules/history_of_type/lessons/the_invention_of_type/lesson.textproto "type_design" => cc-by-sa/knowledge/glossary/terms/type_design/content.md
[...]

PR #5422, opened 2022-10-18
Add additional Knowledge checks

rsheeter:
https://github.com/google/fonts/issues/5421 has refs to a couple of files with breakages that I fixed based on new checks.

Requires https://github.com/googlefonts/gftools/pull/621


nathan-williams (approved):
Thanks a lot for this, Rod!

nathan-williams (approved):

PR #5426, opened 2022-10-18
Update to_sandbox.txt: Hold Climate Crisis and Handjet sandbox pushes

chrissimpkins:
D/W Rod.  The new axis registrations that are required to support these projects must be pushed to production before the font asset sandbox push

cc @RosaWagner @rsheeter @nathan-williams

### DIFF
--- a/to_sandbox.txt
+++ b/to_sandbox.txt
@@ -1,6 +1,4 @@
 # New
-ofl/climatecrisis # https://github.com/google/fonts/pull/5350
-ofl/handjet # https://github.com/google/fonts/pull/5376
 ofl/marhey # https://github.com/google/fonts/pull/5374
 ofl/notosansmendekikakui # https://github.com/google/fonts/pull/5143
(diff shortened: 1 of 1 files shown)


to_sandbox.txt
```diff
@@ -24,3 +22,9 @@ catalog/designers/sophiatai # https://github.com/google/fonts/pull/5408
 
 # Metadata / Description / License
 ofl/notosansbamum # https://github.com/google/fonts/pull/5412
+
+# Hold List
```
chrissimpkins, line 26: @RosaWagner I created a new hold list section at the bottom of the list so that we don't lose track of these files.  They will not be pushed to sandbox this week.  We should add YEAR, ELGR, and ELSH axes to the to_production.txt list, land those axis registrations in production, then confirm with the on-call eng that the pipeline is ready for the **sandbox** font asset push before we add the font families to the to_sandbox list.  This is a change from our original plan.  One week post-axis registration was not enough time.

PR #5442, opened 2022-10-20
Testing Hani, Hira and Kana primary scripts

RosaWagner:
Testing some primary_script on:
 
- Simplified chinese : ZCOOL
- Katakana only : Tsukimi Rounded (currently showing tofu in sandbox)
- Hiragana only : Yuji Hentaigana (currently showing tofu in sandbox)

Wether this test is successful or not, I would revert it once checked in dev-sandbox.


RosaWagner:
These 3 tests failed, they all continue to show latin instead of the primary script.

PR #5452, opened 2022-10-21
Asap: Version 3.001 added

emmamarichal:
a8865d3: [gftools-packager] Asap: Version 3.001 added

* Asap Version 3.001 taken from the upstream repo https://github.com/Omnibus-Type/Asap at commit https://github.com/Omnibus-Type/Asap/commit/927ab390d4ece9eaa70a3b16a6124baa9192e34c.

 c0a7145: Description updated according the update


RosaWagner:
@emmamarichal are the static fonts hinted manually ?

emmamarichal:
@RosaWagner No they aren't, there is problem somewhere?

RosaWagner:
Then you can remove the static folder and all the statics from the upstream.yaml 
(I feel your pain)

emmamarichal:
Haha don't worry, it didn't take me long thanks to find/replace. I had left it because in the API version there is a statics folder, so I didn't dare to remove them.

RosaWagner:
@emmamarichal  superiors missing (check names)

glyphs_missing

Probably same in #5492

emmamarichal:
@RosaWagner Ah yes, there were named uni2080 uni2081 uni2082 uni2083 uni2084 uni2085 uni2086 uni2087 uni2088 uni2089 etc.
It concerns .sups and .subs, I'm going to correct that, thanks!

emmamarichal:
@RosaWagner it concerns .numr, but they look exactly the same in the new and old file.
I thought it was because of component, made from a glyphs that has not the same name in the older version. I decomposed them but the difference was still here. 
Features were a bit different too, so I copy/paste the previous version, but I still have the problem.
If anyone has any idea what's going on here or what I may have missed, I'm all ears!


Capture d’écran 2022-11-03 à 14 18 26


Capture d’écran 2022-11-03 à 14 17 22

RosaWagner:
Ok, they are reported as new and missing by diffenator. It can be the feature that was doing a different substitution, that can provoke this, I don't know. Although if the result is the same, and if the codepoints didn't change for superiors/inferiors, I think we can safely merge it.

PR #5474, opened 2022-10-26
Use the --pre flag for installing Font Bakery on Github Actions

felipesanches:
Use the --pre flag for installing Font Bakery on Github Actions, so that we can benefit from the newest checks introduced in its more frequent pre-releases such as v0.8.11a2 released today:
https://pypi.org/project/fontbakery/#history

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -26,9 +26,12 @@ jobs:
           python-version: 3.8
 
       - name: Install dependencies
+        # The --pre flag below will ensure we use the latest Font Bakery pre-releases
+        # and benefit from its newest checks:
         run: |
           sudo apt install libharfbuzz-dev libharfbuzz-bin libfreetype6-dev libglib2.0-dev libcairo2-dev
           python -m pip install --upgrade pip
+          pip install --pre fontbakery
           pip install gftools[qa] pytest
           
       - name: Check fonts


chrissimpkins:
What is different between a pre-release and a release?

felipesanches:
> What is different between a pre-release and a release?

The **releases** are planned on github milestones and represent a "big bucket of stuff" I intend to do before moving to the next "bucket". For that reason, releases typically happen once every many months. There's no rush: it's ready whenever it's ready.

The goal with **pre-releases** is to be able to cut them more often, so that people can benefit from the newest stuff earlier.

And the most extreme option is to install from git main branch. With that users can get the very latest improvements, as they happen during development.

-----

Here are the considerations for having each of these three tiers:

(1) Some users complain that FontBakery requirements change too often and it feels like a moving target. It may also feel a bit arbitrary for these users to have new requirements out of the blue all the time. These users want stability. For them I recommend using the stable releases.

(2) Some other users want to run the newest checks as soon as possible. This is the case of the Google Fonts team of onboarders. And it may also be the case for other major font distributors who are using fontbakery. I suspect this may also be the case for Adobe Fonts, as @miguelsousa was the person who started cutting pre-releases to get easy access to the newest checks before the stable releases are ready.

(3) Then, some other users (I suppose mostly experienced font engineers who may have an interest in contributing to fontbakery development) may prefer to live on the "bleeding edge" of the git main branch codebase. This is risky because it is much more likely to contain bugs that could disrupt the normal operations (such as the daily work of the Google Fonts onboarders).

chrissimpkins:
Ty!

PR #5491, opened 2022-11-01
Remove pushed to_sandbox.txt items.

garretrieger:


RosaWagner:
Thanks @garretrieger! It's been weeks that I am trying to know why Noto Sans Mende Kikakui doesn't get pushed to sandbox, to you know what is blocking it or how we could have it pushed?

PR #5153, opened 2022-08-30
Noto Sans Khojki: Version 2.003; ttfautohint (v1.8.4.7-5d5b) added

simoncozens:
27491a7: [gftools-packager] Noto Sans Khojki: Version 2.002; ttfautohint (v1.8.4.7-5d5b) added

* Noto Sans Khojki Version 2.002; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/notofonts/khojki.git at commit https://github.com/notofonts/khojki/commit/.

PR #5493, opened 2022-11-02
Noto Serif Khojki: Version 2.003 added

simoncozens:
bab8c31: [gftools-packager] Noto Serif Khojki: Version 2.003 added

* Noto Serif Khojki Version 2.003 taken from the upstream repo https://github.com/notofonts/khojki.git at commit https://github.com/notofonts/khojki/commit/.


simoncozens:
Fixes interpolation issue in #5153.

PR #5508, opened 2022-11-04
Updated push lists

RosaWagner:


RosaWagner:
This is what break the CI: `lang/scripts/Toto.textproto` 
we always push `lang/language/…` et now we are trying with `script`. I gonna merge and we'll see if it annoys the eng team. If not we should update the CI to not fail that path.

PR #5516, opened 2022-11-07
Fix BioRhyme naming.

reitermarkus:


RosaWagner:
Thanks @reitermarkus for noticing that BioRhyme has some inconsistencies.

@chrissimpkins @davelab6 
- The URL on GF is https://fonts.google.com/specimen/BioRhyme (and not Bio+Rhyme)
- Font name tables are also camel case and consistent.

So imo the current METADATA.pb doesn't reflect the actual API metadata. But we would need confirmation to be sure.

m4rc1e:
I'm tempted for add-font to just fetch the best familyName e.g nameID 16 if the font has it, if not then nameID2. I think we already have a pr open for this.

RosaWagner:
I really think that the metadata.pb in the repo is not the actual metadata.pb, these are the current version, above Byo+Rhyme (in repo but not in prod), below is ByoRhyme (in prod but not in repo). So I think we can safely merge this to correct metadata.pb and re QA the font cause 1.001 never got to prod apparently.

Capture d’écran 2022-12-07 à 13 54 10

PR #5518, opened 2022-11-07
Add missing `EARLY_ACCESS.category` for Open Sans Condensed.

reitermarkus:


RosaWagner:
Open Sans has been transferred to the `ofl` repo. It has a width axis and Open Sans Condensed has never been release. I think this directory should be removed instead from `apache`.

Needs confirmation from @davelab6 and @chrissimpkins 
Maybe this is the opportunity to clean up the apache repo in a more general way, wdyt?

chrissimpkins:
Markus will you please let us know what you are attempting to do with the early release distribution?

reitermarkus:
We have a script to automatically import all Google Fonts to https://github.com/Homebrew/homebrew-cask-fonts.

If the `EARLY_ACCESS.category` exists, we use the https://fonts.google.com/earlyaccess page as homepage, otherwise we derive the `https://fonts.google.com/specimen/*` URL.

We also found that quite a few fonts are not actually released, i.e. don't have a matching `https://fonts.google.com/specimen/*` page, but also aren't on the [`earlyaccess`](https://fonts.google.com/earlyaccess) page.

chrissimpkins:
Thanks Markus.  The Early Access page/families should be considered obsolete and unsupported. We recommend that Homebrew not distribute these families.  I'd be happy to discuss how we can support families that you are interested in and are not available beyond the early access metadata that you are adding here.  Let's close this and take this conversation to the issue tracker or feel free to email me.

davelab6:
Agreed

davelab6:
I discussed this a bit further with @RosaWagner and @vv-monsalve and it seems since we have these files for all other EA families and EA isn't actually shut down yet, we should merge this. It will be a null change for GF.

davelab6:
> Thanks @reitermarkus. Open Sans has been transferred to the `ofl` repo. It has a width axis and Open Sans Condensed has never been released.

Oh, wait. Rosalie is right, Open Sans Condensed was never in early access, it seems. So this PR doesn't make sense to me. 

I believe the right thing is to delete the opensanscondensed directory. 

@reitermarkus will that work for you?

PR #5519, opened 2022-11-07
Unbounded: Version 1.700 added

yanone:
bf4c1e5: [gftools-packager] Unbounded: Version 1.700 added

* Unbounded Version 1.700 taken from the upstream repo https://github.com/googlefonts/unbounded at commit https://github.com/googlefonts/unbounded/commit/78f2d464776c37200bd2dc8437c1cf4eb2c6fc94.


RosaWagner:
@yanone I was wondering is NaN should be the only instance credited since the description speaks about more people. Are they copyright holders? Can you confirm with Luke?

There is no AUTHORS.txt in the upstream repo, it should be added to avoid licensing problem.

yanone:
I confirmed about the authors beforehand already. They will stay in the description, but are all secondary to NaN as an umbrella design studio.

PR #5521, opened 2022-11-07
Use target _blank.

andyjgf:


RosaWagner:
Please @andyjgf notify us about any changes to any font directory. At least a comment to explain why this change, and if this change should be tracked all through sandbox etc.

vv-monsalve:
Inspecting the file this PR is fixing the formatting of the option by adding the underscore `_blank`.

However, @davelab6 mentioned in the chat:

> `target=_blank` will opens the link in a tab, and the catalog adds that to all links in the About section iirc; `target=_self` would open the link in the same tab, and is the browser default. since the catalog overrides the browser default, possibly allowing re-overriding that back to self, might make sense

Would the latter `target=_self` make more sense? I've used it on a new PR 5527

RosaWagner:
As a personal opinion, I would not bother with a target at all in the upstream repo URL from the description_en.html here. It looks like something the catalog/API should decide to ensure a consistent link behaviour across the catalog — which it is already doing. If GF team then think that overriding the browser default is wrong, they can modify that at the catalog level, instead of the individual description_en.html level (which would mean modifying over 1300 files).

PR #5535, opened 2022-11-10
Noto Serif HK: Version 2.001-H2 added

simoncozens:
Metadata change here is different to the Sans versions; not sure why...

### DIFF
--- a/ofl/notoserifhk/METADATA.pb
+++ b/ofl/notoserifhk/METADATA.pb
@@ -9,8 +9,8 @@ fonts {
   weight: 400
   filename: "NotoSerifHK[wght].ttf"
   post_script_name: "NotoSerifHK-ExtraLight"
-  full_name: "Noto Serif HK"
-  copyright: "(c) 2017-2022 Adobe (http://www.adobe.com/)."
+  full_name: "Noto Serif HK ExtraLight"
+  copyright: "(c) 2017-2022 Adobe (http://www.adobe.com/), with Reserved Font Name 'Source'."
 }
 subsets: "chinese-hongkong"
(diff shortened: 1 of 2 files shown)


RosaWagner:
@simoncozens Noto Sans HK supports `languages: "za_Hant"   # Zhuang, Traditional Han`, is this something we want to add it here?

RosaWagner:
Looks like linespacing has increased

https://user-images.githubusercontent.com/12222436/202169913-57c9bef2-8d79-474e-b7a5-a1a16b9bb126.mov

simoncozens:
Linespacing issue is deliberate. Noto CJK statics from upstream have ascent=1160,descent=-288 but the upstream VFs have ascent=1000,descent=-200. As we are now updating Noto *Sans* CJK from static to VF, we are hotfixing the VFs to have compatible vertical metrics with the static versions.

But Noto Serif CJK was originally added as a VF before we knew about the static/VF metrics difference, so it was onboarded with the tighter VF metrics. I think it's better to hotfix the metrics here as well, so that Serif and Sans have compatible metrics.

simoncozens:
Please *hold* this one for now. There was another (slight) vertical metrics problem as I made the metrics compatible with the old sans, not from the old serif.

RosaWagner:
Shouldn't it be similar to new Sans, so the entire CJK family has the same v-metrics?

punchcutter:
At the time I'm not sure it was considered. Same for a lot of other things like variable fonts which didn't exist in the current state when this project was started. The CJK have always been a little odd and don't conform to all the new Noto rules that have evolved over the years.

PR #5450, opened 2022-10-20
Fix website links for Huerta Tipográfica

applecuckoo:
Fixes #5416 

Acme
Alegreya
Alegreya SC
Alegreya Sans
Alegreya Sans SC
Andada Pro
Bitter
Cambo
Gochi Hand
Piazzolla
Telex

PR #5555, opened 2022-11-16
Bitter Pro: Corrected credit

RosaWagner:
Author asked for that change by email. Bio will follow.


RosaWagner:
Description fixed in #5450

PR #5559, opened 2022-11-17
Archivo Narrow: Version 3.002 added

emmamarichal:
b23dc2d: [gftools-packager] Archivo Narrow: Version 3.002 added

* Archivo Narrow Version 3.002 taken from the upstream repo https://github.com/Omnibus-Type/ArchivoNarrow at commit https://github.com/Omnibus-Type/ArchivoNarrow/commit/9793ec77b6682a26bc7a6ed523ca65cc3cb90aec.


RosaWagner:
@emmamarichal can you check with them the metrics of /d ? looks like there is additional space in right sidebearing in some styles. Reported in the diff as kern missing, but looks more like a sidebearing problem.

kerns_missing

RosaWagner:
AH sorry I confused the before and after, ooops

PR #5328, opened 2022-09-27
Cairo: Version 3.120 added

yanone:
c62cf64: [gftools-packager] Cairo: Version 3.120 added

* Cairo Version 3.120 taken from the upstream repo https://github.com/Gue3bara/Cairo at commit https://github.com/Gue3bara/Cairo/commit/76d942d5bcc0ea491b6ff7d9d199bd0bb817f127.

PR #4458, opened 2022-03-31
Sofia Sans: Version 4.103 added

vv-monsalve:
1282608: [gftools-packager] Sofia Sans: Version 4.103 added

* Sofia Sans Version 4.103 taken from the upstream repo https://github.com/vv-monsalve/Sofia-Sans at commit https://github.com/vv-monsalve/Sofia-Sans/commit/888b575b900b10ab513a0ca94a7f71c60c7730f7.

PR #5577, opened 2022-11-18
Sofia Sans Condensed: Version 4.100 added

vv-monsalve:
65ee3f3: [gftools-packager] Sofia Sans Condensed: Version 4.100 added

* Sofia Sans Condensed Version 4.100 taken from the upstream repo https://github.com/vv-monsalve/Sofia-Sans at commit https://github.com/vv-monsalve/Sofia-Sans/commit/8cbfd2d7b5bd03c78e3c3045a2e426687f941cc6.


vv-monsalve:
This **test-PR** is part of a series that will update the previous test-PR #4458 with single-width fonts.

vv-monsalve:
- 🔥 FAIL: Check Google Fonts glyph coverage
Was agreed to be ignored.
- 🔥 FAIL: Check a font's STAT table contains compulsory Axis Values. 
I think this can be ignored too. @RosaWagner?

RosaWagner:
Yes I think there is a problem with the check, it is looking at all the name IDs and if the font was sliced, the name IDs such as 257, 258 etc are still there but not linked to a STAT or FVAR value, so it is reporting as missing. We should open an issue in gftools, it was also spotted in this PR: https://github.com/google/fonts/pull/5328

vv-monsalve:
Ok, great. Thanks for the details. What is really weird in this case is that the names reported on those ID are ok, and the only instances that were removed are actually no longer included. 

Screen Shot 2022-11-18 at 8 44 31


 I'll merge this PR to inspect the font in Dev. They are simple single weights now, but a lot has happened, so I prefer to be sure.

PR #5603, opened 2022-11-23
Mada: Version 1.4 added (#5594)

RosaWagner:
* Mada Version 1.4 taken from the upstream repo https://github.com/aliftype/mada.git at commit https://github.com/aliftype/mada/commit/.


RosaWagner:
@khaledhosny previous full name and postcript names were displaying "Regular", we need to keep this when upgrading to VF when possible, and it looks like this is a case where it is possible. This is to prevent third party distributors using GF to have conflicts when upgrading the catalog (some rely on name ID6).

RosaWagner:
Vertical metrics looks tighter: 

@khaledhosny This chapter explains how to handle vertical-metrics when upgrading a family: https://googlefonts.github.io/gf-guide/metrics.html

We indeed need linaGap=0 but since it was 100 before, it should be distributed between typoAscender and typoDescender to avoid reflow when upgrading.

khaledhosny:
It might be breaking change, but I prefer the tighter line spacing since this intended to be a UI typeface. The linegap was probably a mistake in the early release and I don’t think the applications I were using applied it (at least on in the contexts I were testing, which were mainly UI widgets and single line texts).

RosaWagner:
Unfortunately users tend to use the font regardless of design intentions, and this change might breaks thousand of websites' layout. Sometimes we make exception according to how much the font is served, so I'll ask and come back to you.

khaledhosny:
Here is a screenshot of Mada 1.3:


And here is Mada 1.4:


And here is 1.4 after distributing the previous line gap on ascender and descender, notice how the incresed line spacing pushed the last line out of the view :
1.4x

khaledhosny:
Since the goal here is to not break existing documents, the safest option would to keep the old vertical metrics unchanged. They are different from the current GF recommendation but they're what the font is already using. 

Distributing the line gap across asender and descender is no different from dropping it altogether since either will change the metrics for a portion of the users.

davelab6:
These GTK users aren't using the GF API. Would you be happy if we fork the project and maintain it with v metrics suitable for web users?

khaledhosny:
Whatever works for you.

RosaWagner:
@khaledhosny We can do that, although the linegap is quite an issue. In that case we would have to distribute the linegap between ascender and descender to not break website's layout, and match hhea with typo metrics (the goal is to not break websites for windows users, it's okay to introduce regression for mac users if it allows a better cross platform compatibility)

cf https://googlefonts.github.io/gf-guide/metrics.html#2-recalculating-the-vertical-metrics-for-an-upgraded-family

khaledhosny:
My suggestion is to revert to the metrics of the fonts currently served by GF without any changes, so everyone gets the same line metrics they are getting now and we don’t get to chose which users are allowed to suffer from regressions.

I’m a bit surprised about the popularity of this typeface, as I almost never encounter it in the wild. I wonder if people are using it for the Latin (specially that GF used to default to Latin preview when Arabic is not explicitly selected)?

khaledhosny:
OK, I’m implementing this and cutting a new release.

khaledhosny:
> @khaledhosny We can do that, although the linegap is quite an issue. In that case we would have to distribute the linegap between ascender and descender to not break website's layout, and match hhea with typo metrics (the goal is to not break websites for windows users, it's okay to introduce regression for mac users if it allows a better cross platform compatibility)
> 
> cf https://googlefonts.github.io/gf-guide/metrics.html#2-recalculating-the-vertical-metrics-for-an-upgraded-family

I’m a bit confused about the example in that page. If I’m understanding it correctly, I should keep the line gap in OS/2 and hhea tables, right? That contradicts the recommendation I got here. I’m happy to implement whatever metrics you are willing to accept, so it would be better to tell me exactly what values you want me to use for all the six metric entries.

RosaWagner:
Ah right, we forgot to update that part. Thanks for reporting.

We would need:
- use_typo_metrics: `on`
- typoAscender = 900+(linegap/2) = `950`
- typoDescender = 300+(linegap/2) = `-350`
- typoLinegap = `0`
- same metrics to report in `hhea` (950, -350, 0)
- winAcent = family Ymax = `1185`
- winDescent = family Ymin = `414`

khaledhosny:
I updated the files on main branch https://github.com/aliftype/mada, can you check if everything is OK before I cut a new release.

khaledhosny:
New release https://github.com/aliftype/mada/releases/tag/v1.5

RosaWagner:
Updated

## Mada: Version 1.5 added

---
 1426542: [gftools-packager] Mada: Version 1.5 added

* Mada Version 1.5 taken from the upstream repo https://github.com/aliftype/mada at commit https://github.com/aliftype/mada/commit/.

PR #5608, opened 2022-11-24
Noto Sans N'ko: added version 2.003 (#5166)

RosaWagner:
Follow up on #5166 and #5605 
Also updated description and articles.

We now have 2 families with the same file name ("NotoSansNKo"), and the same font name ID1 ("Noto Sans NKo") and seemingly directory name "notosansnko", but not the same API Name (1st "name" is METADATA.pb).

Former one has API Name: `Noto Sans N Ko` :::: has been place in directory `notosansnko_todelist` ::: is to de-list.
New one has API name: `Noto Sans NKo`

SO we expect 2 URLs:
[Noto+Sans+N+Ko](https://fonts.google.com/noto/specimen/Noto+Sans+N+Ko) ::: to de-list
[Noto+Sans+NKo](https://fonts.google.com/noto/specimen/Noto+Sans+NKo)


RosaWagner:
This is the manual diff and it looks good: 
[NKo_Diff.zip](https://github.com/google/fonts/files/10084034/NKo_Diff.zip)

If @simoncozens agrees then I'll merge.

Issue #4773, opened 2022-06-11
Upstream Makefiles don't consistently succeed


rsheeter:
https://github.com/googlefonts/oxidize/issues/25#issuecomment-1153000702 asserts `make build` should work on upstream repos. However, most of them don't have a Makefile and the ones that do seem not to consistently work. I located 24 repos that have Makefiles and 4 out of the 5 I tried failed:

```
(cd changa && time make build; rm -rf venv)
	fontmake.errors.FontmakeError: In 'Changa.glyphs' -> 'master_ufo/Changa.designspace': Generating fonts from Designspace failed: fonts contains incompatible glyphs: 'exclam', 'exclamdown', 'jeem-ar', 'jeem-ar.fina', 'question', 'question-ar', 'questiondown', 'tcheh-ar', 'tcheh-ar.fina'
	make: *** [Makefile:19: build.stamp] Error 1

(cd elmessiri && time make build; rm -rf venv)
	# IT WORKS!

(cd murecho && time make build; rm -rf venv)
	fontmake.errors.FontmakeError: In 'Murecho.glyphs' -> 'master_ufo/Murecho.designspace': Generating fonts from Designspace failed: fonts contains incompatible glyphs: 'bi-kata', 'bu-hira', 'do-hira', 'gi-hira', 'ka_semivoicedcomb-kana-kata', 'ki_semivoicedcomb-kana-hira', 'ki_semivoicedcomb-kana-kata', 'tu_semivoicedcomb-kana-kata'
	make: *** [Makefile:19: build.stamp] Error 1

(cd dosis && time make build; rm -rf venv)
[...]

PR #5631, opened 2022-11-26
Add repository_url where implied by archive

rsheeter:
Add repo_url where missing and implied by archive as suggested in https://github.com/google/fonts/issues/4773#issuecomment-1328101047.

Changes generated by https://github.com/rsheeter/google_fonts_sources/blob/main/add_missing_repo_urls.py.

### DIFF
--- a/apache/roboto/upstream.yaml
+++ b/apache/roboto/upstream.yaml
@@ -24,3 +24,4 @@ files:
   web/static/RobotoCondensed-Medium.ttf: static/RobotoCondensed-Medium.ttf
   web/static/RobotoCondensed-MediumItalic.ttf: static/RobotoCondensed-MediumItalic.ttf
   web/static/RobotoCondensed-Regular.ttf: static/RobotoCondensed-Regular.ttf
+repository_url: https://github.com/googlefonts/roboto-classic
--- a/ofl/abyssinicasil/upstream.yaml
+++ b/ofl/abyssinicasil/upstream.yaml
@@ -3,3 +3,4 @@ branch: master
 files:
   AbyssinicaSIL-2.200/OFL.txt: OFL.txt
   AbyssinicaSIL-2.200/AbyssinicaSIL-Regular.ttf: AbyssinicaSIL-Regular.ttf
+repository_url: https://github.com/silnrsi/font-abyssinica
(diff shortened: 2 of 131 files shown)


m4rc1e (commented):
I think the majority of these urls already exist in the METADATA.pb files e.g

https://github.com/google/fonts/blob/main/ofl/abyssinicasil/METADATA.pb#L20
https://github.com/google/fonts/blob/main/ofl/alkalami/METADATA.pb#L20
https://github.com/google/fonts/blob/main/ofl/andika/METADATA.pb#L49
...

rsheeter:
> I think the majority of these urls already exist in the METADATA.pb files e.g

They _should_ be in upstream though? - I was told that's how you are supposed to find sources. Or is upstream.yaml deprecated?

m4rc1e:
The upstream.yaml file isn't deprecated, it just holds metadata which isn't in the METADATA.pb file. The gftools packager will use both the METADATA.pb file and the upstream.yaml file. Personally, it would be nice if the fields in the upstream.yaml were added to the METADATA.pb file. It feels a bit strange having both files.

Do we have a one pages floating around for including sources instead of binaries in google/fonts? I'm curious.

rsheeter:
> The upstream.yaml file isn't deprecated, it just holds metadata which isn't in the METADATA.pb file

That's a bit of a nuisance to work with. Why do we need two files? Why don't we delete upstream.yaml and put *everything* into metadata?

rsheeter:
> it would be nice if the fields in the upstream.yaml were added to the METADATA.pb file

+1, lets make it happen. What is the authoritative list of fields an upstream.yaml can contain?

m4rc1e:
Here's the packager's yaml template: https://github.com/googlefonts/gftools/blob/main/Lib/gftools/template.upstream.yaml

Since the METADATA proto already includes repo urls and archive urls, I think we only need to include the files field.

rsheeter:
Closing PR since deleting upstream.yaml appears to be the preferred outcome.

rsheeter:
I believe I have added the fields in https://github.com/googlefonts/gftools/pull/639. Will anything break if I migrate the yaml content over?

m4rc1e:
> Will anything break if I migrate the yaml content over?

I'll take a look. We may need to update the packager which I'll do for you.

PR #5672, opened 2022-12-03
amiri: 1.000 added

khaledhosny:


m4rc1e (approved):
Thanks Khaled, this looks great. Marks have improved a lot.

Before:
Screenshot 2022-12-05 at 11 14 25


After:
Screenshot 2022-12-05 at 11 14 39


There's a positive ItalicAngle but I reckon this is ok because the font has a left leaning slant.

PR #5702, opened 2022-12-12
Noto Sans Cypro Minoan: Version 1.503; ttfautohint (v1.8.4.7-5d5b) added

simoncozens:
7cf939b: [gftools-packager] Noto Sans Cypro Minoan: Version 1.503; ttfautohint (v1.8.4.7-5d5b) added

* Noto Sans Cypro Minoan Version 1.503; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/notofonts/cypro-minoan.git at commit https://github.com/notofonts/cypro-minoan/commit/.


simoncozens:
Cypro-Minoan is an undeciphered script and we don't have any texts in it, so I have used the usual trick of just highlighting random glyphs from the range in the METADATA `sample_text` field. There *is* a glyphsets entry of this, so this should be enough to make it work.

PR #5704, opened 2022-12-13
Update axisregistry with new Mutation axis, revised Optical Size axis name casing

chrissimpkins:

PR #5706, opened 2022-12-13
.ci/run.sh: skip dirs which are not used for hosting fonts

m4rc1e:
cc @chrissimpkins


chrissimpkins:
I cherry picked this commit and merged it as part of #5704.  It fixed the issue. Thank you very much Marc!

PR #5720, opened 2022-12-15
Update width.textproto

xconsau:
Maybe the "Width" should be "Weight" here, as the description is already for the Width axis.


RosaWagner:
Thanks @xconsau, any modification to the axis registry should be made in: https://github.com/googlefonts/axisregistry

xconsau:
Thanks for pointing out, Rosalie. I wasn't aware of that repository.

PR #5737, opened 2023-01-03
Noto Traditional Nushu: Version 2.003 added

simoncozens:
6d99dab: [gftools-packager] Noto Traditional Nushu: Version 2.003 added

* Noto Traditional Nushu Version 2.003 taken from the upstream repo https://github.com/notofonts/nushu.git at commit https://github.com/notofonts/nushu/commit/.


simoncozens:
That's weird; it can render it, and that check passes for me locally. Also the packager deleted the article so we need that back.

simoncozens:
Check seems to be passing now. Silly thing.

RosaWagner:
@simoncozens all glyphs seem to be modified, it was all re-design? No problem with the baseline shift?

PR #5776, opened 2023-01-12
Sofia Sans: Version 4.101 added

vv-monsalve:
50e6814: [gftools-packager] Sofia Sans: Version 4.101 added

* Sofia Sans Version 4.101 taken from the upstream repo https://github.com/lettersoup/Sofia-Sans at commit https://github.com/lettersoup/Sofia-Sans/commit/185877d082cebb40facd65143326147431863688.


vv-monsalve:
@m4rc1e This PR introduces the complete range to the STAT table of the font (1 -- 1000). Hence the reported Fails that could be ignored.

m4rc1e (approved):

vv-monsalve:
Thanks for the approval @m4rc1e.

> The "Lettersoup" designer bio info exists both in the METADATA file and in the [catalog](https://github.com/google/fonts/blob/main/catalog/designers/lettersoup/info.pb) but is not displayed in the UI. What is missing for it to get displayed in the About section?

Do you know what would be needed or what is happening in this case?

PR #5788, opened 2023-01-17
Catalog updates for Shantell Sans

yanone:


catalog/designers/arrowtype/bio.html
```diff
@@ -1,11 +1,14 @@
...
+<p>
+    Arrow Type is a type foundry and studio practice based in Brooklyn, NY which specializes in custom type, type
+    design, and font development, run by Stephen Nixon. Previously, Stephen worked in digital product design and brand
+    experience design at IBM. In 2018, Stephen graduated with a Master&rsquo;s degree in Type and Media from The Royal
+    Academy of Art (KABK) in The Hague, Netherlands. In 2019, Google Fonts commissioned and published Arrow Type&rsquo;s
+    first release, <a href="https://fonts.google.com/specimen/Recursive">Recursive</a>. Today, Arrow Type has a focus on
+    creating fonts that are beautiful, uniquely useful, and tell a story.
 </p>
 <p>
     <a href="https://arrowtype.com/" target="_blank">arrowtype.com</a> |
```
m4rc1e, line 10: @thundernixon https://arrowtype.com/ is down.

vv-monsalve: Link is working for me.

m4rc1e: I get:

Screenshot 2023-01-19 at 16 52 02

arrowtype: Thanks for alerting me, Marc! Hmm, it’s working on my end, but I’ll look into whether I can somehow reduce redirects. Maybe I could set up the custom domain for GH pages in a slightly better way.

arrowtype: @m4rc1e maybe a silly question, but does it make any difference to take of the trailing `/`? That is, should we change the link from `https://arrowtype.com/` to simply `https://arrowtype.com`?

m4rc1e: Doesn't make a difference. Interestingly, if I wait it loads 😿. On Safari it is fine. Imo, this isn't a blocker. Just something you may want to improve in the future.

arrowtype: Makes sense! I appreciate you surfacing the issue. I did just recently tweak my site (moving from Netlify to GH Pages), and this is probably a side effect. I'll look into optimizing it further!

arrowtype: Oof, after a computer restart, I’m now getting the same error in Chrome. It does look like a DNS configuration thing. I’m making some adjustments, and I’ll get it sorted ASAP. Thanks again for bringing it to my attention!

arrowtype: Okay, from what I can tell on my end, I think I’ve solved the DNS and redirect issues. @m4rc1e, if you get the chance, could you please test it again?

m4rc1e: Works for me perfectly. Thanks.

arrowtype: Thanks for checking and letting me know!

PR #5796, opened 2023-01-18
Update TRIVIA.md to add special font

jamessouth:
The Binary Clock family implements a binary clock by representing numbers in base 2 as columns of dots. Only glyphs 0-9 are supported. Thank you!


RosaWagner:
Hi @jamessouth, thanks for your interest but this TRIVIA.md is pretty out of date, we probably gonna re-write it without that section.

PR #5799, opened 2023-01-19
Labrada: Version 1.000 added

emmamarichal:
5ecbfea: [gftools-packager] Labrada: Version 1.000 added

* Labrada Version 1.000 taken from the upstream repo https://github.com/Omnibus-Type/Labrada at commit https://github.com/Omnibus-Type/Labrada/commit/b15424f8840bda81f5dfa88f97bb634598faeadc.

 9af9d81: Labrada: description added


emmamarichal:
@Omnibus-Type Usually we only put the name of Omnibus in the designers (in metadata.pb) and that's what I did here. Let me know if you want to add Mercedes Jáuregui.

PR #5808, opened 2023-01-19
Add primary_script for most Noto fonts

tomasdev:
Included all Noto Fonts in this CL based on these conditions:
- Should have at least one language
- Should support the latin subset
- Primary script is the most repeated script in cases where multiple scripts are supported

Noto Sans, Noto Serif, Noto Display are exceptions because there is no clear primary script in those. Perhaps it should be Latn.

Internal reference b/261708463


ofl/notosansarabicui/METADATA.pb
```diff
@@ -106,3 +106,4 @@ languages: "wni_Arab"  # Ndzwani Comorian
 languages: "wo_Arab"  # Wolof, Arabic
 languages: "zdj_Arab"  # Ngazidja Comorian
 languages: "zlm_Arab"  # Malay, Arabic
+primary_script: "Arab"
```
chrissimpkins, line 109: The UI families are not used in fonts.google.com.  We probably do not need these metadata there?

ofl/notosansbengaliui/METADATA.pb
```diff
@@ -42,3 +42,4 @@ languages: "syl_Beng"  # Sylheti
 languages: "trp_Beng"  # Kok Borok, Bangla
 languages: "unr_Beng"  # Mundari
 languages: "unx_Beng"  # Munda
+primary_script: "Beng"
```
chrissimpkins, line 45: UI family that is not released on fonts.google.com

chrissimpkins (approved):
LGTM.  There are two "UI" families that are not released through fonts.google.com.  There is no harm in having the metadata listed in these families, but I don't believe that it is necessary in general.

chrissimpkins:
@simoncozens Any concerns about the addition of primary_script metadata across the Noto projects?  If we do this, do we need to update the builder so that it preserves the data as these family are changed going forward?

simoncozens:
No, it’s fine. Saves me a lot of work. :-)

PR #5809, opened 2023-01-19
Update to_sandbox.txt and to_production.txt with pushed families.

garretrieger:


chrissimpkins:
Resolved a merge conflict that I was responsible for in https://github.com/google/fonts/pull/5809/commits/c5d3e5b91ece394999b0abb433ec133e956454e3

PR #5813, opened 2023-01-20
Voltaire: Version 1.008; ttfautohint (v1.8.4.7-5d5b) added

emmamarichal:
edf3f63: [gftools-packager] Voltaire: Version 1.008; ttfautohint (v1.8.4.7-5d5b) added

* Voltaire Version 1.008; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/SorkinType/Voltaire at commit https://github.com/SorkinType/Voltaire/commit/0954355156a43a9dc7d53c32a7bedc63a187d456.

 ffebbb2: Voltaire: description updated


m4rc1e:
This is a very nice upgrade. The spacing is much better it will cause regressions. I'll ask in Thursday's meeting whether everyone is happy.

Screenshot 2023-01-20 at 15 13 35

Old

Screenshot 2023-01-20 at 15 13 56

New

m4rc1e:
ping @ebensorkin. Is the tighter spacing intentional? we're worried it may cause regressions and users will complain.

EbenSorkin:
I suspect that the spacing is not tighter - it just has kerning now and
didn’t before. I could check this to confirm. Also in theory it would be
possible to loosen the spacing to compensate so that on average it take
more similar space maybe but I’m not sure that makes sense. What do
you think?

EbenSorkin:
Looking on the iphone I didn't see enough to worry about but looking on the bigger screen of my iMac I see your point. If this is the kerning alone it is weirdly powerful. I'll have a look and propose an update next week.

m4rc1e:
Imo this is now a fixed family. The spacing has been improved by adding spacing so imo we should merge.

m4rc1e:
Note to self. Add note in description.

EbenSorkin:
I'm sorry, I can't quite figure out what happened. Did you add some space Marc? Or just onboard it as is?

PR #5853, opened 2023-02-01
subtree pull axisregistry

m4rc1e:


chrissimpkins:
Let's hold on the sandbox push until further notice.  The engineering team is working on an axis registration issue.

RosaWagner:
@chrissimpkins just found this PR, shall the axes be pushed to prod now?

chrissimpkins:
No axis registration pushes until we receive the all clear from Evan.  I will let you know.

vv-monsalve:
@chrissimpkins If the new workflow requires 

> We should add both the axis registration **and** the fonts to the to_sandbox list at the same time

Should we wait to move this subtree pull that includes `MORPH` axis to production until we have a PR for Kablammo? 
And therefore before #6212 moves into production?

PR #5868, opened 2023-02-07
Update th_Thai.textproto

arthusk:
for correct and standard of humanity


m4rc1e:
@chrissimpkins I don't know Thai but I think this PR does raise a collection wide issue we may have. How do we know that our sample texts are any good? If I do a google translate, it's clear the text is based on the UDHR. However, the translation tone is a bit harsh

> "By being indifferent and insulting to the values ​​of human rights."


When this author is suggesting instead,

> "Rights must be based on non-infringement."

simoncozens:
Our Thai text (and most of our texts) is taken from the Unicode.org translation here: https://www.unicode.org/udhr/d/udhr_tha.html - I think in turn these come from the UN High Commissioner for Human Rights' [UDHR Translation Project](https://www.ohchr.org/en/human-rights/universal-declaration/universal-declaration-human-rights/about-universal-declaration-human-rights-translation-project).

There are other conversations going on about which parts of the UDHR we should and should not use, but I'm wary of rewording anything which is essentially an official translation.

m4rc1e:
@arthusk thanks for your suggestion. Since we don't know Thai, we've had a native speaker review the current text. We have decided to use his suggestions instead.

PR #5874, opened 2023-02-08
Pontano Sans: Version 2.000;gftools[0.9.23] added

emmamarichal:
03e8d5a: [gftools-packager] Pontano Sans: Version 2.000;gftools[0.9.23] added

* Pontano Sans Version 2.000;gftools[0.9.23] taken from the upstream repo https://github.com/emmamarichal/PontanoSansFont at commit https://github.com/emmamarichal/PontanoSansFont/commit/9b35392760293f1016d0b6f82f87e61269db8c66.

 9e1c852: Pontano Sans: description updated


emmamarichal:
This PR comes from my fork, while waiting for the Google repo to be created. In the description, I put this link: https://github.com/googlefonts/pontanosans
So the google repo should be named exactly like this, or we should not forget to update the description.

m4rc1e:
Weight for the Regular has changed. However, this is probably intentional because extra weights have been added. I'll raise this tomorrow.

Screenshot 2023-02-08 at 16 59 35


Screenshot 2023-02-08 at 16 59 26

emmamarichal:
Note for @m4rc1e:

Stems thickness (with ‘H’ as reference):

**Old:**
*Regular: 146

**New:** 
*Light: 138
*Regular: 175
Medium: 194
SemiBold: 212
*Bold: 231

*: master

If we want to keep the same thickness for the regular, we should probably remove the master/instance Light, as it would become too close to the regular.

aaronbell:
Just adding a note here that it appears a GoogleFonts repro was never created for this font per @emmamarichal's comment above. Is that something that we want to correct? @RosaWagner

emmamarichal:
Hi @aaronbell, are you able to create one? Otherwise I can ask to Chris or Dave :)

emmamarichal:
@aaronbell I just checked, and we already have the google one: https://github.com/googlefonts/PontanoSansFont
did I answered to your question? or you were talking about something else?

aaronbell:
Oh! It looks like the metadata and license file are set incorrectly (since you’d submitted this without “font” in the url). I didn’t notice the correct one mentioned.

Probably be good to update that.

PR #5878, opened 2023-02-09
wixmadefortext: v1.200 added

m4rc1e:
Taken from the upstream repo https://github.com/wix/wixmadefor at release https://github.com/wix/wixmadefor/releases/tag/1.200


applecuckoo:
Fixes #5458?

chrissimpkins:
Adding blocked label until we address the new required public metadata field

RosaWagner:
this is in sandbox but without designer nor description

RosaWagner:
@emmamarichal there is a problem with the vietnamese diacritics. In the API they are made through the mark2mark feature, there might be an anchor missing or the combined accent glyphs

Capture d’écran 2023-04-26 à 15 04 12

emmamarichal:
@RosaWagner, ah, I didn't work on this font, @m4rc1e did all the exports, maybe he can take a look if it's a technical issue?

m4rc1e:
@madig was the project meant to have functioning Vietnamese?

m4rc1e:
If I download the family from sandbox and run `hb-view WixMadeforText-VariableFont_wght.ttf "ẵ" > wix.png`, it produces 

wix



cc @nathan-williams @simoncozens

garretrieger:
I only see 0x1eb5 in the vietnamese subset definition, however, the unicode range listing in the CSS for latin-ext has the range U+1E00-1EFF which includes 0x1eb5. I think the problem is that range needs to not cover 0x1eb5.

garretrieger:
This is now fixed on sandbox (you may need to clear cache to get updated CSS)

Issue #5913, opened 2023-02-20
Cairo Play glyphmapping issue


RosaWagner:
Weirdest glyphmapping in Cairo Play: for example `A` -> `AEacute`, `é` -> `ebreve`. Happens in VF from upstream repo + VF in GF zip.
Although when typing in GF specimen page, the font behaves correctly.
The issue cannot be reproduced with Cairo although the fonts are built from the same source. 

cc @yanone @m4rc1e

m4rc1e:
I don't understand the issue. Are you saying that  the "A" glyph are mapped to "Aacute" unicode? If I check the font, that isn't the case.

Are you checking the fonts in applications which support COLR tables?

RosaWagner:
Yeah by glyphmapping I meant, I type a letter and another one is displayed. To be more precise, it sometimes displays the next glyph ID of the one typed. This is the basic alphabet:

<img width="787" alt="Capture d’écran 2023-02-21 à 14 15 11" src="https://user-images.githubusercontent.com/12222436/220354971-ed9fec1d-ede8-4ae0-aea5-a97a0a9a59f6.png">


I checked it in different apps in Mac; textEdit, Indesign, Word. I know some don't support the COLR table, but they are supposed to display the proper glyphs with or without color. Nabla doesn't have this issue, that is why I think something is happening with Cairo Play.

yanone:
I made these on my old computer which is at home. I can check soon when I'm back. Though I didn't check for macOS compatibility explicitly, I have a feeling I should have seen that while working on it, which would suggest a macOS issue. I'll let you know.

yanone:
FLTR: Brave on latest macOS, Safari on latest macOS, latest Edge on Windows. Only Safari shows the anomaly (and all other native macOS apps).
<img width="2815" alt="Bildschirm­foto 2023-02-21 um 15 14 34" src="https://user-images.githubusercontent.com/175386/220369533-cc919b6e-e0e2-4eed-a762-a18dec6b1db6.png">

yanone:
> Although when typing in GF specimen page, the font behaves correctly.

Not in Safari. This is a legitimate bus crash: 
<img width="1479" alt="Bildschirm­foto 2023-02-21 um 15 32 07" src="https://user-images.githubusercontent.com/175386/220373386-f8ed0b6f-e80a-4935-bd36-9e07035e626d.png">

yanone:
[...]

PR #5924, opened 2023-02-23
Cairo: Version 3.130;gftools[0.9.24.dev9+gaf4c611] added

yanone:
444238f: [gftools-packager] Cairo: Version 3.130;gftools[0.9.24.dev9+gaf4c611] added

* Cairo Version 3.130;gftools[0.9.24.dev9+gaf4c611] taken from the upstream repo https://github.com/Gue3bara/Cairo at commit https://github.com/Gue3bara/Cairo/commit/617ed31e442f2be1ae31bdd7f3a88693fc41ae8b.


m4rc1e (commented):
Seems some kerns have gone missing for the reh forms

Screenshot 2023-02-23 at 11 31 37


I don't mind pushing this if we've fixed the terrible issues in #5913

yanone:
I'll take a look soon. Thanks

yanone:
We've actually lost *all* kerning because of the ongoing RTL kerning issue. I asked Simon to make his dumbifier available for pip to install.

anthrotype:
> dumbifier

what's that?

yanone:
Simon's solution to the impossibility of `glyphsLib` to support what Glyphs itself supports, "dumbing" down all smart stuff in Glyphs files without caring about round-tripping: https://github.com/simoncozens/dumbifier

It's a `glyphsLib`-based pre-processor. He solved the RTL-kerning in 13 lines of code: https://github.com/simoncozens/dumbifier/blob/main/Lib/dumbifier/rtl_kerning.py

I don't know when Simon was going to make that public. He did discuss it with the onboarding team. cc @simoncozens

anthrotype:
interesting. Can we please have those 13 lines of code in glyphslib and be done with this? :)

anthrotype:
like this? https://github.com/googlefonts/glyphsLib/compare/rtl-kerning-dumbifier?expand=1

yanone:
Please discuss this with Simon. Note that there are other processors in there (smart components, (or was it corner components?) other sanitizers), and I proposed one more layer sanitizer that he so far didn't incorporate.

Personally I don't care where what code resides. My job is to onboard fonts. What I notice is an ever-growing disconnection between what glyphsLib does and what we need it to do. So if Simon's aim is to host a package where he can quickly add solutions in a few lines of code, I support that 100%. But again, in the end I don't care. Stuff needs to _work_.

yanone:
Another thought would be to include the dumbifier as an external dependency into glyphsLib and have an official pre-processing stage by default. It just needs to be understood by all parties that these dumbifications aren't round-trippable. 

If round-tripping is a quasi-requirement of glyphsLib, then having the preprocessor in a separate library makes a lot of sense, tho sadly then it won't be active by default and available only to those who know about its existence.

anthrotype:
I don't think we need that. We already said we're ok with the RTL kerning not be rountrippable.

anthrotype:
hmm that rtl_kerning dumbifier looks suspiciously small, and in fact the tests @yanone wrote for https://github.com/googlefonts/glyphsLib/pull/838 do not pass when I apply by patch. It looks like the groups aren't updated as expected (in fact I only touch the kerning.py module, not the groups.py).
Mind taking a quick look at https://github.com/googlefonts/glyphsLib/pull/865?

yanone:
I'll check it tomorrow.

yanone:
@m4rc1e 
I committed a new font file with fixed kerning (using my old glyphsLib RTL kerning PR). I reported the red layer issue [here](https://github.com/googlefonts/glyphsLib/issues/866) but I don't expect a solution to be found anytime soon (and anyway I won't be able to use an improved glyphsLib version while simultaneously having RTL kerning), so IMO please go ahead with the update so that at least the macOS rendering disaster can be fixed before I'm off next week.
The red layers aren't a regression anyway; they're already visible in the live font. Will address that as soon as possible.

anthrotype:
@yanone I have updated my glyphsLib PR (https://github.com/googlefonts/glyphsLib/pull/865) to swap left/right groups for glyphs used in kerningRTL, do you mind giving it a try with Cairo to see if it works as you expect?

yanone:
@anthrotype IT WORKS!

RosaWagner:
Updated

## Cairo: Version 3.130;gftools[0.9.24.dev9+gaf4c611] added

---
 0521bb6: [gftools-packager] Cairo: Version 3.130;gftools[0.9.24.dev9+gaf4c611] added

* Cairo Version 3.130;gftools[0.9.24.dev9+gaf4c611] taken from the upstream repo https://github.com/Gue3bara/Cairo at commit https://github.com/Gue3bara/Cairo/commit/617ed31e442f2be1ae31bdd7f3a88693fc41ae8b.

yanone:
Holding this back because Khaled may have fixed the glyphsLib issue that led to the duplicate red layer. Will check today and report back

yanone:
This was actually the wrong PR. I didn't realize someone made a separate PR for Cairo non-Play. But the kerning issue applies to both, so I'll check

yanone:
@m4rc1e Please review again. The red layer issue is resolved, and kerning should work.
This PR now contains both Cairo & Cairo Play updates because of a mistake of mine.

ofl/cairo/METADATA.pb
```diff
@@ -1,8 +1,8 @@
 name: "Cairo"
 designer: "Mohamed Gaber, Accademia di Belle Arti di Urbino"
 license: "OFL"
-category: "DISPLAY"
 category: "SANS_SERIF"
+category: "DISPLAY"
```
RosaWagner, line 5: Display category should remain first in the list (secondary category)

PR #5929, opened 2023-02-24
Sigmar: Version 1.000; ttfautohint (v1.8.4.7-5d5b);gftools[0.9.24] added

emmamarichal:
8fad694: [gftools-packager] Sigmar: Version 1.000; ttfautohint (v1.8.4.7-5d5b);gftools[0.9.24] added

* Sigmar Version 1.000; ttfautohint (v1.8.4.7-5d5b);gftools[0.9.24] taken from the upstream repo https://github.com/googlefonts/sigmarone at commit https://github.com/googlefonts/sigmarone/commit/1fb4311d78953da0f943e0e4b806acdd076af81b.

 a40fa74: Sigmar: description added


m4rc1e:
Any reason why this hasn't been released as Sigmar One?

emmamarichal:
Yes! :) https://github.com/googlefonts/sigmarone/pull/31#issuecomment-1442078674

m4rc1e:
Thanks for the link. All sounds good. Only concern is that by naming it Sigmar, we won't be able to expand it by adding weights since the current Regular is a Black. Happy to merge if others are ok with this.

emmamarichal:
Ah yes indeed. I can still change that to Regular. @RosaWagner, @davelab6 what do you think?

RosaWagner:
I have no opinion, I transmitted what Dave's decision, but maybe we should talk about it more. cc @chrissimpkins 

Context: the problem is that Sigmar One is a small caps font and was not onboarded with "SC" in the family name. So either we call it Sigmar, or Sigmar Heavy / Two / etc.

m4rc1e:
The font is Regular, https://github.com/google/fonts/pull/5929/files#diff-e80717af937dc118f9bc8594574a4de67e42bdaf7371634e44fc592bc8823f11R6-R14. The issue is that the current Regular has the appearance of a Black weight. If we release the family as Sigmar in its current state, we won't be able to add a Regular weight in the future.

m4rc1e:
If we don't plan to extend, I'm happy for it to be released as is.

PR #5964, opened 2023-03-03
Add Nabla and Climate Crisis minisite_url metadata

### LINKED ISSUE #5957: Add minisite_url for Nabla and Climate Crisis
Hi there, I suggested the addition of the https://nabla.typearture.com/ site to the API metadata back in December of last year, at https://github.com/google/fonts/pull/5570#issuecomment-1320384563. I have also since found that the new font Climate Crisis has the minisite https://kampanjat.hs.fi/climatefont/index.html (it's even in the description) as well, so that should also be added.

chrissimpkins:
Closes #5957


chrissimpkins:
@RosaWagner Mind having a look to confirm that none of the fontbakery fails in these families are new following the fontbakery package update earlier today?  Shouldn't be a blocker to the merge here but we'll want to track and address any new issues that are popping up with the update.  Ty!

PR #5966, opened 2023-03-04
Bump CI workflow checkout action to v3 and setup-python action to v4

chrissimpkins:
Addresses node.js 12 Actions deprecation warnings that are being raised by GitHub

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/setup-python@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,9 +12,9 @@ jobs:
   test_server_files:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


chrissimpkins:
@m4rc1e mind having a look to confirm that you have no concerns about the Actions version bumps here?

m4rc1e:
Thanks!

I've just added a test font. If the CI works, I'll drop the test font commit and merge.

Issue #5965, opened 2023-03-03
Fix Tiro Tamil description


tiroj:
Tiro Tamil: https://fonts.google.com/specimen/Tiro+Tamil/about
Note: ‘About’ bug, not font bug.

**Describe the bug**
There is a small typo in the description text for Tiro Tamil: it says Kannada where it should say Tamil.

This has been fixed in the upstream source repo:
https://github.com/TiroTypeworks/Indigo/commit/aba722a749cd4c389cb548835b6831fe9ee1efc2

**Screenshots**
<img width="856" alt="Screenshot 2023-03-03 at 12 58 33 PM" src="https://user-images.githubusercontent.com/5614500/222828592-317b6aec-f466-449e-9b26-ebe6eded3ced.png">

chrissimpkins:
https://github.com/google/fonts/pull/5968

Thanks John!

PR #5968, opened 2023-03-04
Update Tiro Tamil DESCRIPTION.en_us.html

chrissimpkins:
Fixes https://github.com/google/fonts/issues/5965


RosaWagner:
I was surprised this fix wasn't showing in dev nor sandbox, as a matter of fact, the article is duplicating the description and taking precedence over it. So the article should also be fixed :/

PR #5984, opened 2023-03-08
Designer added: Bakken & Bæck

emmamarichal:
Designers of Schibsted Grotesk: https://github.com/google/fonts/pull/5962

Not sure about how these characters are supported: `æ` `&`


RosaWagner:
@emmamarichal I think the path will be problematic. In folder and file name, better to replace "&" by "and", and "æ" by "ae". In the the name string inside the file it's okay, no problem.

emmamarichal:
@RosaWagner done!

PR #6020, opened 2023-03-14
Noto Serif JP: Version 2.001-H2 added

simoncozens:


simoncozens:
There may be a small vertical metrics issue here; just checking.

m4rc1e (commented):
- non-CJK scripts have changed. Latin and Cyrillic is touch looser
- Color of Hiragana glyphs is worse

cjk-color

m4rc1e:
Rendering on Windows is a touch lighter now.

old:

Screenshot 2023-03-15 at 11 14 11

new:
Screenshot 2023-03-15 at 11 14 04


i reckon this is due to changing the format from otf to ttf.

simoncozens:
I agree that the colour of the hiragana looks worse (especially に, こ and た) in that browser sample. However, I wonder if something odd is going on with rendering. We will have lost hinting with the change from OTF static to TTF VF. It renders fine using hb-view and Cairo:

wght=100
shape

wght=400
shape

m4rc1e:
There are slight glyph changes which appear to be ok when skimming. However, judging 65k glyphs is insane. I'm wondering if we can do something we OCR to help assist us. I wouldn't mind exploring this at some point.

cc @chrissimpkins

m4rc1e:
> However, I wonder if something odd is going on with rendering.

@simoncozens the screenshots were taken from mac OS X. I've noticed this fading problem when a glyph has duplicate outlines or the curve direction is wrong.

simoncozens:
This seems to be a problem with Adobe's variable font conversion. Here is Adobe's static OTF builds (odd line is `Serif/OTF/Japanese/NotoSerifCJKjp-Regular.otf`) versus their variable *OTF* builds (even line is `Serif/OTF/Japanese/NotoSerifCJKjp-Regular.otf`) (i.e. both CFF):

Screenshot 2023-03-15 at 11 46 20

m4rc1e:
@simoncozens  I'm curious, has the Latin, Greek and Cyrillic been taken from https://github.com/google/fonts/pull/4956?

simoncozens:
> @simoncozens I'm curious, has the Latin, Greek and Cyrillic been taken from https://github.com/google/fonts/pull/4956?

No, I don't think they would use Noto for this; the upstream is Source Han Serif, which uses Source Serif as its LGC.

> Shall we also send this back to Adobe so they can fix the color?

We can raise an issue with them, certainly. I'd check with @chrissimpkins before working out whether to delay because of this; I'm not totally aware where the desire to upgrade Noto Serif CJK came from and if there's any urgency to it.

simoncozens:
> We can raise an issue with them, certainly. 

https://github.com/adobe-fonts/source-han-serif/issues/182

simoncozens:
> the curve direction is wrong.

Bingo. https://github.com/adobe-fonts/source-han-serif/issues/182#issuecomment-1476034026

PR #6021, opened 2023-03-14
Noto Serif KR: Version 2.001-H2 added

simoncozens:


simoncozens:
Here is the diffbrowsers output. New spacing in the Latin looks like an improvement, and Hangul spacing looks better. However, we are also seeing the contour direction problems in the Japanese hiragana - this will be true for all Noto Serif CJK fonts. Again, depending on priority we should wait for a new upstream release.

[notoserifkr-diff.zip](https://github.com/google/fonts/files/11050250/notoserifkr-diff.zip)

PR #6023, opened 2023-03-14
Noto Serif TC: Version 2.001-H2 added

simoncozens:


chrissimpkins:
@simoncozens Are you able to open new PRs for each of the Noto Serif CJK projects so that we can push the latest update from Adobe out to the prod web API?  Let's close all of these open PRs with the previous release of the family.  Thanks!

simoncozens:
Yes, working on it now.

PR #5982, opened 2023-03-08
Designer added: Julián Tunni

emmamarichal:
Designer of Julee: https://github.com/google/fonts/pull/5959

PR #6038, opened 2023-03-16
Julee metadatas

emmamarichal:
Updated: Juliàn Tunni instead Julian Tunni
see: https://github.com/google/fonts/pull/5982#issuecomment-1471662105


RosaWagner:
@emmamarichal between the designer file and the metadata.pb the accent is wrong (grave in one, acute in the other)

emmamarichal:
@emmamarichal acute is the good one, I'll change that!

PR #6058, opened 2023-03-22
updated to_lists

RosaWagner:


RosaWagner:
@m4rc1e I don't know why the CI is not happy, all path are correct. 
Maybe it is not liking: `lang/scripts/Nagm.textproto` (cause it's script, not lang)

PR #6064, opened 2023-03-22
GFK 2023 Q1 pre-launch fixes

elliotjaystocks:


elliotjaystocks:
Still errors in the checks, but they appear to be caused by the `to_sandbox.txt` file, which @nathan-williams has said we can ignore. (I believe it gets overwritten?)

rsheeter:
> Still errors in the checks, but they appear to be caused by the to_sandbox.txt file, which @nathan-williams has said we can ignore. (I believe it gets overwritten?)

to_sandbox.txt is manually maintained. This CL broke it by renaming a directory iiuc.

Lets err on the side of not merging if CI is failing in future please. If the CI is wrong we should fix the CI. In this case it seems to have been correctly flagging an error. Should be fixed now, no action required.

PR #6066, opened 2023-03-22
Shantell sans article

shailenpatel2:
added article and png to Shantell Sans font


nathan-williams (approved):
There is some bizarre spacing in the HTML markup. If there's an easy way to clean that up, that would be great. Not hugely important, though.

PR #6110, opened 2023-03-27
Noto Sans Oriya: Version 2.005 added

simoncozens:
4d7d100: [gftools-packager] Noto Sans Oriya: Version 2.005 added

* Noto Sans Oriya Version 2.005 taken from the upstream repo https://github.com/notofonts/oriya.git at commit https://github.com/notofonts/oriya/commit/.


RosaWagner:
@simoncozens can you check the diff? there are big changes

simoncozens:
Sorry, I should have left a note. Yes, I expect big changes. We fixed a lot of anchor attachment issues. I looked through the diffenator output and I am very happy with it - big improvement!

simoncozens:
Argh. We don't have the right masters for this to work with our toolchain. (No regular condensed master.) I'll see if Glyphs can synthesise one.

RosaWagner:
Ah one should always refresh the page.

PR #6118, opened 2023-03-27
Noto Serif Gurmukhi: Version 2.004 added

simoncozens:
03fde36: [gftools-packager] Noto Serif Gurmukhi: Version 2.004 added

* Noto Serif Gurmukhi Version 2.004 taken from the upstream repo https://github.com/notofonts/gurmukhi.git at commit https://github.com/notofonts/gurmukhi/commit/.


simoncozens:
It's fine. This font has both Latin and Gurmukhi punctuation; the Gurmukhi punctuation is the default and has names like `braceright.guru`.

PR #6129, opened 2023-03-29
Designer added: Smartsheet Inc

emmamarichal:
Designer of Outfit project https://github.com/google/fonts/pull/6126


catalog/designers/smartsheetinc/bio.html
```diff
@@ -0,0 +1,2 @@
+<p>Smartsheet Acquired Outfit.io - co-creators of the Outfit typeface</p>
```
RosaWagner, line 1: This is not a biography

emmamarichal: Yes, I was writing to the authors to modify it!

PR #6130, opened 2023-03-29
Noto Sans Devanagari: Version 2.004 added

simoncozens:
0fa956a: [gftools-packager] Noto Sans Devanagari: Version 2.004 added

* Noto Sans Devanagari Version 2.004 taken from the upstream repo https://github.com/notofonts/devanagari.git at commit https://github.com/notofonts/devanagari/commit/.


m4rc1e (commented):
Is the missing acaron intentional? If it is, then LGTM.

simoncozens:
I haven't done anything to remove an acaron. Was it maybe in GF Latin Core but it isn't any more?

m4rc1e:
Yep, that's exactly what's happened. Let's merge anyway.

Issue #2202, opened 2019-10-16
backquote is not inline with other characters in Noto+Sans font


sdarshanam:
Representation of the text with backqoutes is not inline with the rest of the characters, alignment goes for a toss. Here is the image:

![image](https://user-images.githubusercontent.com/748942/66884392-e9942600-ef85-11e9-85a4-fdb93507763a.png)

thlinard:
You intended to use a ‘ ‎(2018 LEFT SINGLE QUOTATION MARK), but this is a grave accent, and probably the grave accent symbol : ‎` (U+0060 GRAVE ACCENT).

Anyway, the problem is fixed in the Adobe version:
https://fonts.adobe.com/fonts/noto-sans

dgealow:
> You intended to use a ‘ ‎(2018 LEFT SINGLE QUOTATION MARK), but this is a grave accent, and probably the grave accent symbol : ‎` (U+0060 GRAVE ACCENT).
> 
> Anyway, the problem is fixed in the Adobe version: [fonts.adobe.com/fonts/noto-sans](https://fonts.adobe.com/fonts/noto-sans)

` (U+0060 GRAVE ACCENT) is not supposed to be a combining diacritic. The combining version of the grave accent is U+0300 (combining grave accent).

This problem appears to be due to Google serving an ancient version of Noto Serif -- this has been fixed in the main branch of notofonts for over a year. See https://github.com/notofonts/latin-greek-cyrillic/issues/11

dgealow:
Looks like this is finally fixed in https://github.com/google/fonts/pull/6134 ! Contributors can probably close this issue now. Combining and non-combining graves now work as intended in https://fonts.google.com/noto/specimen/Noto+Serif/tester and it looks fixed on the site I originally noticed this in, https://www.seximal.net/punctuation.

PR #6134, opened 2023-03-29
Noto Serif: Version 2.010 added

simoncozens:
7d6c6ad: [gftools-packager] Noto Serif: Version 2.010 added

* Noto Serif Version 2.010 taken from the upstream repo https://github.com/notofonts/latin-greek-cyrillic.git at commit https://github.com/notofonts/latin-greek-cyrillic/commit/.


simoncozens:
Expected changes:

```
This release:
* Fixes soft-dotting of i and j glyphs with accents (#64, thanks to Denis Jacquerye)
* Improves the position of marks below small capital J (#60)
* Removes the erroneous U+11AB0/11AB1 glyphs (#405)
* Adds slashed pnum/onum zeros (#159)
* Adds the U+10FB GEORGIAN PARAGRAPH SEPARATOR character for medievalists (#37)
* Adds the U+2183 ROMAN NUMERAL REVERSED ONE HUNDRED character (#129)
```

RosaWagner:
@simoncozens there is no diff images in the artifacts. We should also discuss the change of postscript name, see if it annoys third party distributors

RosaWagner:
@simoncozens I generated the diff for Noto Serif:
[out.zip](https://github.com/google/fonts/files/11220209/out.zip)

RosaWagner:
I must say, I am a bit lost in the changes. What I would raise would be:
- lots of "missing encoding glyphs" in the upright styles

- top anchor might be off on upright epsilon (I circled also eta, but it's actually okay)
Capture d’écran 2023-04-13 à 10 09 41

- top anchor looks off in italic s
Capture d’écran 2023-04-13 à 10 13 35

- CYRILLIC SMALL LETTER SHHA that now looks like a latin h (no idea if it's okay)

https://user-images.githubusercontent.com/12222436/231712387-c00e0484-877b-4bb5-b0ea-693707186ef5.mov

simoncozens:
> I must say, I am a bit lost in the changes.

Unfortunately this font has developed a lot since the last onboard. Hopefully we will be able to keep more up to date from now on.

> lots of "missing encoding glyphs" in the upright styles

I am not worried about this. These have moved to other fonts - many symbols did not belong in a "serif" font.

> top anchor might be off on upright epsilon 
> top anchor looks off in italic s

Good catch, thank you. I'll do a new release.

> CYRILLIC SMALL LETTER SHHA that now looks like a latin h (no idea if it's okay)

The Cyrillic is currently being redesigned, so am not worried about this.

simoncozens:
Updated

## Noto Serif: Version 2.011 added

---
 5d24da0: [gftools-packager] Noto Serif: Version 2.011 added

* Noto Serif Version 2.011 taken from the upstream repo https://github.com/notofonts/latin-greek-cyrillic.git at commit https://github.com/notofonts/latin-greek-cyrillic/commit/.

dgealow:
Would this PR solve the issue with backticks mentioned in https://github.com/notofonts/latin-greek-cyrillic/issues/11 and https://github.com/google/fonts/issues/2202 ?

simoncozens:
No. Those issues are an issue with the font, and they're still open; this PR just takes the existing font and onboards it to Google Fonts.

RosaWagner:
@simoncozens it got confirmed by email that changing the PS name would be problematic for adobe

simoncozens:
OK. That was not an intentional change; it just happened as the file went through the build process:

```
$ fontmake -o variable -g sources/NotoSerif.glyphspackage
...
$ ttx -o - -t name variable_ttf/NotoSerif-VF.ttf
...
     <namerecord nameID="6" platformID="3" platEncID="1" langID="0x409">
      NotoSerif-Regular
    </namerecord>
```

I am not sure how to explicitly set the PostScript name in a .glyphs file when doing a variable font build.

simoncozens:
I've been looking at the glyphset issues.

The missing symbols and math symbols are not a problem because Google Fonts would never serve these anyway (we would need to have the `symbols` and/or `math` subsets declared in the METADATA.pb, and we have never had that).

That leaves the Coptic letters which would be served under the greek subset. That is a little concerning, because now we have no serif Coptic any more. I don't know why not, but it happened a very long time ago; the earliest sources in the noto-source repo (v1.903, 7 years old) do not have the Coptic letters.

I don't have a solution. They're just gone, I'm afraid.

RosaWagner (approved):

chrissimpkins:
@RosaWagner are we prepared to merge this?

RosaWagner:
@chrissimpkins I would wait for @m4rc1e's approval first, but after the discussion we had, I don't see issues on my side anymore.

m4rc1e:
I'll look at this after lunch.

m4rc1e:
Only thing I'm not sure of is the Cyrillic Omega

Old:
Screenshot 2023-05-18 at 14 47 25

New:
Screenshot 2023-05-18 at 14 47 31

Imo, old form looks more like an Omega but I could be wrong. If this change is intentional, ignore me and merge.

m4rc1e:
There may not be anything we can do but the color on the superior glyphs was better before imo. New looks a bit too light.

old:
Screenshot 2023-05-18 at 14 52 53



new:
Screenshot 2023-05-18 at 14 52 17

simoncozens:
>  If this change is intentional, ignore me and merge.

I think this is intentional, given that the *next* version of Noto Serif (I like to keep you busy) which has a redesigned Cyrillic from Ksenya, also has the same form of capital omega.

moyogo:
@simoncozens @m4rc1e The cyrillic omega was round in hands with round cyrillic o and pointy in hands with pointy cyrillic o. It would make more sense to have a round cyrillic omega in antiqua. One reason this current shape is used is to avoid confusion with A64D cyrillic broad omega ꙍ but that one should be broad.
The glyph isn’t part of the latest Cyrillic PR so I don’t think it’s designed has been reviewed yet.

JanisE:
> I've been looking at the glyphset issues.
> 
> The missing symbols and math symbols are not a problem because Google Fonts would never serve these anyway (we would need to have the `symbols` and/or `math` subsets declared in the METADATA.pb, and we have never had that).
> 
> That leaves the Coptic letters which would be served under the greek subset. That is a little concerning, because now we have no serif Coptic any more. I don't know why not, but it happened a very long time ago; the earliest sources in the noto-source repo (v1.903, 7 years old) do not have the Coptic letters.
> 
> I don't have a solution. They're just gone, I'm afraid.

Are you referring to the lost characters in this comparison?

NotoSerif-changes

Why are you removing characters from a font, and say it's not a problem?

First of all, Google Fonts would have served the characters if asked nicely (see my issue #9687).

Second of all, these fonts are downloaded fully and used in all kinds of situations, not just in webfonts. For example, "Fonts-noto" Ubuntu package downloads the fonts and lets users use them locally as installed fonts. So, with the current "Fonts-noto" update a lot of characters just gone missing that were previously there.

Can you somehow get them back? If a character is designed in Noto Serif style, it shouldn't just get removed. Besides, if Noto's aim is to cover all characters, then they would need to be added at some point anyway.

simoncozens:
> For example, "Fonts-noto" Ubuntu package downloads the fonts and lets users use them locally as installed fonts. So, with the current "Fonts-noto" update a lot of characters just gone missing that were previously there.

No characters have been removed from Noto Serif. If Ubuntu downloads the fonts from [the Noto release site](https://github.com/notofonts/notofonts.github.io), which it should do, then it will get the full fonts.

JanisE:
I'm not an expert of fonts and the automated processings you have here and at that other site, I just hope you understand the problem for a regular font user.

All this font infrastructure seems overwhelming to me. I got to https://github.com/notofonts/notofonts.github.io/tree/main/fonts/NotoSerif, and there are so many versions of the font in so many folders...

I tried to open several TTF files there, both variable and not variable, hinted and unhinted an whatnot, and I am yet to find where the infinity glyph resides there. Can you, please, point me to the actual NotoSerif source of all these files?

simoncozens:
> I am yet to find where the infinity glyph resides there.

As far as I'm aware, Noto Serif has never contained the infinity glyph. (At least not in the past ten years - I just checked all the sources we got from Monotype.)

JanisE:
Please, check out commit bfafca8339 of this repository (year 2023).

The difference between that and the next (9d5f2ed0bc) commit is what I show in the animated GIF in my post above.

simoncozens:
Oh my, that was a version of Noto Serif which is older than ten years! Of course, if you like that one, you are free to use it.

JanisE:
It's not about liking it but about the characters I need.

So, Noto is being developed all the time, its versions are changing, and sometimes moving forwards Noto developers decide just to remove glyphs without leaving anything in their place. That's what's happening right?

How do you think, what would be the proper way to get these glyphs back into the Noto font?

simoncozens:
> Noto developers decide just to remove glyphs without leaving anything in their place. That's what's happening right?

No, that's not what is happening.

The first step is to understand what Noto is. Noto is a *collection* of fonts. "Noto Serif" is one of the fonts in the collection. Noto Serif is not a mega-font which not covers every single codepoint - this would be impossible because there is a 64k glyph limit on any one font. So Noto Serif is designed to used within the context of a font fallback stack which contains multiple Noto fonts which *together* cover all the codepoints that you wish to use. We may sometimes move codepoints from one of the Noto fonts into another one, but we don't remove codepoints from Noto leaving nothing in their place.

Noto Serif is designed to cover basic and extended Latin characters. If you want to use Arabic and Latin, you need to also use Noto Naskh Arabic as well as Noto Serif in your font fallback stack. If you want to use mathematical symbols (and it sounds like you do), you need to add Noto Sans Math to your fallback stack - the infinity character you want has moved to there.

If you want to know which Noto fonts contains which characters, please check out the [NotoVerse](https://notofonts.github.io/overview/).

JanisE:
I see, thank you for the explanation!

I just tried out that the infinity glyph is available for "Noto Sans Math" and "Noto Sans Mono":
* https://fonts.googleapis.com/css2?family=Noto+Sans+Math&text=%E2%88%9E
* https://fonts.googleapis.com/css2?family=Noto+Sans+Mono&text=%E2%88%9E

And these do not work (the reference returned within is invalid):
* https://fonts.googleapis.com/css2?family=Noto+Sans&text=%E2%88%9E
* https://fonts.googleapis.com/css2?family=Noto+Serif&text=%E2%88%9E (this one worked before the changes of this PR)

OK, yet another thing to pay attention when selecting fonts.

simoncozens:
> OK, yet another thing to pay attention when selecting fonts.

Yes. It's always the case that you need to check whether the font you want contains the codepoints you want, because not all fonts contain all of Unicode.

PR #6141, opened 2023-03-30
Update figcaptions

tomasdev:
Moved all <figcaption> to be inside a <figure>

Added a validator so we don't break it in the future.


.github/workflows/knowledge_graph.py
```diff
@@ -175,6 +176,12 @@ def _check_md_file_contents(repo_root: Path, md_file: Path, ast: List[MdValue])
         if re.search(' id="[^"]+"', text):
             print("INVALID ", _safe_relative_to(repo_root, md_file), "attr.id not allowed:", text)
             return False
+    f = open(md_file,"r")
+    content = "".join(f.readlines())
```
tomasdev, line 180: I needed to use the file as text because the AST has different entries for "<figure>", "</figure>", the image, and the figcaption, making it impossible to validate within the iterator.

nathan-williams: Thanks for the proactive explanation.

I think we need to revise the regex to make sure that we don't match the closing `figcaption` tag to the closing tag of some other `figure`. Instead of `.*`, suggest using `[\s]*`. I don't recall what the matcher is for newlines, but you'll want to add that to the character class.

tomasdev: It is tested, as seen in the files in this PR where multiple <figcaption> had to be updated to make it pass.

elliotjaystocks:
@tomasdev @nathan-williams Looks good — thanks for fixing!

PR #6164, opened 2023-04-06
Update Noto Sans Math to v2.539 full so we have a MATH table

rsheeter:


rsheeter:
Looks to me like the error is with the code rather than the font (broken_heart ERROR Failed with UnboundLocalError: local variable 'link' referenced before assignment) so proceeding.

Issue #6076, opened 2023-03-23
Add kana-only fonts


RosaWagner:
They should all be visible in sandbox, but not in prod. At some point hey were pushed to prod but de-listed because of sample text showing tofu.

- [x] #2864
- [x] #2983
- [x] #2880
- [x] #2867
- [x] #2874
- [x] #2918
- [ ] #2934
- [x] #2866
- [x] #2871
- [x] #2870
- [x] #3680
- [x] #3308

simoncozens:
OK, so we have a reasonable number of fonts which were designed for *display* Japanese. The backstory is this: Japanese has three scripts, one of which - *kanji* - is so massive (minimum of 2000 characters for daily use) that it's infeasible for most foundries to design. But for display use, you can get away with just using the other two (the "kana" scripts, *hiragana* and *katakana*) - as an analogy, if you have a Latin display font which is caps-only, you can still make a lot of poster headlines, etc.

The problem is that when fonts get detected as Japanese, they get the `ja_Jpan` language added to them which expects them to have a full Japanese character set. The sample text has *kanji*, the fonts don't, so we get tofu.

A potential solution is that we have two pseudo-languages already in the lang repo, `ja_Kana` and `ja_Hira` for *katakana* and *hiragana*. I suggest that we fill these pseudo-languages with sample text, transliterating the *kanji* into each kana script, but that we don't add exemplar chars to them, so they will not be automatically picked up by the packager. Instead, kana-only fonts will get manually assigned with these two languages.

I can prepare a PR for the lang repo for this, and a PR here to tag the Japanese fonts with the pseudo-languages.

RosaWagner:
Sounds good. There is a test I made few month ago that failed, probably because of the missing exemplar characters: https://github.com/google/fonts/pull/5442

Once the PR in the lang repo is merged, don't forget to make a release + pull subtree here ;)

chrissimpkins:
Assigning myself to chat with the eng team about this so that we can confirm that the language definition approach will work on the front end

simoncozens:
The lang definition has already been added to lang v0.4.6, and that was merged to google/fonts in https://github.com/google/fonts/pull/6082

chrissimpkins:
I discussed this issue with Simon today.  As I understand it, these merged language updates have landed in production and these families are no longer blocked on this issue.

RosaWagner:
@chrissimpkins @simoncozens we still need to add the proper primary script, right? + repush the families to sandbox

[...]

PR #6172, opened 2023-04-06
Add Kana script and languages to DarumaDropOne

simoncozens:
This is a test of the approach for kana-only fonts described in #6076.


RosaWagner:
@simoncozens this is looking good in dev

simoncozens:
Although we seem to have duplicated the family...

Screenshot 2023-04-06 at 15 43 08

It would also be ideal if we could have some hiragana samples in there as well, but I don't know if that's possible.

RosaWagner:
the duplication comes from a naming issue 2 years ago

PR #6198, opened 2023-04-19
Lateef: Version 4.000 added

emmamarichal:
86cbbfd: [gftools-packager] Lateef: Version 4.000 added

* Lateef Version 4.000 taken from the upstream repo https://github.com/silnrsi/font-lateef at commit https://github.com/silnrsi/font-lateef/commit/.


RosaWagner:
same as #6201, we need https://github.com/googlefonts/glyphsets/pull/92 merged and pushed to prod to move forward.

PR #6203, opened 2023-04-19
Karla: Version 2.003;gftools[0.9.28] added

emmamarichal:
7b2df1a: [gftools-packager] Karla: Version 2.003;gftools[0.9.28] added

* Karla Version 2.003;gftools[0.9.28] taken from the upstream repo https://github.com/googlefonts/karla at commit https://github.com/googlefonts/karla/commit/e04a98d57933cf18a8e4913506aa9aac99a1756c.


RosaWagner:
@emmamarichal the description was not updated, can you remove the mention to the early access? There is also the mention of the font supporting Tamil… we may need to investigate that.

PR #6345, opened 2023-06-06
Aleo: Version 2.001;gftools[0.9.29] added

emmamarichal:
630dd51: [gftools-packager] Aleo: Version 2.001;gftools[0.9.29] added

* Aleo Version 2.001;gftools[0.9.29] taken from the upstream repo https://github.com/AlessioLaiso/aleo at commit https://github.com/AlessioLaiso/aleo/commit/ce875e48d9983031648e87f38b7a269f4fbf5eb5.

 b279c4b: Description updated according the upgrade


emmamarichal:
v-metrics are different because previously linegap was set up to 200 and Use Typo Metrics wasn't enable

RosaWagner:
If use_typo_metrics was not enabled, then the Win metrics were used to display line spacing. Therefore, the previous win metrics should be taken to populate typo and hhea to avoid regression on windows, like describe in the second case of this section: https://googlefonts.github.io/gf-guide/metrics.html#2-recalculating-the-vertical-metrics-for-an-upgraded-family

emmamarichal:
Yes, I know, but when I tried to use the previous win metrics, I still had a difference. The only way to get the same result was to addition the previous linegap (200), to the typoAscenders and descenders, and enable Use_Typo_metrics

PR #6185, opened 2023-04-13
Wavefont: Version 3.003 added

RosaWagner:
596861f: [gftools-packager] Wavefont: Version 3.003 added

* Wavefont Version 3.003 taken from the upstream repo https://github.com/dy/wavefont at commit https://github.com/dy/wavefont/commit/89ededb4d5bbf9f837ab095081e9249fdaab219f.

 9338cb9: Wavefont: updated description and metadata.pb

fixes #4631

PR #6358, opened 2023-06-09
Subtree pull Axis Registry introducing `YELA`

vv-monsalve:
- [x] Includes `YALN - Vertical Alignment` axis required by Wavefont

This PR should navigate through the pipeline simultaneously with #6185


vv-monsalve:
@RosaWagner added it to the old good Traffic Jam. Let me know if it should be differently now.

vv-monsalve:
Dave has a concern about the type of axis (if it should be an absolute range per-mile). So blocking it until we review it again tomorrow.

vv-monsalve:
Axis was renamed to `YELA`. This PR is now ready for review.

PR #6378, opened 2023-06-15
Bricolage Grotesque: Version 1.000;gftools[0.9.30] added

emmamarichal:
92f1a0b: [gftools-packager] Bricolage Grotesque: Version 1.000;gftools[0.9.30] added

* Bricolage Grotesque Version 1.000;gftools[0.9.30] taken from the upstream repo https://github.com/ateliertriay/bricolage at commit https://github.com/ateliertriay/bricolage/commit/95ddb0da9087a5f49f4616faea519e9e27d80a99.

 b936917: Description updated


emmamarichal:
We need to add the minisite link to the description (not ready yet). But @RosaWagner, the review can be started in the meantime.

RosaWagner (approved):

RosaWagner:
@emmamarichal just seeing that it has an opsz axis:

- the default should _always_ be included in the STAT table and the FVAR instances, and it should be 14 if it's not overrode in METADATA.pb
- The API is providing 3 locations: 18, 28, 36

Since the chosen default in the font is 24, we should probably move it to 28 to match one of the API location, and add the axis override entry for that value.

emmamarichal:
I put 'blocked" since I'll onboard the 1.001 version with the opsz values changes (and small fixes in the design)

PR #6389, opened 2023-06-16
Fix mistune create_markdown to work with mistune 3

rsheeter:
We don't pin mistune's version; v3 altered the way you ask for an ast renderer per https://mistune.lepture.com/en/latest/upgrade.html#astrenderer.

CI should be green again.


RosaWagner (approved):

RosaWagner:
Thanks @rsheeter

PR #6458, opened 2023-06-27
Noto Sans Cham: Version 2.003 added

simoncozens:
7718df6: [gftools-packager] Noto Sans Cham: Version 2.003 added

* Noto Sans Cham Version 2.003 taken from the upstream repo https://github.com/notofonts/cham.git at commit https://github.com/notofonts/cham/commit/.


simoncozens:
```
This release:

- Fixes many collisions between subjoined marks and following bases (#7)
- Swaps the order of features to allow the calt feature to apply (#8)
```

PR #6468, opened 2023-06-28
Lang v051

RosaWagner:
Mega change in the lang folder


RosaWagner:
@moyogo 
I should probably wait for https://github.com/googlefonts/lang/issues/91 and https://github.com/googlefonts/lang/issues/92 to get fixed and added to this PR before merging to avoid pushing multiple times the same textprotos.

Issue #4354, opened 2022-03-09
SIL fonts onboarding (Summary)


RosaWagner:
See if we onbaorded all these fonts, and create issues for the missing ones (and then link issue here)

**Most important:**
- [x] #4369
- [x] #3756 
   - [x] de-list Andika New Basic
- [x] #3757 
    - ~Charis SIL Compact~
- [x] #3755
    - [x] de-list Gentium Basic
    - [x] de-list Gentium Book Basic
    - [x] Add Gentium Book Plus
    - [x] Add Gentium Plus
    - ~Gentium Plus Compact~
    - ~Gentium Book Plus Compact~
- [x] #2509
- [x] #4424

**Next priority:**
- [x] #4420
- [x] #4421
- [ ] #4512
- [x] #4514
- [x] #4515
- [x] #4516
- [x] #4517
- [ ] #4518
- [x] #4520
- [x] #4834 
- [x] #6190
- [x] #6189
- [x] #6188
- [x] #5243 
- [x] #6463
- [ ] #6493
- [x] #6496
- [x] #6494

**To not onbaord**
cf https://github.com/google/fonts/issues/4354#issuecomment-1079066885
- Apparatus SIL
- Awami Nastaliq
- Bosa
- Bosa Lisu
- CR Udom
- Dai Banna SIL
- Doulos SIL (#5797)
- Doulos SIL Compact
- Ezra
- Galatia SIL
- Kay Pho Du
- Kyebogyi
- Khmer Busra
- Khmer Mondulkiri
- Khmer Oureang
- Khmer Ratanakiri
- LisuTzimu
- Namdhinggo SIL
- Nastaliq Navees
- Nokyung
- Payap Lanna
- Salaowu
- Sapushan
- Scheherazade (replaces by Scheherazade new)
- ShiShan
- Sophia Nubian
- Tagmukay
- Taogu
- Zaghawa Beria

jvgaultney:
Many of these fonts probably shouldn't be on Google Fonts as they are old and no longer maintained, have been replaced by other fonts, or are still pre-release. Our suggestion would be that you consider adding these with the following priority:

Most important:
- Abyssinica
- Andika
- Charis SIL
- Gentium
- Harmattan
- Scheherazade New

Next priority:
- Akatab
- Alkalami
- Annapurna SIL
- Ezra SIL
- Mingzat
- Narnoor
- Padauk
- Padauk Book
- Shimenkan
- Tagmukay
- Tai Heritage Pro

[...]

PR #6510, opened 2023-07-06
Lisu Bosa: Version 2.000 added

emmamarichal:
c2e8f43: [gftools-packager] Lisu Bosa: Version 2.000 added

* Lisu Bosa Version 2.000 taken from the upstream repo https://github.com/silnrsi/font-lisu-bosa at commit https://github.com/silnrsi/font-lisu-bosa/commit/.

 634ef90: Description added

 aa981ef: Metadata updated: primary script added


RosaWagner:
So from Simon's comment in https://github.com/google/fonts/issues/4354#issuecomment-1621929001 we might want to see how the sample text render in dev-sandbox

PR #6522, opened 2023-07-10
ADLaM Display: Version 2.000; ttfautohint (v1.8.4.7-5d5b);gftools[0.9.28] added

aaronbell:
4069951: [gftools-packager] ADLaM Display: Version 2.000; ttfautohint (v1.8.4.7-5d5b);gftools[0.9.28] added

* ADLaM Display Version 2.000; ttfautohint (v1.8.4.7-5d5b);gftools[0.9.28] taken from the upstream repo https://github.com/microsoft/ADLaM-Display at commit https://github.com/microsoft/ADLaM-Display/commit/879176243e9f7161a8aefdab8c36a4a7318ebe15.


aaronbell:
Generally I think this font is looking good, but there's a number of FAILS that I wanted to talk through. 

🔥 FAIL: Check Google Fonts glyph coverage.
- The font is intended primarily for ADLaM language support and therefore does not contain the full GF-latin-core set.

🔥 FAIL: Check license file has good copyright string. 
🔥 FAIL: Copyright notices match canonical pattern in METADATA.pb 
🔥 FAIL: Copyright notices match canonical pattern in fonts 
- I've followed Microsoft's preferred copyright string here. Let me know if you are concerned about this. 

🔥 FAIL "ADLaM Display" contains an abbreviation.
- FAIL is not accurate as this is the language name.

🔥 FAIL: Ensure dotted circle glyph is present and can attach marks.
- AFAIK ADLaM does not require a complex shaper.

chrissimpkins:
> The font is intended primarily for ADLaM language support and therefore does not contain the full GF-latin-core set.

Latin kernel is sufficient here.

> Check license file has good copyright string

We discussed in our pipeline meeting and this exception is approved in both the font file and the metadata configuration file.

eliheuer:
https://googlefonts.github.io/gf-guide/onboarding.html says "no references to languages or writing systems" was there an exception made for this?

PR #6554, opened 2023-07-21
Subtree pull Axis Registry updating `ZROT`

vv-monsalve:
Including `fallback_only` metadata field.


RosaWagner:
unblock #6527 

@eliheuer Since this PR arrived in sandbox and fixed #6527, we can remove it from the project board, and change the block status of #6527 to to_production.

PR #6576, opened 2023-07-28
Remove Sanskrit sample text from Dogra

simoncozens:


RosaWagner:
@simoncozens this PR will also push the blocked version 1.005 in production

simoncozens:
Awesome.

PR #6587, opened 2023-08-02
Update method of installing FontBakery

felipesanches:
FontBakery now (starting at version 0.9.0a1) requires the name of the desired profile to be passed as an extra when installing it. For google/fonts, we run the `Google Fonts` profile, so the installation is done with `pip install --pre fontbakery[googlefonts]`

Without passing extras, only the dependencies needed by the Universal and OpenType profiles (and any vendor-specific profile not requiring additional deps) are installed.


m4rc1e:
Thank you!

PR #6591, opened 2023-08-02
Remove Ubuntu Condensed

djrrb:
Ubuntu Condensed is folded into the width axis of the Ubuntu variable font.


m4rc1e:
Thanks DJR. Unfortunately, the backend process is a touch trickier than just deleting the file. I'm going to keep this open to remind me that we could delist the family if needed.

RosaWagner:
We usually don't delist condensed styles cause they are still in use in Workspace which supports only weight styles

PR #6608, opened 2023-08-09
wixmadefortext: v3.100 added

m4rc1e:
Taken from the upstream repo https://github.com/wix-incubator/wixmadefor at release https://github.com/wix-incubator/wixmadefor/releases/tag/3.100


RosaWagner:
@m4rc1e according to issue #6666, METADATA.pb won't parse

m4rc1e:
The metadata files have a position field which is included in the internal fonts_public.proto files. I've just requested that this field is added to the public fonts_public.proto files.

Issue #6615, opened 2023-08-11
Update Lateef (SIL)


LornaSIL:
**Font Project Git Repo URL:**
https://github.com/silnrsi/font-lateef/releases/tag/v4.200

**Current version:**
4.000

**New available version:**
4.200

**Super short description of the upgrade:**
The glyphset was expanded to support 

- FD42 ARABIC LIGATURE RADI ALLAAHU ANHAA
- FD47 ARABIC LIGATURE ALAYHI AS-SALAAM
- FDCF ARABIC LIGATURE SALAAMUHU ALAYNAA
- FDFA ARABIC LIGATURE SALLALLAHOU ALAYHE WASALLAM
- FDFB ARABIC LIGATURE JALLAJALALOUHOU
- FDFD ARABIC LIGATURE BISMILLAH AR-RAHMAN AR-RAHEEM

**Bugs reported in this repo which this upgrade would fix:**

This issue is still open, but this new font doesn't fix it. That seems to be an issue on your end. https://github.com/google/fonts/issues/1234

eliheuer:
There is classification effort going on now, and I can confirm #1234 is fixed in the new dataset.

LornaSIL:
We just did a quick update adding two more honorifics. Is it possible to use the v4.200 instead of v4.100 or do I need to create a new issue?

eliheuer:
The pull request for v4.200 has been made and should be reviewed soon: https://github.com/google/fonts/pull/6627#issuecomment-1698490379

PR #6627, opened 2023-08-16
Lateef: Version 4.200 added

eliheuer:
016aa52: [gftools-packager] Lateef: Version 4.100 added

* Lateef Version 4.100 taken from the upstream repo https://github.com/silnrsi/font-lateef at commit https://github.com/silnrsi/font-lateef/commit/.


eliheuer:
Updating  to 4.100 from 4.000 as suggested in this issue: https://github.com/google/fonts/issues/6615
Not sure where all the fails came from in this update, looking into it now.

RosaWagner:
@eliheuer we have a special agreement with SIL to ignore the FAILS. 

The only thing we need to check before merging is if the additional glyphs are already included in the subset defined here: https://github.com/googlefonts/glyphsets/blob/main/Lib/glyphsets/encodings/arabic_unique-glyphs.nam, and it the case, so good to merge.

RosaWagner (approved):

RosaWagner:
Although when I linked the issue to the PR I saw the 4.2 was ready, so will wait for you to update the PR before merging.

eliheuer:
I can confirm that all the new codepoints are Arabic ligatures that are in the [arabic_unique-glyphs.nam](https://github.com/googlefonts/glyphsets/blob/main/Lib/glyphsets/encodings/arabic_unique-glyphs.nam) file

eliheuer:
Updated

## Lateef: Version 4.200 added

---
 1d555c6: [gftools-packager] Lateef: Version 4.200 added

* Lateef Version 4.200 taken from the upstream repo https://github.com/silnrsi/font-lateef at commit https://github.com/silnrsi/font-lateef/commit/.

PR #6628, opened 2023-08-16
Repush Noto Sans/Serif Ethiopic

simoncozens:
Noto Ethiopic supports Gurage codepoints (Unicode 14.0), and I added these to the glyphsets, but we haven't repushed these since the glyphset update so the Garage isn't available yet. This should, I think, make it available.


RosaWagner:
added it in th elast version of the list

PR #6631, opened 2023-08-18
Test noto cjk fonts

m4rc1e:


chrissimpkins:
Will close this.  We are looking for the updates in the Noto *Serif* CJK families.

PR #6685, opened 2023-09-08
Kameron: Version 1.100 added

emmamarichal:
0cc0849: [gftools-packager] Kameron: Version 1.100 added

* Kameron Version 1.100 taken from the upstream repo https://github.com/googlefonts/kameronFont at commit https://github.com/googlefonts/kameronFont/commit/4c36d3f42f93806b911562d50927406fbb043f89.

 e3266cc: description and metadata


emmamarichal:
There are some regressions with the Bold v-metrics: in the old one, regular and bold didn't have the same values of v-metrics. Since I've upgraded it to a variable font, I kept the same in Regular and Bold.
Let me know if I should keep the previous one for Bold, I'll change them

Issue #6534, opened 2023-07-17
Update Noto Serif CJK to address the GB18030-2022 revision


chrissimpkins:
The GB18030 specification was updated in the last year and the Noto Serif CJK family will require an update to address the revision.  Adobe is working on the updates now and the production fonts should be ready by approximately August.  Adding the onboarding to our Q3 2023 milestone.

chrissimpkins:
Files are available from Adobe. Ready for QA testing and onboarding.

PR #6695, opened 2023-09-11
Update Noto Serif TC

simoncozens:
See https://github.com/google/fonts/issues/6534.

### DIFF
--- a/ofl/notoseriftc/METADATA.pb
+++ b/ofl/notoseriftc/METADATA.pb
@@ -9,7 +9,7 @@ fonts {
   weight: 400
   filename: "NotoSerifTC[wght].ttf"
   post_script_name: "NotoSerifTC-ExtraLight"
-  full_name: "Noto Serif TC"
+  full_name: "Noto Serif TC ExtraLight"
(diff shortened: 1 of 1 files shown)


ofl/notoseriftc/METADATA.pb
```diff
@@ -10,7 +10,7 @@ fonts {
   filename: "NotoSerifTC[wght].ttf"
   post_script_name: "NotoSerifTC-ExtraLight"
   full_name: "Noto Serif TC"
```
RosaWagner, line 12: @simoncozens The full font name should indeed contain the the style name. I overlooked that before. It is the postcript name that we want the same as before to not disturb Adobe.

m4rc1e:
@simoncozens  I've just cut a new release of diff2 which includes your latest commits. Hopefully these prs should now work.

RosaWagner (approved):

chrissimpkins:
I believe that we need CJK subsetter definition updates to onboard the latest Noto Serif CJK family releases that address the GB18030-2022 revision.  @simoncozens mind looking into this?

PR #6719, opened 2023-09-18
sourcesans 3.052 added

m4rc1e:
Taken from the upstream repo https://github.com/adobe-fonts/source-sans at release https://github.com/adobe-fonts/source-sans/releases/tag/3.052R and run through gftools fix-family with include-source-fixes flag enabled.

@RosaWagner the ps name has changed. Shall I use the older name?


m4rc1e:
Thanks, I've reverted the ps and full name changes.

RosaWagner (approved):

PR #6682, opened 2023-09-08
Gotu: Version 2.320;hotconv 1.0.109;makeotfexe 2.5.65596; ttfautohint (v1.8.1) added

eliheuer:
c7cfb08: [gftools-packager] Gotu: Version 2.320;hotconv 1.0.109;makeotfexe 2.5.65596; ttfautohint (v1.8.1) added

* Gotu Version 2.320;hotconv 1.0.109;makeotfexe 2.5.65596; ttfautohint (v1.8.1) taken from the upstream repo https://github.com/EkType/Gotu at commit https://github.com/EkType/Gotu/commit/.

PR #6725, opened 2023-09-19
Update Gotu description to fix a broken repo link

eliheuer:
See the Gotu description here: https://fonts.google.com/specimen/Gotu/about


RosaWagner:
It would be good to re-QA the font as you started with the previous PR. The [fontbakery report](https://github.com/google/fonts/pull/6682#issuecomment-1711037051) shows that the font needs to be rebuilt with gftools builder to fix the naming issue, `use_typo_metrics` need to be enabled etc.

I see several stuff that would need correction in the original PR (from almost 4 years ago).

- OFL.txt copyright string
- AUTHORS.txt and CONTRIBUTORS.txt that need to be removed

Also re-packaging with packager would allow to create a link with the upstream to update it more easily next time.
I added the original (lost) PR into the GF board, but I gonna block and re-open the issue.

RosaWagner:
Sorry sorry I thought Gotu was not released on GF, so actually merging this to correct the Description; but still would be good to correct what's wrong with the font since we are at it if possible, but then priority is a bit lower since the font is already published.

RosaWagner (approved):

Issue #3682, opened 2021-08-09
Illegal html in description


rsheeter:
Invalid html description for Allison: attribute a.target.

Please don't list Allison for to_*.txt until resolved.

rsheeter:
Great Vibes too

andyjgf:
This is a validation problem on our end, the font is fine. We will re-process shortly....

felipesanches:
@rsheeter, I did not understand exactly what was the problem here. Was it about an anchor's `target` attribute such as in `<a href="some-url" target="_blank">`?

I see a very large number of these in the repo. If that's a problem, then we'll need to fix many description files.

felipesanches:
* **108** of those are **bio.html** files on the designers catalog.
* **131** of them are **DESCRIPTION.en-us.html** files
* And the remaining **4** are ARTICLE.en_us.html files

RosaWagner:
good remark @felipesanches, from experience I can say that:

- target _blank hasn't been _such_ a blocker over the years
- It is a unnecessary addition to html pages in the repo
- therefore we avoid to use it and when we see it in a font we update we try to correct it
- Since it is not visible to the users I personally don't think it is worth the trouble correcting it library-wide (before fixing the more urgent issues)

But I am also curious to know why a a.target was illegal at that moment, and which target exactly?

felipesanches:
An anchor with `target="_blank"` will open the link on a separate tab. And anchors without it will open the link in the same tab. So it is not exactly invisible to users. It makes sense to me that from a user experience point of view we would like to have consistent behavior on all links across the website.

The reason why this technique may be considered "illegal" seems to be related to a stricter conceptual reasoning about the role of HTML versus the role of javascript. HTML should be used as mark-up for content, while controlling behavior would be a task for javascript. For that reason, it seems that some stricter HTML validation tools may complain about the target attribute, because it encodes behaviour, which should not be the role of HTML. If Google Fonts internal systems contain a validation step for this kind of thing, I would expect it to be consistent. The mere fact that there are plenty of these target attributes across the repo, makes me believe any warning about this matter may have been ignored. Or maybe there's no internal check for this in place.

So, apart from these technical minutiae, we really have to decide which behavior we want on the site. Do we want links to be opened in a separate tabs or in the same tab. And based on that decision we should make the change across all HTML files.

PR #6749, opened 2023-09-23
Fix text content of anchors on DESCRIPTION files

felipesanches:
Fix text content of anchors according to the policy described at https://github.com/fonttools/fontbakery/issues/3497#issue-1043902311 and enforced by FontBakery's check **com.google.fonts/check/description/urls**

(https://github.com/fonttools/fontbakery/pull/3499)


eliheuer (approved):
LGTM, I hope this doesn't cause trouble in the sandbox, but lets try it.

RosaWagner:
So I didn't check everything but for example: 

- "Asap VF Beta" is not a family we want to see pushed at all, I believe it should be removed and from the github repo, so we don't want it in a separate PR so it doesn't end up in the sandbox list.
- Fira Go, Goldman, Red Rose have a `target` in the <a> div, we want to remove that (cf https://github.com/google/fonts/issues/3682 that would require more info on why etc so we can make a check)

We should check which family has an upstream.yaml and which not, and review them in separate PRs.

So reverting the PR so we can sort the families before merging.

PR #6762, opened 2023-09-25
Noto Sans Coptic: Version 2.004; ttfautohint (v1.8.4.7-5d5b) added

simoncozens:
578527c: [gftools-packager] Noto Sans Coptic: Version 2.004; ttfautohint (v1.8.4.7-5d5b) added

* Noto Sans Coptic Version 2.004; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/notofonts/coptic.git at commit https://github.com/notofonts/coptic/commit/.


simoncozens:
```
This release corrects the combination U+605 U+102E5 (numbersign + five) (#14)
```

eliheuer (approved):
Looks good to me.

@simoncozens This is not a regression, but there are many Latin glyphs with very bad mark alignment, Ideally this would be fixed before merging, but it not necessary if there is a good reason to ignore it. 

Screenshot 2023-11-15 at 7 47 34 PM
Screenshot 2023-11-15 at 7 47 52 PM
Screenshot 2023-11-15 at 7 48 25 PM

simoncozens:
This is happening because the Coptic diacritics are anchored differently to the Latin ones; when the Latin glyphs are subsetted into the font, they pick up the Coptic diacritics. It's probably impossible to avoid given the way we do the UFO merging. I think it's not really a problem, since precomposed glyphs will still work, and Latin Core doesn't expect to cover stuff like ẅ.

RosaWagner (approved):

PR #6764, opened 2023-09-25
Noto Sans Glagolitic: Version 2.004; ttfautohint (v1.8.4.7-5d5b) added

simoncozens:
843301a: [gftools-packager] Noto Sans Glagolitic: Version 2.003; ttfautohint (v1.8.4.7-5d5b) added

* Noto Sans Glagolitic Version 2.003; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/notofonts/glagolitic.git at commit https://github.com/notofonts/glagolitic/commit/.


simoncozens:
```
This release:

* Adds a dotted circle (#2)
* Rationalizes height/anchors of titlos and macrons (#4)
* Restores some more missing anchors (#5)
```

eliheuer:
@simoncozens The release notes say this update adds a dotted circle, but there is still one dotted circle related FontBakery fail. Should this be fixed, or can it be ignored?

🔥 FAIL: Ensure dotted circle glyph is present and can attach marks. ([com.google.fonts/check/dotted_circle](https://font-bakery.readthedocs.io/en/stable/fontbakery/profiles/%3CSection:%20Shaping%20Checks%3E.html#com.google.fonts/check/dotted_circle))
🔥 FAIL The following glyphs could not be attached to the dotted circle glyph:
acutecomb
gravecomb
tildecomb
u1E000
u1E001
u1E002
56 more.

simoncozens:
This one *is* an issue, as the font does have mark glyphs. I'll check it out.

eliheuer (changes requested):
@simoncozens Looks good to me. The one FontBakery fail just needs to be addressed (should we ignore it, or fix it?), and then this can be merged.

simoncozens:
Updated

## Noto Sans Glagolitic: Version 2.004; ttfautohint (v1.8.4.7-5d5b) added

---
 b8825b1: [gftools-packager] Noto Sans Glagolitic: Version 2.004; ttfautohint (v1.8.4.7-5d5b) added

* Noto Sans Glagolitic Version 2.004; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/notofonts/glagolitic.git at commit https://github.com/notofonts/glagolitic/commit/.

RosaWagner:
@simoncozens tilde is off since previous version (not a regression).

Screenshot 2023-11-07 at 15 42 12

eliheuer (approved):
This looks good to me! 

For the two FontBakery fails:

🔥 FAIL: Shapes languages in all GF glyphsets:
This is an experimental check with known issue and can be ignored

🔥 FAIL: Check that legacy accents aren't used in composite glyphs:
There are other Noto PRs with this fail that have been merged recently, so I assume it is ok, would be nice to have a rationale, but not necessary. For example:
https://github.com/google/fonts/pull/6867
https://github.com/google/fonts/pull/6779

The issues with mark alignment should be fixed, but I'm not saying it is a blocker because the issues existed in the previous font. I would prefer if these issues were fixed before merging though. 

Thanks!

simoncozens:
This won't be fixed easily. Glagolitic uses the combining tilde, and we have anchored it in a certain way; when the Noto Sans glyphs get merged in, the UFO merger sees the existing `tildecomb` and does not replace it, so the Noto Sans subset inherits the glyph (with anchors) from the Glagolitic glyphset. This means a Glagolitic-designed tilde gets anchored onto the Latin glyphs, which doesn't work correctly. I guess the right solution would be to have a new mode for `ufomerge` which brings in existing glyphs with a new name (e.g. `tildecomb.1`) but that has horrible implications. (e.g. any feature code would need to be rewritten). Or we redesign the Glagolitic tilde so that the anchoring is the same as LGC and fix the anchors on the base glyphs. Either way a lot of work for minimal gain.

The companion Latin on these fonts is not intended for general use anyway.

RosaWagner (approved):

PR #6766, opened 2023-09-25
Noto Sans Georgian: Version 2.003 added

simoncozens:
c5cfe73: [gftools-packager] Noto Sans Georgian: Version 2.003 added

* Noto Sans Georgian Version 2.003 taken from the upstream repo https://github.com/notofonts/georgian.git at commit https://github.com/notofonts/georgian/commit/.


eliheuer:
@simoncozens is there a reason why this is failing the `Ensure dotted circle glyph is present and can attach marks.` check?

I noticed that the PR for Noto Serif Georgian: Version 2.003 is not failing this check, so this fail should probably be fixed before merging unless there is a good reason to keep it.

Thanks!

simoncozens:
Bug in the fontbakery check. Georgian doesn't use a complex shaper, it's the same as Latin.

simoncozens:
OK, it's *actually* a bug in Noto Sans, which I'll fix, but the point remains: this font is fine.

RosaWagner (approved):

eliheuer:
@RosaWagner Did you notice that there are issues with "ǎ" and "Ǎ"? Shouldn't this be fixed?
Screenshot 2023-10-26 at 8 30 31 PM

eliheuer:
“ĳ” (U+0133), “Ĳ” (U+0132), and “¯” (U+00AF) also have tofu.

Screenshot 2023-10-26 at 8 46 26 PM

eliheuer (changes requested):
I reviewed this while following the new GF-Guide section on reviewing PRs: https://googlefonts.github.io/gf-guide/onboarder-workflow.html#the-pull-request

There are issues with`ǎ (U+01CE)`, `Ǎ (U+01CD)`, `ĳ (U+0133)`, `Ĳ (U+0132)`, and `¯ (U+00AF)`. I left more detailed information in comments in this PR thread. See `checks > Google Fonts QA > qa`.

Please address these issues and update the PR.

simoncozens:
> ĳ (U+0133), Ĳ (U+0132), and ¯ (U+00AF) also have tofu.

My understanding is that these have been removed from Latin Core.

> I know you have an automated system for keeping to Latin parts of Noto fonts in sync, can you update the upstream Noto Latin font sources to fix this?

The upstream Noto Sans is fine, it's the merging code that's doing something weird. Will try to fix.

RosaWagner:
- A/a caron are not really "expecting" to work
- IJ/ij have been removed from core so expected to be removed from the non-LCG Notos
- Legacy macron is in Latin Core so should be there in the file, and I see the codepoint. The strange thing about it is that it is called "overscore" and is longer than the macroncomb. It would be a good idea to look it up but it looks like a deliberated choice (maybe for a better use of the IPA set in Noto Sans?). I personally don't see a problem with that since it doesn't affect the use of the font.

What I find a disturbing though is this issue with acutecomb that I noticed just now:
Screenshot 2023-10-27 at 12 12 42

I don't find this issue in Sans Chorasmian nor Sans Gujarati. But I find it with macron in Sans Coptic.

Ideally we would like this to get fixed, but I don't think it goes against merging this PR in the sense that the live version is already like that and we are trying to improve the Georgian script (making sure there is no regression in the primary script). It's a bit hard to ignore though now that it is found. @simoncozens what would recommend here? Shall we wait for a next wave of Noto upgrades to fix mark attachment everywhere, or do you want to update this PR first?

simoncozens:
I think the problem here is that there is an acutecomb in the Georgian sources, and it's different to the one in the Noto Sans sources, and since we're merging sources, the Latin components are getting rebuilt with the wrong glyphs.  I *think*. I'll take a look; let's leave this for now, and not merge this PR until I have fixed.

RosaWagner:
Okay so we would apply the same reasoning for Sans Coptic.

@eliheuer @emmamarichal let's remember to take a quick look pay at the latin composite issues we could have missed because there is no regression in itself.

emmamarichal:
@simoncozens did you have the time to take a look to this one?

simoncozens:
Updated

## Noto Sans Georgian: Version 2.004 added

---
 227f3d5: [gftools-packager] Noto Sans Georgian: Version 2.004 added

* Noto Sans Georgian Version 2.004 taken from the upstream repo https://github.com/notofonts/georgian.git at commit https://github.com/notofonts/georgian/commit/.

emmamarichal (approved):

emmamarichal:
Sorry, I approved that PR a bit to soon :) 

I found some interpolation issues here:

Screenshot 2024-02-14 at 10 56 25

Screenshot 2024-02-14 at 10 56 32

simoncozens:
Thank you, will fix. Removing these semibold masters exposes lots of interesting (old) bugs...

simoncozens:
Updated

## Noto Sans Georgian: Version 2.005 added

---
 07eb586: [gftools-packager] Noto Sans Georgian: Version 2.005 added

* Noto Sans Georgian Version 2.005 taken from the upstream repo https://github.com/notofonts/georgian.git at commit https://github.com/notofonts/georgian/commit/.

PR #6778, opened 2023-09-25
Noto Sans Gujarati: Version 2.105 added

simoncozens:
dcbe8bc: [gftools-packager] Noto Sans Gujarati: Version 2.105 added

* Noto Sans Gujarati Version 2.105 taken from the upstream repo https://github.com/notofonts/gujarati.git at commit https://github.com/notofonts/gujarati/commit/.


simoncozens:
```
This release:

* Fixes the position of nukta below other marks (#28)
* Reorders the sequence U+0AFD, U+0AFC (#30)
* Improves the display of the sequence U+0AFF U+0AFA (#31)
* Improves the positioning of ligated marks with repha or anusvara (#33)
```

emmamarichal:
Hi @simoncozens,
It's the first time I review a Noto font, so maybe don't take everything I've found into account. I'm not sure how important this is or isn't, let me know!

_____

- Some glyphs are not displayed anymore:

https://github.com/google/fonts/assets/64773544/4b4b889f-ea5c-4843-8b46-d1a4d2cdf2a2

- I know that the position of this accent is part of the update, but is it normal that it is no longer centered on the circledot?

https://github.com/google/fonts/assets/64773544/a46afe80-a628-419b-ac30-b143e43d6e7d

- There is a interpolation issue on uuvowelsignnuktagujr's dot

Capture d’écran 2023-10-26 à 10 22 35

simoncozens:
> Some glyphs are not displayed anymore:

IJ was removed from the Google Fonts Latin Core set, this is expected. Danda and double danda... oops, they should be there. Good catch.

>  it is no longer centered on the circledot

My interpretation of dotted circles is that they are meant to show a representative example of where the mark will be positioned on a glyph. i.e. the dotted circle stands for "any glyph" and the mark is shown where it would appear. The nukta is usually positioned on the stem, which is on the right of the glyph, so this is fine.

> There is a interpolation issue on uuvowelsignnuktagujr's dot

Nice. I will fix that and the danda, and re-release.

simoncozens:
Updated

## Noto Sans Gujarati: Version 2.106 added

---
 922dd2d: [gftools-packager] Noto Sans Gujarati: Version 2.106 added

* Noto Sans Gujarati Version 2.106 taken from the upstream repo https://github.com/notofonts/gujarati.git at commit https://github.com/notofonts/gujarati/commit/.

simoncozens:
> Check that legacy accents aren't used in composite glyphs

Oh jeepers what now? This is going to affect all Noto until Noto Sans/Serif are "fixed". Does it really matter whether a spacing or a non-spacing accent glyph is used to make a composite?

RosaWagner:
> Oh jeepers what now? This is going to affect all Noto until Noto Sans/Serif are "fixed". Does it really matter whether a spacing or a non-spacing accent glyph is used to make a composite?

It does, yes. It is the only way we can insure the subsetter would display the same accented letter wether it calls it by the codepoint or by combining the base letter and the accent. 

I would obviously ignore the fail though for the Noto collection because, as you said, Noto Sans/Serif need to be fixed first and the fonts are already onboarded like that, so let's just move on and this will be for the next wave of upgrades. Tip: if the composites seem okay in the source, it means this is caused by the un-nesting of components.

emmamarichal (approved):

RosaWagner (approved):

PR #6786, opened 2023-09-26
Noto Sans Tangsa: Version 1.506 added

simoncozens:
99e3658: [gftools-packager] Noto Sans Tangsa: Version 1.506 added

* Noto Sans Tangsa Version 1.506 taken from the upstream repo https://github.com/notofonts/tangsa.git at commit https://github.com/notofonts/tangsa/commit/.


simoncozens:
Tangsa doesn't have marks, so I'm not worried about the dotted circle.

```
This release improves the shape of the final NG after community feedback (#9, thanks to David Williams).
```

m4rc1e:
I don't know this writing system so I cannot comment on whether the changes are correct:

old:

Screenshot 2023-10-10 at 12 13 20


new:

Screenshot 2023-10-10 at 12 13 25

simoncozens:
These are all correct; see https://github.com/notofonts/tangsa/issues/3

m4rc1e:
Thanks for the confirmation

PR #6789, opened 2023-09-26
Noto Sans Thai: Version 2.002 added

simoncozens:
64fcce1: [gftools-packager] Noto Sans Thai: Version 2.002 added

* Noto Sans Thai Version 2.002 taken from the upstream repo https://github.com/notofonts/thai.git at commit https://github.com/notofonts/thai/commit/.


simoncozens:
```
This release:

* Fixes the reordering of nikahit after an above mark (#3)
* Uses the alternate form of loChula when there are both above and below marks (#7)
* Improves the display of COMBINING MACRON BELOW with various bases (#6)
* Correctly orders COMBINING MACRON BELOW with another mark below (#5)
* Adds a stylistic set for tailless tho and yo (#9)
```

m4rc1e (approved):

Issue #3103, opened 2021-02-19
Fix "Multiple Designers" 


RosaWagner:
These are the fonts where we need to add the actual designers in METADATA.pb.
I suggest to separate it between @vv-monsalve, @yanone and me, and we take care of it when possible.

Shall we give one issue per font ?

Roro
AmaticaSC (deprecated)
- [x] Amita (#3206)
- [x] ArimaMadurai (#3207)
- [x] Arya (#3208)
Bhavuka (unreleased)
- [x] Chathura (#3209)
- [x] ConcertOne (#3253)
- [x] Convergence (#3227)
- [x] Ewert (#3228)
- [x] Galada (#3229)

Vivi
- [x] Homenaje (#3247)
- [x] Lekton (#3226)
- [x] MeieScript (#3239)
- [x] Oranienbaum (#3240)
- [x] Pattaya (#3241)
- [x] Pavanam (#3217)
- [x] PirataOne (#3242)
- [x] ProstoOne (#3243)
- [x] RalewayDots (#3244)
- [x] Raleway (#3245)
- [x] Revalia (#3246)

Yaya
- [x] Ruluko
- [x] SedgwickAveDisplay
- [x] SedgwickAve
Sniglet (Had proper designer credit, did nothing)
- [x] StalinistOne
StalinOne (Missing from live library, maybe duplicate of Stalinist One?)
- [x] TitilliumWeb
- [x] Underdog
- [x] Voces

vv-monsalve:
> Shall we give one issue per font ?

I would say yes! particularly after our conversation about process and tasks yesterday.

On the other hand, I could take the Spanish speakers designers that could be there. E.g. Ruluko is one of them.

RosaWagner:
Ok, I'll ask @davelab6 his batch script for issues.

Feel free to rearrange the list :)
With a little bit of luck, some font won't need to contact anyone cause the designer is already in the catalog.

I already see some potential conflict with Titlium for example, which has an open PR for ages.

yanone:
One issue per onboarder

RosaWagner:
It is the same case as the Baskervville of the students of the ANRT. Principal design could be, as said, the Institution (ISIA Urbino) and the description could give more info, something like « ... was initiated by the students of ISIA Urbino of the year 20.., (under the direction of Luciano Perondi?). » then, they can drop all the names in their GitHub repo.

vv-monsalve:
[...]

PR #6805, opened 2023-09-28
Fix URLs for description files with associated upstream yaml files only

eliheuer:
Fixes issues from PR 6749: https://github.com/google/fonts/pull/6749

Specifically this comment here: https://github.com/google/fonts/pull/6749#issuecomment-1737670711

Only fonts with associated upstream.yaml files have been updated, all other fonts from PR #6749 will be updated separately


eliheuer:
Here is a breakdown of what I found while looking at what fonts have upstream.yaml files, yes+target means it had a HTML target attribute, and the attribute was removed in my PR:
```
font                    has yaml
--------------------------------
ofl/adlamdisplay             yes
ofl/asapvfbeta               no
ofl/balsamiqsans             yes+target
ofl/bebasneue                no
ofl/besley                   yes
ofl/dongle                   yes
ofl/dosis                    yes
ofl/firacode                 no
ofl/geologica                yes
ofl/goldman                  no
ofl/gowundodum               yes
ofl/hahmlet                  yes
ofl/kaiseidecol              yes
ofl/kaiseiharunoumi          yes
ofl/kaiseiopti               yes
ofl/kaiseitokumin            yes
ofl/kumbhsans                yes+target
ofl/mochiypopone             yes
ofl/mochiypoppone            yes
ofl/mplus1                   yes
ofl/mplus1code               yes
ofl/mplus2                   yes
ofl/mpluscodelatin           yes
ofl/orbitron                 no
ofl/overpass                 yes
ofl/overpassmono             yes
ofl/padauk                   yes
ofl/palettemosaic            yes
ofl/paytoneone               yes
ofl/podkovavfbeta            no
ofl/redrose                  yes+target
ofl/rowdies                  no
ofl/shipporiantique          yes
ofl/shipporiantiqueb1        yes
ofl/shipporimincho           yes
ofl/shipporiminchob1         yes
ofl/varta                    no
ofl/xanhmono                 no
ofl/zenmarugothic            yes

```

felipesanches:
It is not clear to me what's the problem with target attributes. Please see https://github.com/google/fonts/issues/3682#issuecomment-1737805306

There are hundreds of HTML files with the target attribute on the google/fonts repo, and I can see many of these anchors with `target="_blank"` are live on the families' About pages. So, if this will change, there should be a better explanation for why this is being done. And then the change should be made consistently on the entire collection, which will mean hundreds of HTML page updates. These target attributes are also present in the designers catalog `bio.html` files.

RosaWagner:
@felipesanches I personally don't think that the target is such a problem, but we had a discussion last year that it was unnecessary because the website was already set up to open links in a new tab or because it can be a personal browser set up. So I see it as an aesthetic change and since we are doing right now here we may as well get that at the same time. But good to ask Rod why it was a blocker at that moment and not in other moments.

RosaWagner:
@eliheuer you can make an issue listing all the families to keep track if it helps, eg https://github.com/google/fonts/issues/3103

RosaWagner (approved):

PR #6827, opened 2023-10-02
Add primary_script to Murecho

chrissimpkins:
Add a primary_script definition to Murecho as recommended by @nathan-williams to address the front end language


RosaWagner:
In other fonts supporting only kana we also have specified languages:

```
primary_script: "Hira"
languages: "ja_Kana"
languages: "ja_Hira"
```

I want to merge this to see if only adding the primary script is enough or if we also need the languages. I will amend if so.

PR #6843, opened 2023-10-09
Onest metadatas and article

emmamarichal:
I've updated the order of authors (it's quite urgent)
For the images in the article, I followed Lexend's example. For the size, the images are quite small and in 72ppp. For now, I'll leave them like that to try, but I can remove them if we're not sure.


RosaWagner:
@emmamarichal Can you make 2 separate PRs for the metadata and article? I sense that the article thing is going to take us few PRs to get it right.

emmamarichal:
@RosaWagner I removed the article folder, I'll create a new PR for it when specifications will be decided

PR #6859, opened 2023-10-13
Noto Sans Arabic: Version 2.011 added

simoncozens:
3920071: [gftools-packager] Noto Sans Arabic: Version 2.011 added

* Noto Sans Arabic Version 2.011 taken from the upstream repo https://github.com/notofonts/arabic.git at commit https://github.com/notofonts/arabic/commit/.


simoncozens:
This release:

* Brings codepoint support up to Unicode 14
* Includes extensive re-engineering to fix bugs (particularly in minority languages), reduce glyphset size, and increase consistency with other Noto Arabic families

khaledhosny:
> Unicode 14

Unicode 15 (there were 3 new Arabic code points added in Unicode 15 that are npw supported as well)
https://www.unicode.org/charts/PDF/Unicode-15.0/

eliheuer:
@simoncozens There is one FontBakery vertical metrics fail, is there a good reason to ignore this? Or should it be fixed and the PR updated before merging?

🔥 FAIL: Checking OS/2 usWinAscent & usWinDescent. ([com.google.fonts/check/family/win_ascent_and_descent](https://font-bakery.readthedocs.io/en/stable/fontbakery/profiles/universal.html#com.google.fonts/check/family/win_ascent_and_descent))
🔥 FAIL OS/2.usWinAscent value should be equal or greater than 1431, but got 1374 instead [code: ascent]

I looks around in previous PRs, but couldn't find any prior discussion of this.

eliheuer:
One thing to note for anyone else reviewing this, the change to `ARABIC DISPUTED END OF AYAH` `U+08E2` looks legit to me after some research: https://www.unicode.org/L2/L2014/14095-quranic-marks.pdf

eliheuer:
There is a regression with Arabic Letter Reh with Stroke `U+075B`

Before:
Screenshot 2023-10-30 at 6 59 14 PM
After:
Screenshot 2023-10-30 at 6 59 20 PM

eliheuer:
@simoncozens the Heh`U+0647` ه has been changed to look like an initial Heh ﻫ. I know this is done sometimes for legibility, on my Macbook Arabic keyboard they do this for example, but what is the rational for this?

I have basic Arabic reading skills, and I would find this strange and would expect the circular form without the middle loop when typing `U+0647`

Before:
Screenshot 2023-10-30 at 8 26 38 PM
After:
Screenshot 2023-10-30 at 8 26 44 PM

simoncozens:
> Edit: there is a PR fixing this open : https://github.com/notofonts/arabic/pull/207

No, that PR fixes something different - FEE9, rather than 0647.

Khaled's explanation for choice to use a double-eyed form for ہ is [here](https://github.com/notofonts/arabic/pull/199#issuecomment-1705598183):

>  So heh has a different form for final unconnected (the old isolated form) and really isolated (the new one). The single-eye form should only be used after another Arabic letter, and the double-eye is used otherwise (e.g. when heh is shown alone in alphabet charts, or after numbers for Hijri dates, in abbreviations, etc.). This is because the single-eye form looks very close to Arabic Indic 5.

However: the Unicode Standard and code charts use the single-eyed form as "Xn" (isolated):

Screenshot 2023-10-31 at 07 42 15

Which means we seem to have got ahead of Unicode here. @khaledhosny, I think since Noto is an implementation of Unicode, we may need to back out this change.

khaledhosny:
The [Unicode chart](https://unicode.org/charts/PDF/U0600.pdf) uses this form:
image

and the Arabic chapter has this text (though the text is rather misleading, as it is not the initial form but a form very close to it and the initial form is used as a fallback by users):
image

So there is awareness that isolated heh takes a different form I don’t know why this was not reflected in the table, but this probably would be a 4th column, since heh has 4 forms not 3.

eliheuer:
I am in favor of using the double-eyed form isolated heh. It is a legibility improvement, it is rendered this way on the Apple keyboards, and I trust Khaled's expertise. This is not a merge blocker from me, and I personally think it is an improvement.

eliheuer (changes requested):
Almost all the issues I opened in [the upstream git repo](https://github.com/notofonts/arabic/issues) have been closed by [this PR](https://github.com/notofonts/arabic/pull/212) and the one remaining one does not need to be resolved anytime soon: https://github.com/notofonts/arabic/pull/212

There is a commit to the upstream repository fixing the one remaining FontBakery fail: [Update OS/2 usWinAscent for all fonts](https://github.com/notofonts/arabic/pull/212/commits/7f6fb579b598b1fcb0808f5b8fd5c84e1be492c1). @simoncozens can we rebuild the fonts and update this PR with all the new changes? After that this should be ok to merge. Thanks!

eliheuer:
@simoncozens There are two new FontBakery fails, should they be ignored or fixed?

simoncozens:
The experimental fail is just... wrong. The second one we have discussed at length and my position remains that if the designer knows what they are doing, using legacy accent glyphs as components of a precomposed glyph is fine. So yes, please ignore both.

eliheuer:
I reviewed this with Behdad's `fonttools varLib.interpolatable` QA tool: https://github.com/behdad/google-fonts-ofl-interpolatable-results/blob/main/reports/NotoSansArabic%5Bwdth%2Cwght%5D.ttf.pdf and all the issues are fixed.

RosaWagner (approved):

eliheuer (commented):
Looks good!

Issue #6903, opened 2023-10-25
Inclusion of feedback mechanism section


Akshat111111:
#This issue also works for requesting features. 
The feedback_mechanism section includes various elements like URL of feedback form, community form, bug tracking system etc.

**Describe the issue**
This issue is regarding an idea of implementing a feedback section inorder to make it more user friendly and which will also include responsive support.This will make the task easy for the users to easy access and participate in the feedback process.

**To Reproduce**
In order to implement it, various elements should be added like URL of feedback form, community form, bug tracking system.

**Expected behavior**
This will make the application more responsive and user friendly and will give them the chance to participate in the feedback process.

eliheuer:
Thanks for the issue @Akshat111111.

I agree this is something I want to work on improving, the Noto Fonts have something like this already: 
<img width="1777" alt="Screenshot 2023-10-25 at 4 40 57 PM" src="https://github.com/google/fonts/assets/5162664/d3716dc1-ed75-4e20-90ae-3e301f3f6b58">

There are not as many people working on Google Fonts as an outside observer might think, so for now improving the maintenance and responsiveness of this Git repository is the main goal in this direction. 

There is a Google Fonts guide explaining how most of the Google Fonts system works here: https://googlefonts.github.io/gf-guide/

Since a lot of what is Google Fonts is open source, I want to encourage users to read the guide linked above and feel empowered to solve their own problems.

eliheuer:
There is a PR here related to the issue: https://github.com/google/fonts/pull/6904

I'm going to close it, this is something the Google Fonts backend developers need to implement.

Akshat111111:
> Thanks for the issue @Akshat111111.
> 
> I agree this is something I want to work on improving, the Noto Fonts have something like this already: <img alt="Screenshot 2023-10-25 at 4 40 57 PM" width="1777" src="https://user-images.githubusercontent.com/5162664/278172915-d3716dc1-ed75-4e20-90ae-3e301f3f6b58.png">
> 
> There are not as many people working on Google Fonts as an outside observer might think, so for now improving the maintenance and responsiveness of this Git repository is the main goal in this direction.
> 
> There is a Google Fonts guide explaining how most of the Google Fonts system works here: https://googlefonts.github.io/gf-guide/
> 
> Since a lot of what is Google Fonts is open source, I want to encourage users to read the guide linked above and feel empowered to solve their own problems.

[...]

PR #6904, opened 2023-10-25
Inclusion of feedback mechanism section

Akshat111111:
This is regarding an idea of implementing a feedback section inorder to make it more user friendly and which will also include responsive support.This will make the task easy for the users to easy access and participate in the feedback process.This will make the application more responsive and user friendly and will give them the chance to participate in the feedback process.

### DIFF
--- a/apache/syncopate/METADATA.pb
+++ b/apache/syncopate/METADATA.pb
@@ -1,8 +1,22 @@
+
 name: "Syncopate"
 designer: "Astigmatic"
 license: "APACHE2"
 category: "SANS_SERIF"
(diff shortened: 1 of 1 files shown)


eliheuer:
Thanks for the PR, but I'm going to close it, this is something the Google Fonts backend developers need to implement.

I'm leaving the related issue open so this can be discussed more there: https://github.com/google/fonts/issues/6903

Akshat111111:
Yeah, I understand your point of view and let it to be handled by Google Fonts Backend Developers.

PR #6913, opened 2023-10-26
Removed 8 families to not onboard

RosaWagner:
There are not in any servers, only the repo


RosaWagner:
Merging this, @davelab6 please come back to that PR to know which families to archive in googlefonts repo.

PR #6817, opened 2023-09-29
Update OFL/alike*/METADATA.pb

davelab6:
Per designer request

I'll add the other family and review the description html files shortly

PR #6929, opened 2023-11-02
Alike: Version 1.301; ttfautohint (v1.8.4.7-5d5b) added

emmamarichal:
f86569d: [gftools-packager] Alike: Version 1.301; ttfautohint (v1.8.4.7-5d5b) added

* Alike Version 1.301; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/cyrealtype/Alike at commit https://github.com/cyrealtype/Alike/commit/2de9216949bad46e02460fcb9e455be87d75462d.

 d18e688: Update Metadatas

 1a82b39: Deleted FONTLOG

 40e9b72: Updated Description


emmamarichal:
This PR update the whole font, and replace this one: https://github.com/google/fonts/pull/6817

emmamarichal:
@davelab6 the font isn't in the exception fontbakery list for the RFN, but I saw you kept it in the previous PR you made: https://github.com/google/fonts/pull/6817
Can you confirm we keep it?

emmamarichal:
RFN thing is solved, we keep it, as well in Alike Angular. I'll update the exception list on fb.

Just same question as in Alike Angular:
I never know what is the good way to write it, can someone show me?

🔥 FAIL: Ensure METADATA.pb does not use escaped strings. ([com.google.fonts/check/metadata/escaped_strings](https://font-bakery.readthedocs.io/en/stable/fontbakery/profiles/googlefonts.html#com.google.fonts/check/metadata/escaped_strings))

RosaWagner:
@emmamarichal you have an escape string because of the `"` that is significant for the syntax. If you change it to `'` in the copyright string, it should work. Or you can remove the quotes in the copyright string.

vv-monsalve:
@RosaWagner taking [another font with RFN](https://github.com/google/fonts/blob/main/ofl/akatab/METADATA.pb) as a reference on the syntax, I see the `"` are actually used. Could you explain or show how it should/could be written?

RosaWagner:
@vv-monsalve I usually ignore this FAIL because here `"` needs escape strings in order to preserve the correct file syntax. I believe this check was made to ensure Designer names were written with the proper accented letters without escaped strings. But since strings are defined between 2 `"`, we must use escaped string if the string includes `"`. And I don't see a problem with that so I usually ignore the FAIL.

To not have this fail you can:
- not include the RFN mention and leave it only in the license file — not recommended
- change `"` to `'` so the metadata.pb doesn't interpret the beginning of the RFN as the end of the string
- not include quotes in the string

vv-monsalve:
In light of what you say, I agree we could ignore the Fail. However, to have the whole reasoning on this, from your options, it could be either one of the following options:

> - change " to ' so the metadata.pb doesn't interpret the beginning of the RFN as the end of the string

`"Copyright 2009 The Alike Project Authors (https://github.com/cyrealtype/Alike) with Reserved Font Name \'Alike\' and \'Alike Angular\'."`

> - not include quotes in the string

`"Copyright 2009 The Alike Project Authors (https://github.com/cyrealtype/Alike) with Reserved Font Name Alike and Alike Angular."`

vv-monsalve:
Alternatively, we just ignore it since the check will be updated.

vv-monsalve (approved):

RosaWagner (approved):

PR #6932, opened 2023-11-02
Narnoor: Version 3.000 added

emmamarichal:
9baa8c6: [gftools-packager] Narnoor: Version 3.000 added

* Narnoor Version 3.000 taken from the upstream repo https://github.com/silnrsi/font-narnoor at commit https://github.com/silnrsi/font-narnoor/commit/.

 8a9708d: description and metadatas updated


emmamarichal:
Regressions on latin: darker + xheight

https://github.com/google/fonts/assets/64773544/5502d0f7-91e0-4d33-8972-a48631344056

RosaWagner:
@emmamarichal since the latin is not the primary script, and since the font has been served only 81 thousands/week (very low), we can accept this regression. But it would be good to be good to post a gif in comment of the issue to ask SIL to confirm that is what they want.

PR #6951, opened 2023-11-09
Bodoni Moda: Version 2.005 added

emmamarichal:
0f1d0cc: [gftools-packager] Bodoni Moda: Version 2.005 added

* Bodoni Moda Version 2.005 taken from the upstream repo https://github.com/indestructible-type/Bodoni at commit https://github.com/indestructible-type/Bodoni/commit/87fc08d47160f9eb7802960b5f7a641576803700.

 6c1b8b5: metadatas and description


emmamarichal:
I automatically have the opsz in the name ID1 when I export, we usually ignore it.

`🔥 FAIL METADATA.pb font.name field ("Bodoni Moda") does not match correct font name format ("Bodoni Moda 6pt"). [code: mismatch]`

I'll make the modification for the copyright thing. I also open a issue on fontbakery since I haven't been notified for this fail: https://github.com/fonttools/fontbakery/issues/4329

RosaWagner:
@emmamarichal since the default point size was elided before, we would need it elided now to avoid regressions in the postscript name etc. Also METADATA.pb declare 11pt as default, so this size has to be the origin. The STAT table needs to be the same as before etc.

emmamarichal:
@RosaWagner I see, but there is no master at 11pt, so the origin is automatically 6pt or 96pt. Concerning the default point size elided, I don't know how to do it? In the last released fonts with opsz, the point size is automatically in the font name (I think we already have this discussion with Marc but I can't find in which PR). It was the case for DMSans or Nunito for example. 

I can try to remove it and set up the origin to 11pt again, but let me know if you have already had these kind of problem :)
I think last time, for Nunito, I had to add a master to fix the origin issue.

emmamarichal:
At least I should add something like this:

registry_default_overrides {
  key: "opsz"
  value: 12
}

RosaWagner:
The process is different from Nunito or DM Sans because it was already onboarded with an opsz axis, before we had the rule of not eliding the default size. And we don't want regression. 

This is the set up of the published version of Bodoni Moda, and we need to keep exactly that:
Screenshot 2023-11-17 at 11 52 52

This was the axis mapping It needs to stay exactly that:
```
    <axis tag="wght" name="Weight" minimum="400" maximum="900" default="400">
      <map input="400" output="400"/>
      <map input="500" output="470"/>
      <map input="600" output="550"/>
      <map input="700" output="650"/>
      <map input="800" output="766"/>
      <map input="900" output="900"/>
    </axis>
    <axis tag="opsz" name="Optical size" minimum="6" maximum="96" default="11">
      <map input="6" output="0"/>
      <map input="11" output="4880"/>
      <map input="16" output="6653"/>
      <map input="24" output="7953"/>
      <map input="36" output="9265"/>
      <map input="48" output="9535"/>
      <map input="72" output="9805"/>
      <map input="96" output="10000"/>
    </axis>
  </axes>
```

emmamarichal:
@RosaWagner 
I generated a master in between, to solve the name issue, and it works now. I also updated the values of masters and instances, and added an axis mapping. The avar table and all the fonts infos look ok in Samsa, but there is still a difference in term of design I can't find why. 
I'll continue to investigate and test it this week, but if you have any idea, it's welcomed!

**NEW**

Screenshot 2023-12-06 at 12 29 45

**OLD**

Screenshot 2023-12-06 at 12 29 53

RosaWagner:
@emmamarichal did you generate the 11pt master before or after making the axis mapping?

emmamarichal:
OH! Yes it was that, thank you, you're the best 💪
I'll update the PR soon!

emmamarichal:
wait @m4rc1e, it's not ready

emmamarichal:
Updated

## Bodoni Moda: Version 2.005 added

---
 634558f: [gftools-packager] Bodoni Moda: Version 2.005 added

* Bodoni Moda Version 2.005 taken from the upstream repo https://github.com/indestructible-type/Bodoni at commit https://github.com/indestructible-type/Bodoni/commit/30ce6cdc354ef179a3b72ba0f0e71826e599348c.

 6c5d387: OFL and metadatas updated

PR #6976, opened 2023-11-18
Vernon Adams: updated biography

davelab6:


RosaWagner:
@davelab6 we have a workflow for updating the lists, any manual changes disrupt that workflow. Please don't update the list manually.

RosaWagner (approved):

PR #6982, opened 2023-11-20
Add Kalnia minisite URL

chrissimpkins:
cc @vv-monsalve

### DIFF
--- a/ofl/kalnia/METADATA.pb
+++ b/ofl/kalnia/METADATA.pb
@@ -32,3 +32,4 @@ source {
 }
 stroke: "SERIF"
 classifications: "DISPLAY"
+minisite_url: "http://www.fridamedrano.com/kalnia.html"


vv-monsalve:
The current microsite for Kalnia still needs to be fully developed and is only a front page with an animation and two really short phrases about the font. In its current state, it is not a useful resource. 

I think it would be better to include it once it is fully developed (if); otherwise, it would lead more to "user frustration" from a user experience pov.

RosaWagner:
so… closing?

vv-monsalve:
In a private chat conversation, we agreed on the following:

> add a couple of images to the About section. They are more illustrative about the font than the current page.
> If the designer further develops the info in the webpage for the Kalnia families once Kalnia Glaze, the colored version, is ready, we could add the enriched minisite later once we publish the latter. 

I'll add the images to the About section and update it in a new PR.

PR #7001, opened 2023-11-22
Playpen Sans About: deleting the table

vv-monsalve:
Another take on the PPS about section.


vv-monsalve:
Notes:
- This PR deleted the "illegal" table from the Article but is still not showing up on the Dev server. 
- I'm adding it to the traffic jam under the blocked list so we know there is an issue with it.
- I will still need to update the About section after the text receives the copy editing revision and with adjusted images.

PR #6919, opened 2023-10-26
Sixtyfour: Version 2.000;gftools[0.9.33] added

vv-monsalve:
d82ab2b: [gftools-packager] Sixtyfour: Version 2.000;gftools[0.9.33] added

* Sixtyfour Version 2.000;gftools[0.9.33] taken from the upstream repo https://github.com/jenskutilek/homecomputer-fonts at commit https://github.com/jenskutilek/homecomputer-fonts/commit/9b182d8ab290d98477ddc0a4f84c8137d540994b.

PR #7025, opened 2023-11-28
Sixtyfour: Version 2.001 added

vv-monsalve:
e0fab0e: [gftools-packager] Sixtyfour: Version 2.001 added

* Sixtyfour Version 2.001 taken from the upstream repo https://github.com/jenskutilek/homecomputer-fonts at commit https://github.com/jenskutilek/homecomputer-fonts/commit/09b2f79d686453d3bc160d04a687488e66962c43.


vv-monsalve:
@RosaWagner This PR updates Sixtyfour, following what was described in [#6919-comment ](https://github.com/google/fonts/pull/6919#issuecomment-1825966537)

RosaWagner:
🔥 FAIL: Validates that when an instance record is included for the default instance, its subfamilyNameID value is set to a name ID whose string is equal to the string of either name ID 2 or 17, and its postScriptNameID value is set to a name ID whose string is equal to the string of name ID 6. ([com.adobe.fonts/check/varfont/valid_default_instance_nameids](https://font-bakery.readthedocs.io/en/stable/fontbakery/profiles/fvar.html#com.adobe.fonts/check/varfont/valid_default_instance_nameids))

I don't think it is necessary to solve this check but it highlights a small conflict in the display of the style name of the default instance, so just as info:

- In Fontbook (Mac): "Regular"
- In TextEdit (Mac): "Regular"
- In Indesign (Mac): "Close"
- In Words (Mac): "Close" (2 times)
- In font dir (Windows): "Close"
- In Words (Windows): "Close"

So Adobe and Windows are both displaying the font (family name + style name) as we expect, but native apps from Mac are not. Which I don't think is the end of the world tbh; no issue were reported with Kablammo so far. A way to solve this would be to have the Default instance called "Close" instead of Regular but I have no idea if the API would actually like it. Maybe that is a conversation to have with the eng team next year for future projects without weight axes.

RosaWagner (approved):

vv-monsalve:
Thanks for the detailed note. Yes, we'll need to find a final resolution for fonts with only custom axes. Eventually, if following the last option (to which I agree) we'll need to adjust the build & test tools to accept it.

PR #7006, opened 2023-11-23
Playpen Sans: Version 1.002 added

vv-monsalve:
0fcca06: [gftools-packager] Playpen Sans: Version 1.002 added

* Playpen Sans Version 1.002 taken from the upstream repo https://github.com/TypeTogether/Playpen-Sans at commit https://github.com/TypeTogether/Playpen-Sans/commit/5f1828480be5843503a4f55e8b6094637370d795.

PR #7032, opened 2023-11-29
Emoji subset added to PPS metadata

vv-monsalve:
Changing subset `symbols` to `emoji` to #7006.

@RosaWagner The PR updating the font was merged this morning. In cases like these small quick fixes, should we add the `METADATA` label to this PR in any case?

Edit: Not adding labels to only track the changes through the font's PR linked above.

### DIFF
--- a/ofl/playpensans/METADATA.pb
+++ b/ofl/playpensans/METADATA.pb
@@ -12,11 +12,11 @@ fonts {
   full_name: "Playpen Sans Regular"
   copyright: "Copyright 2023 The Playpen Sans Project Authors (https://github.com/TypeTogether/Playpen-Sans)"
 }
+subsets: "emoji"
 subsets: "latin"
 subsets: "latin-ext"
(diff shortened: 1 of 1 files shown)


RosaWagner:
@vv-monsalve Can't we include symbols and emoji? I am thinking of the arrows or other codepoints that could be related to symbols

RosaWagner:
(I usually add the proper label to know what this PR is about, and then remove from the project board manually if another PR is already tracked)

vv-monsalve:
I accidentally merged the metadata PR before finishing reading your comments. I had the doubt about deleting "Symbols". I'll reinstate it in another PR.

vv-monsalve:
Actually, checking at the font again the codepoints that would require "symbols" are the same ones that would be covered by "emoji".

eliheuer:
This PR is being tracked through production here: https://github.com/google/fonts/pull/7041

PR #5609, opened 2022-11-24
Noto Serif NP Hmong: version 1.001 added (#5257)

RosaWagner:
Following up on #5257 and created new directory for NP Hmong. 

We now have two families with the same displayed name: "Noto Serif Nyiakeng Puachue Hmong".

Former with font name: `Noto Serif Nyiakeng Puachue Hmong` :::: is to de-list
New one with font name: `Noto Serif NP Hmong`

We expect 2 URLs:
[Noto+Serif+Nyiakeng+Puachue+Hmong](https://fonts.google.com/noto/specimen/Noto+Serif+Nyiakeng+Puachue+Hmong) ::: to de-list
[Noto+Serif+NP+Hmong](https://fonts.google.com/noto/specimen/Noto+Serif+NP+Hmong)

PR #5308, opened 2022-09-24
Noto Sans Phags-Pa: Version 2.003; ttfautohint (v1.8.4.7-5d5b) added

simoncozens:
e469712: [gftools-packager] Noto Sans Phags-Pa: Version 2.003; ttfautohint (v1.8.4.7-5d5b) added

* Noto Sans Phags-Pa Version 2.003; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/notofonts/phags-pa.git at commit https://github.com/notofonts/phags-pa/commit/.

PR #7049, opened 2023-12-01
Noto Sans PhagsPa: Version 2.004; ttfautohint (v1.8.4.7-5d5b) added

simoncozens:
dc92351: [gftools-packager] Noto Sans PhagsPa: Version 2.004; ttfautohint (v1.8.4.7-5d5b) added

* Noto Sans PhagsPa Version 2.004; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/notofonts/phags-pa.git at commit https://github.com/notofonts/phags-pa/commit/.


ofl/notosansphagspa/METADATA.pb
```diff
@@ -1,20 +1,29 @@
...
+subsets: "math"
 subsets: "menu"
 subsets: "phags-pa"
+subsets: "symbols"
+source {
+  repository_url: "https://github.com/notofonts/phags-pa.git"
+  archive_url: "https://github.com/notofonts/phags-pa/releases/download/NotoSansPhagsPa-v2.004/NotoSansPhagsPa-v2.004.zip"
+}
 is_noto: true
+display_name: "Noto Sans Phags-pa"
```
RosaWagner, line 26: @simoncozens I added the display name. Wikipedia seems to not capitilised "Pa", can you confirm it looks alright to you?

RosaWagner:
@vv-monsalve @emmamarichal @eliheuer @m4rc1e there is an important naming situation here that might create confusion in the servers. We had a font which `name ID 1` (PhagsPa) is differing from the Metadata's `font_name` key (Phags Pa). Probably because it was made manually or something else, I don't know. Cf comment in previous PR: https://github.com/google/fonts/pull/5308#discussion_r1238456859

It is a bit similar as this family which changed name https://github.com/google/fonts/pull/5609. Previous PR was introducing a new name and a new directory (`phags-pa`). But here, thanks to the new PR, we are interacting in the same directory as before, so we are not creating a new one. The 2 correct orthographies for this script seem to be "Phagspa" and "Phags-pa", with a preference for "Phags-pa", so I added the latter into METADATA.pb. So it also make more sense to have the `font_name` key without a space. We can't really have a font file name such as "NotoSansPhags-Pa" because it would introduce a hyphen on the family side of the path, which we want to avoid (the hyphen is used to separate the family name from the style name). 

Since the `font_name` key is changing, the URL of the specimen page might change too: right now it is `Noto+Sans+Phags+Pa`, and it will become `Noto+Sans+PhagsPa`. So we might need to delist the previous version `Phags Pa` (with a space). **Therefore we need to be very clear with the eng team about what is happening so we don't have the same 2 families in production.**

So here the new PR allows to:
- update the same font file by keeping the same `name ID1` and `file name` as previous version
- correct the inconsistency between the metadata `font_name`, and the font `name ID 1`
- Add a `display_name` key to make sure it appears correctly to the users on the API.

We can check in dev-sandbox once merged if it creates 2 families or 2 URLs as I suspect. If yes we need to inform the eng team about the trickiness of the situation. Either they delist the additional URL, either they ask us to revert the change in the `font_name` key (and therefore keep the inconcistency). If no double stuff, then no problem.

ofl/notosansphagspa/METADATA.pb
simoncozens: Phags-pa is good.

RosaWagner:
The family looks good in dev, and I see only one family appearing. Although the 2 URLs exist so the old one, so Noto+Sans+Phags+Pa needs to be delisted.

davelab6:
Also Nko had a similar issue, iir

vv-monsalve:
> the old one, so Noto+Sans+Phags+Pa needs to be delisted.

This is included in the delist list. We are waiting for the delist to land to unblock this PR.

vv-monsalve:
@nathan-williams This is the other case where we need the delist to happen to unblock this landing into production :)

`Noto Sans Phags Pa` was added to the [delist.txt](https://github.com/google/fonts/blob/417578b383e3384c5cc682f235df38fae013005f/to_delist.txt#L29C1-L29C58) some moths ago.

Dev server
Screenshot 2024-11-13 at 16 25 02

vv-monsalve:
Bringing @nathan-williams chat answer here:

> we can delist Noto Sans PhagsPa in favor of Noto Sans Phags-pa once the latter makes it to production.

So, I'm unblocking this PR now to allow this font to reach prod.

PR #7060, opened 2023-12-04
Update kalnia/article/ARTICLE.en_us.html and DESC

davelab6:
Keep typical description text in the DESC file and leave images in the ARTICLE file


RosaWagner:
@davelab6 we actually talked about it in the chat and I transfer those notes in the Guide: https://googlefonts.github.io/gf-guide/description.html#the-article

The eng team told us the description and the article complete themselves only for the Noto collection. For all other families, the article overwrite the description. So we need all the text in the article. And to avoid duplicated info, we empty the description file. If we delete the description file, it simply breaks the dev sandbox.

RosaWagner:
So except if you have new information I don't know about, I think we should close that PR without merging.

davelab6:
There was an internal issue with having an empty DESC but I think that is resolved another way so yes this can be closed :)

PR #7065, opened 2023-12-05
Update to-lists for dec 5th

eliheuer:


RosaWagner (approved):

RosaWagner:
@chrissimpkins since it was confirmed Kalnia, Afacad and Rethink Sans were not moved to prod, I'd like to merge the lists so Vivi can update them with the new elements. The new fonts in prod list shouldn't move if you want to check in Workspace.

vv-monsalve:
Bringing here a note from Rod in the chat:

> The following are pending and don't need to be added back to the list: 
Family data {YouTube Icons Outlined, Kalnia, Material Symbols Outlined, Material Symbols Rounded, Google Symbols, Noto Color Emoji, Rethink Sans, Afacad, Material Symbols Sharp}

> World won't end if they *are* added

PR #7080, opened 2023-12-08
Delist : added Phag+Pa URL

RosaWagner:
Added the URL to delist and some indications in muted comment.
Eng team confirmed it was low priority so no need to follow up with it until after the holidays.


vv-monsalve:
Excellent! Thank you.

PR #7087, opened 2023-12-11
Noto Znamenny Musical Notation: Version 1.003; ttfautohint (v1.8.4.7-5d5b) added

simoncozens:
72df96a: [gftools-packager] Noto Znamenny Musical Notation: Version 1.003; ttfautohint (v1.8.4.7-5d5b) added

* Noto Znamenny Musical Notation Version 1.003; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/notofonts/znamenny.git at commit https://github.com/notofonts/znamenny/commit/.


simoncozens:
So Felipe and I have been looking at check overrides in the configuration file. This is a useful check for designers to take notice of when they're putting their fonts together, as it enforces good practice. But once we've got a binary file, as you say, who cares. It should really be a source-level check, TBH.

But anyway, we should use a google/fonts-wide fontbakery config file to demote this to a WARN.

m4rc1e:
SGTM thank you. PR LGTM so merging,

m4rc1e:
@simoncozens thoughts on adding a sample text to this family? The same page on dev looks a bit uninspiring.

Metadata files have a `sample_text` field which we could use.

simoncozens:
Borys provided some samples in https://github.com/notofonts/znamenny/blob/main/documentation/test.html - I'll put something together in a separate PR.

m4rc1e:
Thank you!

PR #7111, opened 2023-12-14
Update upstreams [7/7]

simoncozens:
Add upstream repo information to METADATA.pb

### DIFF
--- a/ofl/tirodevanagarihindi/METADATA.pb
+++ b/ofl/tirodevanagarihindi/METADATA.pb
@@ -25,4 +25,7 @@ subsets: "devanagari"
 subsets: "latin"
 subsets: "latin-ext"
 subsets: "menu"
+source {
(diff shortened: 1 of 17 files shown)


RosaWagner:
@vv-monsalve when you review these 7 PRs, it should be confirmed that these changes don't need to go to sandbox/production, and therefore can be removed from the project board.

vv-monsalve:
Thanks, @RosaWagner.
What (or how) should we do with the reported fails that can't be ignored for one of the projects included in a multi ofl PR like this one? Should we a). First resolve the Fail before merging this PR? or b) we can Merge this here and then report and solve the fails?

RosaWagner:
Since it won't be pushed to sandbox, I think you can merge (except if a FATAL appear). I guess if something wrong happen, it will break the dev-sandbox, and the eng team would report it.

PR #7122, opened 2023-12-20
abhayalibre metadata

simoncozens:
- Add upstream repository URL
- Add primary script


m4rc1e:
@emmamarichal we'll add this one to the push lists because a primary script has been added as well.

PR #6773, opened 2023-09-25
Noto Sans Khojki: Version 2.005; ttfautohint (v1.8.4.7-5d5b) added

simoncozens:
78eb315: [gftools-packager] Noto Sans Khojki: Version 2.005; ttfautohint (v1.8.4.7-5d5b) added

* Noto Sans Khojki Version 2.005; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/notofonts/khojki.git at commit https://github.com/notofonts/khojki/commit/.

PR #7144, opened 2024-01-04
Lang v0514

emmamarichal:
unblock #6508 and #6773


emmamarichal:
@m4rc1e First time I'm doing this, I let you check if it's correct and merge it, thanks :)

emmamarichal:
will open a new one with the 0515 version

Issue #7171, opened 2024-01-17
Add Post No Bills?


davelab6:
In #3569 and #3581 `Stick No Bills` family was added, but in the repo there are 2 ofl dirs, [postnobillscolombo](https://github.com/google/fonts/tree/d27d4149a40a9db16637f493f039255f5cb52d1f/ofl/postnobillscolombo) and [postnobillsjaffna](https://github.com/google/fonts/tree/d27d4149a40a9db16637f493f039255f5cb52d1f/ofl/postnobillsjaffna), with Latin + Sinhala and Latin + Tamil, and it seems that Stick No Bills is the former (Sinhala) without the latter (Tamil) combined. 

It was unfortunate that the family was renamed from Post to Stick. 

@pathumego given the possibility to republish with the original name and a single family with 3 scripts, what do you recommend?

(Assigning to myself to track making a decision about this in Q1 :)

AlvesJorge:
@davelab6 shouldn't postnobills be removed from https://github.com/google/fonts/blob/main/tags/all/families.csv ?

PR #7172, opened 2024-01-17
git rm -rf ofl/postnobills* to tidy up the repo

davelab6:
Tidies up the repo; @kalapi noticed these dangling during the tagging effort last month.

Tracking potentially adding these back in #7171

Assigning to @vv-monsalve as on repo maintainer rotation


vv-monsalve:
@davelab6 are these fonts that currently exist on our `google/fonts` repo but never got into the online catalog?

davelab6:
Exactly. We have them in the history so we can remove them to clean things up.

vv-monsalve:
Thanks. So I guess this is a safe PR with no need to delist anything. I'll merge it here then.

PR #7204, opened 2024-01-19
Annapurna SIL: Version 2.000 added

emmamarichal:
06fa574: [gftools-packager] Annapurna SIL: Version 2.000 added

* Annapurna SIL Version 2.000 taken from the upstream repo https://github.com/silnrsi/font-annapurna at commit https://github.com/silnrsi/font-annapurna/commit/.

 41e4b08: description and OFL updated


vv-monsalve:
There is a marks placemente issue.
Screen Shot 2024-01-23 at 18 14 21

emmamarichal:
@vv-monsalve, ah, it's because 'ă' isn't in the sources: the GF Latin Core isn't complete, and even if there is a breve accent (a legacy one), there is no brevecomb accent with proper anchors (all glyphs are decomposed).
But I think it's on the list of the things we can ignore, since here we are interested by the devanagari

jcoblentz:
Thank you for noting that discrepancy. We will work on it for the next version. In the mean time we are happy for you to proceed with this release without the fix.

cc:  @jvgaultney

vv-monsalve:
Thank you, @jcoblentz, for the confirmation.

vv-monsalve:
@emmamarichal There is an extra `>` in the Description file that is visible in the Dev Server.
Since this doesn't break the font, I'll leave the project to move ahead in the pipeline. If you want to update it quickly, you could do it before reaching Prod.  

Screen Shot 2024-01-27 at 20 47 38

PR #7207, opened 2024-01-19
Woowahan Brothers fonts: hotfixed space & nbspace chars

yanone:
I fixed as many things as I could in OTMaster as a hotfix. Try to ignore all other FB FAILs, as these fonts are terribly old. They were missing the space character.


vv-monsalve:
@yanone, please confirm if you added the "bad rendering" label because you discovered the issue after the pull request, or if the current update addresses a bad rendering problem. 
If it is the latter, please note that in PRs, labels like "bad rendering" (or similar) should only be added after the PR reviewing process. The information about the update should be included on the font issue, or any additional details can be provided as a comment here to help with the revision process. ;)

vv-monsalve:
> 🔥 FAIL: METADATA.pb: Check URL on copyright string is the same as in repository_url field. ([com.google.fonts/check/metadata/consistent_repo_urls](https://font-bakery.readthedocs.io/en/stable/fontbakery/profiles/googlefonts.html#com.google.fonts/check/metadata/consistent_repo_urls))
🔥 FAIL Please add a family.source.repository_url entry. [code: lacks-repo-url]

There is an ongoing effort to add the source metadata field to all the projects on the catalog. This would be a perfect time to add it to all these projects.

vv-monsalve:
### Jua
> 🔥 FAIL: METADATA.pb should contain at least "menu" and "latin" subsets.
> 🔥 FAIL: Check METADATA.pb includes production subsets.

More metadata fixes for this font

yanone:
The linked issue has a "glyphset issue" label and I guess I applied the wrong label when manually adding those labels.

yanone:
> There is an ongoing effort to add the source metadata field to all the projects on the catalog. This would be a perfect time to add it to all these projects.

Sure, but there aren't any sources or repos, hence the hotfixes. I checked that.

> More metadata fixes for this font

So, I'm not sure how this works. The fonts actually only have the most minimal set of Latin characters. So I'm not sure if it's even correct to put "latin" there in the metadata. And I don't know how to confirm. I checked the "glyphsets" and new "nam-files" repos and couldn't find any files named "latin" or "menu". Of course I can just put "latin" in the metadata but I'm not sure if that's even correct given the circumstances.

If you know where the definitions for these subsets reside, I would confirm that.

vv-monsalve:
> Sure, but there aren't any sources or repos, hence the hotfixes. I checked that.

Each of the ofl/font directories have a metadata file. E.g. Jua [https://github.com/google/fonts/blob/main/ofl/jua/METADATA.pb](https://github.com/google/fonts/blob/main/ofl/jua/METADATA.pb)

> The fonts actually only have the most minimal set of Latin characters. So I'm not sure if it's even correct to put "latin" there in the metadata

It is related to certain thresholds defined for each script/language. If FB is raising the Fail it is because the range was reached. 

`latin` is actually included in all the fonts even if the primary script is different just from the fact we require Latin Kernell to acompany all the fonts. Finally the description of the font says `Jua is a Korean and Latin font…` 

Screen Shot 2024-01-24 at 15 38 57


The font is currently displaying latin in the Specimen, so the `primary_script: "Kore"` metadata field should also be added.

`menu` is added so the font is used when displaying the font name in WorkSpace (WS) menus.

vv-monsalve:
Following the above, perhaps creating independent PRs for each font would be better. Let's discuss it on the pipeline meeting.

vv-monsalve (approved):

vv-monsalve:
> I’m aware. But we don't have a source to put into that metadata file.

Right. Merging this now.

vv-monsalve:
@yanone Do Hyeon is still showing tofu on Dev server
Screen Shot 2024-01-27 at 20 43 19

emmamarichal:
@vv-monsalve Don't see tofu anymore, I guess it has been fixed thanks to one of the lang updated!
However, we should update the description, there is no punctuation, and no repo link.

yanone:
They don't have sources or repositories, hence the hotfixes on the binaries.

emmamarichal:
alright, np!

PR #7214, opened 2024-01-23
robotoflex ARTICLE html: Remove youtube iFrame, rewrite to merge 2 parts

davelab6:
Remove youtube iFrame, rewrite to merge 2 parts


vv-monsalve:
@m4rc1e, it seems the reported `Error: Process completed with exit code 1.` for Diffenator originated on there is no font included, so it's not finding it. So merging this here.

PR #7215, opened 2024-01-24
Yarndings 12: Version 1.000; ttfautohint (v1.8.4.7-5d5b) added

emmamarichal:
deb409e: [gftools-packager] Yarndings 12: Version 1.000; ttfautohint (v1.8.4.7-5d5b) added

* Yarndings 12 Version 1.000; ttfautohint (v1.8.4.7-5d5b) taken from the upstream repo https://github.com/scfried/soft-type-yarndings at commit https://github.com/scfried/soft-type-yarndings/commit/200cc40a30ffbe590807cdfc8b8e17d0d63c43ff.

 68a5493: Description and metadatas updated


emmamarichal:
@vv-monsalve, In that case, the designer used latin unicodes for symbols, here is her explanation:

> I do think that it is fairly important that the symbols apply to the unicode for the standard keyboard because if a knitter is designing a pattern in Google Docs they will have no idea that there are symbols for them to use otherwise. (The idea is that it works similarly to Wingdings.)

Chris is ok, it will work like Wavefont for example.

vv-monsalve:
What would happen with e.g. font menus in that case?

emmamarichal:
@vv-monsalve, Yes, Chris gave me the example of Wavefont, so I thing we should definitely treat that case like that.
I'll make the change in metadatas after the meeting or tomorrow :)

m4rc1e:
Have you looked into using the unicode dingbat block? https://unicode.org/charts/PDF/U2700.pdf

m4rc1e:
Just checked the webdings fonts on my machine and they're mapped to Latin codepoints so probably ignore what I've posted. We can always add double points later if they're requested.

ofl/yarndings12/DESCRIPTION.en_us.html
```diff
@@ -0,0 +1,3 @@
+<p>The Soft Type Collection is designed for knitters to chart out their typographic projects. Yarndings is an ornamental typeface inspired by a mix of typographic ornamentation, traditional Fair Isle & Nordic knitting patterns, and dingbat fonts from the 1990's.</p>
```
vv-monsalve, line 1: The description file still needs to include the note about numbers in the File name + the link to the charted version.

vv-monsalve (changes requested):

m4rc1e (approved):

PR #7221, opened 2024-01-24
Update families.csv

nyshadhr9:
Remove language related tags and repetitive ones like `/Display/Display`.

### DIFF
--- a/tags/all/families.csv
+++ b/tags/all/families.csv
@@ -2,7 +2,6 @@ Family,Group/Tag,Weight
 ABeeZee,/Flow/Calm,88
 ABeeZee,/Sans/Geometric,10
 ABeeZee,/Sans/Humanist,75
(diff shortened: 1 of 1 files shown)


m4rc1e:
This is intentional. If we look at the spreadsheet, we can see that the parent categories such "Display" do get assigned values for families.

Screenshot 2024-01-25 at 10 09 27

nyshadhr9:
I talked to Evan about this and the other issue with such tags is that there is only a single family tagged with them. For example only ABeeZee has the /Sans/Sans tag. Maybe they were accidentally assigned?

m4rc1e:
> I talked to Evan about this and the other issue with such tags is that there is only a single family tagged with them.

@nyshadhr9 Sorry, there was an error in my tooling. I've just fixed it and created another pr that has all parents tags, https://github.com/google/fonts/pull/7235

PR #7223, opened 2024-01-25
Adobe Fonts game dead link

Bonzey2416:
I tried opening the Adobe Fonts game but it no longer exists and the server couldn’t be found. So I removed the paragraph on Adobe Fonts game.


davelab6:
Thanks @Bonzey2416 - since this is just deleting something, we can accept it without the CLA if you don't want to do that. 

@vv-monsalve this is good to merge and push through :)

vv-monsalve:
> we can accept it without the CLA if you don't want to do that.

The check that was not successful is related to the CLA. But, taking Dave's above comment I'll merge this now.

PR #7235, opened 2024-01-29
update font tags

m4rc1e:
Include all families that have values for main groups.


nyshadhr9:
I think we'd still want the language tags to be removed because we can get that info from other fields in the metadata. Let me confirm with Evan.

m4rc1e:
@nyshadhr9 I spoke with Evan earlier today and he told me that we don't need the main scores for each meta category. I've just updated the csv so they're not included and updated the tooling to boot.

Hogonbeuhho (approved):

m4rc1e:
Closing since #7370 is newer

PR #7244, opened 2024-02-01
Add primary_script for Bayon and Koulen

tomasdev:
Fixes b/323348218


davelab6:
@emmamarichal please could you track this :)

davelab6:
@chrissimpkins we should probably do a complete audit of primary script metadata for the whole library........ 😭

chrissimpkins:
> @chrissimpkins we should probably do a complete audit of primary script metadata for the whole library........ 😭

@simoncozens what is the status of the 2023 work that you did on a shotgun approach to address all existing prod projects?

cc @nathan-williams

PR #7251, opened 2024-02-02
Ojuju: Version 1.000 added

emmamarichal:
d101d99: [gftools-packager] Ojuju: Version 1.000 added

* Ojuju Version 1.000 taken from the upstream repo https://github.com/jobosonchisa/ojuju at commit https://github.com/jobosonchisa/ojuju/commit/923e574a4f8003b2683319003b5e50f9e55fcdc5.

PR #7151, opened 2024-01-12
Updated designer name

emmamarichal:
I updated Raphael Alegbeleye to Raphael Alẹ́gbẹ́lẹ́yẹ̀ but I'm not sure these accents can be supported.
The doc says _Non-ascii characters are supported as UTF-8_ but I will need confirmation.

This PR can be closed if it's not supported

@garretrieger

PR #7267, opened 2024-02-08
Designer added: Ụdị Foundry

emmamarichal:
Foundry of https://github.com/google/fonts/pull/7251 project


catalog/designers/udifoundry/info.pb
```diff
@@ -0,0 +1,5 @@
+designer: "Ụdị Foundry"
```
vv-monsalve, line 1: Perhaps we should wait for an Garret's answer on #7151 to decide for these special characters?

emmamarichal: This dotbelow seems to be well supported, but yes, I'll check with him!

emmamarichal:
@vv-monsalve, I think we should go ahead, and see how is the rendering in the dev-sandbox!

PR #7286, opened 2024-02-16
Format HTML with new gftools formatter

yanone:
See https://github.com/googlefonts/gftools/pull/832


m4rc1e:
Blocking due to the line breaking algo being too primitive. It will be on initials e.g "H.R. Giger" becomes "H.
R
Giger"

m4rc1e:
`gftools format-html` produces the following:

```
gftools format-html test.html 
(venv) Marcs-Air-2:tools marcfoley$ cat test.html 
<p>
 H.R.
 Giger
</p>
```

I'd prefer it if we only inserted a line break if the line length > than say 20, so we'd get:

<p>
 H.R. Giger
</p>

yanone:
> line length > than say 20

I'm pretty sure that's the wrong approach.

Try to think of an algo that correctly breaks these two sentences:
`The characters of the film were designed by H.R. Giger. His alien characters became iconic throughout pop culture.`
`The characters of the film were designed by H.R. Giger, a Swiss sculptural artist. His alien characters became iconic throughout pop culture.`

In the first case you could find that one sentence consists only of `Giger. ` and therefore append it to the previous sentence again. In the second case you can't, at least not with a 20 character limit. This isn't gonna work.

I say we either find an algo that identifies abbreviations and won't break them, or simply use nbsp for these rare cases, which the nbsp character was literally invented for.

m4rc1e:
Nice test case. Totally agree with you.

m4rc1e:
Chat gpt is pretty good at regex


```PYTHON
import re

text = "This is a sentence. This is an abbreviation, e.g. etc. This is another sentence."

pattern = r'(?<!\b(?:e\.g|i\.e|etc))\.\s*(?=[A-Z])'

result = re.split(pattern, text)
print(result)
```

prints

```PYTHON
['This is a sentence', 'This is an abbreviation, e.g. etc. This is another sentence.']
```

yanone:
It's wrong though. There should be three sentences.

yanone:
And it's literally only for the three abbreviations i.e., e.g., and etc., and won't work for H.R. Giger.

emmamarichal:
@m4rc1e, since you've updated it, should we close that PR?

yanone:
Marc didn't update it, he deactivated it. So maybe we could first talk about a possible solution?

I looked into the issue with punctuation after tags, and indeed, _all_ HTML formatting software make that same mistake, at least all available in Python. Unbelievable, but true.

So the only way I see here to still use the formatter that I implemented (or any other) is to let go of indentation, meaning the leading spaces in front of each line that indicate HTML hierarchy. Those were causing that.

What do you guys think about that?

emmamarichal:
I'm ok with that, even if I found the previous versions a bit more lisible. I can get used to it!

@m4rc1e @vv-monsalve?

m4rc1e:
I'm going to close this PR since merging it will cause the Google Fonts backend to regenerate every font subset in our collections and it will take a very very long time (possible weeks).

Going forward, I propose we make a formatter that doesn't use prettify. I had a demo working nicely that just used lxml. I'll see if i can find it.

Issue #7277, opened 2024-02-13
CI for GFK link breakage prevention


davelab6:
A user reported:

> https://fonts.google.com/knowledge/glossary/serif and numerous others throughout the Google Fonts website include broken links to non-existent specimen pages for the Source Serif Pro and Source Sans Pro fonts.
> 
> It looks like these fonts were renamed to Source Serif 4 and Source Sans 3 and their specimen pages moved to new URLs (https://fonts.google.com/specimen/Source+Serif+4 and https://fonts.google.com/specimen/Source+Sans+3).
> 
> It would probably be a good idea to update these broken references on the various Knowledge pages and other fonts' specimen pages that refer to the two renamed fonts (including their own font specimen pages which refer to each other in their about section).

Ideally we would have CI checks on all URLs returning 2xx http codes so that we don't ship, and quick correct, broken links.

PR #7288, opened 2024-02-16
knowledge_graph.py: check outbound links

m4rc1e:
Some caveats I need to mention:

- I'm only checking for 404 response codes. Many servers will send 403s, 406s etc when using Python requests but are ok when using a webbrowser. If we want to check all 400 codes, I'll need to add more urls to the whitelist.
- We're checking ~500 urls so running the knowledge graph now takes 2 minutes. It used to take < 1 sec. If we want a speed up, we can async the code. Personally, 2 minutes seems fine since we rarely update knowledge items.
- Since it's slow, it's off by default. Include cli arg `--check_outbound_links` in order for it to run.
- SSL warnings do not fail the CI since these pages do exist.

Fixes https://github.com/google/fonts/issues/7277

### DIFF
--- a/.github/workflows/knowledge_graph.py
+++ b/.github/workflows/knowledge_graph.py
@@ -68,6 +68,7 @@ FLAGS = flags.FLAGS
 
 
 flags.DEFINE_bool("print_valid", False, "Whether to print valid links")
+flags.DEFINE_bool("check_outbound_links", False, "Check outbound urls")
 
 
 MdValue = Union[Mapping[str, "MdValue"]]
@@ -222,7 +223,8 @@ def _check_md_files(knowledge: KnowledgeContent) -> bool:
             if not target:
                 continue  # TODO: are empty links bad
             if re.search("^http(s)?://", target.lower()):
-                result = _check_outbound_link(target) and result
+                if FLAGS.check_outbound_links:
+                    result = _check_outbound_link(target) and result
             else:
                 target_path = knowledge.link_target_to_path(target)
                 result = _check_file_present(knowledge.repo_root, md_file, target, target_path) and result
--- a/.github/workflows/knowledge_graph.py
+++ b/.github/workflows/knowledge_graph.py
@@ -190,21 +190,51 @@ def _check_md_file_contents(repo_root: Path, md_file: Path, ast: List[MdValue]) 
 
 @lru_cache()
 def _check_outbound_link(url: str):
-    # Following urls work correctly on a web browser but 404 when using python requests
-    whitelist = frozenset(["https://www.jessicahische.is/talkingtype"])
+    # Following urls work correctly on a web browser but raise a 400 code when using python requests
+    whitelist = frozenset([
+        "https://www.jessicahische.is/talkingtype",
+        "http://www.layoutgridcalculator.com/typographic-scale/",
+        "https://www.figma.com/community/plugin/1088610476491668236/Material-Symbols",
+        "https://www.researchgate.net/publication/274013793_Legibility_in_Industrial_AR_Text_Style_Color_Coding_and_Illuminance",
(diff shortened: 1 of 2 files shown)


m4rc1e:
Results I get are:

```
INVALID url 'https://fonts.google.com/specimen/Source+Serif+Pro?sort=popularity&amp;category=Serif' returned response status code '404'
INVALID url 'https://fonts.google.com/specimen/Source+Serif+Pro?sort=date&amp;category=Serif' returned response status code '404'
INVALID url 'https://fonts.google.com/specimen/Source+Sans+Pro' returned response status code '404'
INVALID url 'https://fonts.google.com/specimen/Source+Serif+Pro' returned response status code '404'
INVALID url 'http://www.typografi.org/justering/gut_hz/gutenberghzenglish.html' returned response status code '404'
INVALID SSL 'https://www.very-able-fonts.com'
INVALID SSL 'https://kupferschrift.de/cms/2012/03/on-classifications/'
```

.github/workflows/knowledge_graph.py
```diff
@@ -185,21 +188,46 @@ def _check_md_file_contents(repo_root: Path, md_file: Path, ast: List[MdValue])
...
+    try:
+        url_status_code = requests.get(url).status_code
+    except requests.exceptions.SSLError:
+        print(f"INVALID SSL '{url}'")
+        # The url does exist but the cert is expired. I'm going to claim this is ok.
+        return True
+    if url_status_code == 404:
+        print(f"INVALID url '{url}' returned response status code '{url_status_code}'")
+        return False
+    return True
```
rsheeter, line 207: if you set url_status_code to 495 (ssl cert error) at line 203 then you could drop the return at 206 and have one return at what is currently 207 that is true if the status code is a success value.

Perhaps you could hang onto the response instead of the status code and simply return [response.ok](https://requests.readthedocs.io/en/latest/api/#requests.Response.ok). You could also use ok to determine whether to print the INVALID message.

rsheeter (approved):
Nit's but generally LGTM

Edit; ofc please don't submit until CI passes, either by adding a known-failure list or by fixing bad urls.

.github/workflows/knowledge_graph.py
```diff
@@ -185,21 +188,46 @@ def _check_md_file_contents(repo_root: Path, md_file: Path, ast: List[MdValue])
...
 
+@lru_cache()
+def _check_outbound_link(url: str):
+    # Following urls work correctly on a web browser but 404 when using python requests
+    whitelist = frozenset(["https://www.jessicahische.is/talkingtype"])
+    if url in whitelist:
+        return True
+
+    try:
+        url_status_code = requests.get(url).status_code
```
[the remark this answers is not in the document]
davelab6: I had suggested `requests.head` indeed, as that should make it faster and less bandwidth, possibly therefore cheaper. Regarding redirects, I noticed one of the URLs I fixed was redirecting, so I think a WARN result on the detection of redirects, and a FAIL if the redirect 4xx's, would be ideal.

davelab6:
> by fixing bad urls.

All fixed in #7291

.github/workflows/knowledge_graph.py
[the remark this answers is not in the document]
m4rc1e: > Why a full get instead of requests.head?

I had two urls fail when using this method, they worked just find with `requests.get` (allow_redirects was enabled).

.github/workflows/knowledge_graph.py
m4rc1e: `response.ok` will return false for all 400 codes. I'm only returning false for 404s. Around 20 urls in the knowledge base will return either 403 (requested resource is forbidden) or 406 (Not Acceptable client error) when using requests. If I inspect these urls in a web browser, they're fine.

m4rc1e: On second thoughts, I'll just add these 20 urls to the whitelist and drop the hammer approach.

m4rc1e:
Thanks both of you. Rod, I've implemented as much as your feedback as possible. Dave, thanks for fixing the links!

PR #7323, opened 2024-02-23
Update push lists

m4rc1e:
cc @emmamarichal @chrissimpkins 

Emma, would you mind double checking this as well?


chrissimpkins (changes requested):
We'll need menu families to be defined on the Yarndings families before we push them into production.  They remap ASCII glyphs to shapes and the Workspace editor menu names show up in those shapes, not the name of the font family

emmamarichal:
@chrissimpkins I removed the family from the list so we can move on with this one.
I block these PRs too, until I solve the metadatas

m4rc1e:
Thank you both!

PR #7352, opened 2024-03-06
Ubuntu Sans: Version 1.006 added

yanone:
d880f0e: [gftools-packager] Ubuntu Sans: Version 1.006 added

* Ubuntu Sans Version 1.006 taken from the upstream repo https://github.com/canonical/Ubuntu-Sans-fonts at commit https://github.com/canonical/Ubuntu-Sans-fonts/commit/68a263744202da55feece82d1ac33a95fa556d1a.


yanone:
About the FAILs:

Ignore the `tabular_kerning`. I confirmed that these are false positives and I'm preparing a fontbakery PR that removes at least most of them.

Dave signed off on accepting the missing case glyphs (`case_swapping`) because the authors currently are incapacitated to add them.

Dave also signed off on accepting the long `fvar` instance names.

The other checks are either false because they're dealing with the Ubuntu license (which we rarely ever have) in a wrong way, or they're metadata checks about copyright strings etc.

Basically, I think all of these should be accepted for a corporate font like the Ubuntu fonts.

yanone:
I added the tags to the second sheet in the document, just duplicating the existing data for Ubuntu & Ubuntu Mono to Ubuntu Sans & Ubuntu Sans Mono

emmamarichal:
@yanone I just noticed this interpolation issue. Otherwise it looks good to me!

Screenshot 2024-03-07 at 11 18 59

ufl/ubuntusans/DESCRIPTION.en_us.html
```diff
@@ -0,0 +1,45 @@
+<p>
+ The
+ <a href="http://font.ubuntu.com">
+  Ubuntu Font Family
+ </a>
+ are a set of matching new libre/open fonts in development during 2010-2011, with further expansion work and bug fixing in 2015 and 2022â€“2023.
```
emmamarichal, line 6: 2022â€“2023. -> 2022/2023?

emmamarichal: are a set of -> maybe "is a set" ?

ufl/ubuntusans/METADATA.pb
```diff
@@ -0,0 +1,44 @@
...
+}
+axes {
+  tag: "wght"
+  min_value: 100.0
+  max_value: 800.0
+}
+source {
+  repository_url: "https://github.com/canonical/Ubuntu-Sans-fonts"
+  commit: "68a263744202da55feece82d1ac33a95fa556d1a"
+}
```
emmamarichal, line 44: should we add stroke: "SANS_SERIF" ? at the end

yanone:
Updated

emmamarichal (approved):

PR #7376, opened 2024-03-12
Sawarabi Mincho: Version 1.082 added; ttfautohint (v1.8.4.7-5d5b) added

simoncozens:
Fails:

* check/case_mapping, check/glyphsets/shape_languages, check/glyph_coverage - this is a non-commissioned font, we have the glyphset that we have.
* check/metadata/designer_profiles - designer is relatively anonymous, we don't even have a full name.
* check/description/git_url - the description *does* have the project's git URL, but we are using an intermediate repository to bring it up to GF build standards. So contributions should go to the repo shown in the description, not to our repo.


m4rc1e (changes requested):
Some Latin marks are off.

marks

Since the amount is so much, I suspect the feature isn't working or anchors have been removed.


It seems the ascenders and descenders on lowercase letters are taller, which makes them look more aesthetic. However, I don't think the marks have been raised to compensate for this change e.g lacute

before:

Screenshot 2024-03-13 at 15 11 27

after:
Screenshot 2024-03-13 at 15 11 45

simoncozens:
Thanks for the review. Yes, it looks like anchoring has gone wrong here. This is not entirely surprising given the "experimental" nature of the toolchain (FontForge->Glyphs via babelfont). I'll try to fix.

simoncozens:
Actually these are broken in upstream's build as well (and indeed it always has been):

shape

I wonder how we created the original GF version...

simoncozens:
OK, so according to Aaron (who was not involved in the onboarding but took it upon himself to do a bit of sleuthing, thank you Aaron!), what seems to have happened is that this font got *substantially* redesigned at some point. There's a FontSquirrel version and an Adobe version, and the current PR refers to the new version.

Here's my argument for onboarding as-is:

* Per @davelab6 we don't want to maintain a fork of this font.
* Even if we did it is in Fontforge sources, so urgh.
* This is a free, non-commissioned font, and we don't want to invest in design time for the reasons above, ergo the sources are as they are. So the question is not "do we fix it" but "do we switch to this version or not"?
* And I would say "yes" - it does have (extended) Latin regressions but it's a Japanese font and it definitely has Japanese *progressions*, both improved design and improved kanji coverage.

m4rc1e:
SGTM

PR #7447, opened 2024-03-20
Edu VIC WA NT Arrows: Version 1.000 added



simoncozens:
Taken from the upstream repo https://github.com/SorkinType/VICWANTSchoolHandAustralia at commit https://github.com/SorkinType/VICWANTSchoolHandAustralia/commit/abab94d19985f24132e45db4c4822b042eb63302.
## PR Checklist:

- [x] Family categorization tags collected from the type design team with the Categories Form
- [ ] Minisite_url definition in the METADATA.pb file for commissioned projects
- [ ] Primary_script definition in the METADATA.pb file for all projects that have a primary non-Latin based language support target


simoncozens:
This is a deliberately minimal subset (A-Za-z0-9), and we are OK with that. This explains most of the FB FAILs.

vv-monsalve (approved):

vv-monsalve:
@simoncozens I merged it and then remembered this fails. 

> 🔥 FAIL Checking correctness of monospaced metadata.
🔥 FAIL
Value of hhea.advanceWidthMax should be set to 2337 but got 2586 instead.
[code: bad-advanceWidthMax]

> 🔥 FAIL MaxAdvanceWidth is consistent with values in the Hmtx and Hhea tables?
🔥 FAIL
AdvanceWidthMax mismatch: expected 2337 (from hmtx); got 2586 (from hhea)
[code: mismatch]

After a quick revision to the `htmx` table, I can confirm there is no 2586 value.

simoncozens:
It's fine, the way the average is calculated is confused by these small glyphsets.

vv-monsalve:
But the `hhea.advanceWidthMax` == higher value in `htmx`, right? In this case, it is not a calculated value.

And the second FAIL is complaining about the mismatch :)

vv-monsalve:
Working on the tagging effort, I've noticed the first Edu family merged as benchmark included the `AU` particle in the name and 'Australia' in the display name: `display_name: "Edu Australia VIC WA NT Hand"`

We had agreed on a unified naming strategy, in which case this font should be updated accordingly.

vv-monsalve:
Currently `display_name: "Edu AU VIC WA NT Hand Arrows"` 
Should be `display_name: "Edu Australia VIC WA NT Hand Arrows"`

vv-monsalve:
> But the `hhea.advanceWidthMax` == higher value in `htmx`, right? In this case, it is not a calculated value.
> 
> And the second FAIL is complaining about the mismatch :)

What about this?

vv-monsalve:
> Currently `display_name: "Edu AU VIC WA NT Hand Arrows"` Should be `display_name: "Edu Australia VIC WA NT Hand Arrows"`

I've fixed the above. However, inspecting this again, similar to the [Arrows-Guides variant](https://github.com/google/fonts/pull/7873#issuecomment-2182434882) the figures need to include the arrows.


Screenshot 2024-08-20 at 17 14 45

It needs to updated to unblock it.

cc @EbenSorkin

EbenSorkin:
I didn't get to this lost night before I fell asleep but I think I will
today.

-e.

emmamarichal:
Arrows Guide version is almost ready (some menu issues need to be solved). https://github.com/google/fonts/pull/7873
Should we stop tracking this one, so Simon can create a new PR (following also the Arrows Guide corrections)?

@vv-monsalve @simoncozens

PR #7469, opened 2024-03-25
Fix duplicate tags for Sedan, Sedan SC.

garretrieger:
Note: there was two different values for /Expressive/Sincere I choose to retain the first occuring one.


m4rc1e:
Thanks, I've just updated the tool so it'll raise an error if it encounters duplicate families across both spreadsheets, https://github.com/googlefonts/gftools/pull/885

PR #7517, opened 2024-04-03
Merge upstream.yaml into METADATA.pb [skip ci]

simoncozens:
The packager now reads from and writes to METADATA.pb instead of a separate upstream.yaml file, since we have adequate fields in the METADATA.pb file to represent the same information. This PR moves the upstream metadata from upstream.yaml into METADATA.pb. It was generated with this script:

```python
from gftools.packager import load_metadata
from gflanguages import LoadLanguages
from gftools.util import google_fonts as fonts
from pathlib import Path
import os
from tqdm import tqdm

for directory in tqdm(list(Path(".").glob("*/"))):
    fp = directory / "METADATA.pb"
    if not fp.exists() or not (directory / "upstream.yaml").exists():
        continue
    metadata = load_metadata(fp)  # Merges in upstream
    language_comments = fonts.LanguageComments(LoadLanguages())
    fonts.WriteProto(metadata, fp, comments=language_comments)
    # Remove redundant upstream.yaml file
    upstream_yaml_fp = directory / "upstream.yaml"
    os.remove(upstream_yaml_fp)
```

When reviewing please run this script yourself on the main branch, and compare the output to this branch, to confirm I am not doing anything nefarious.

I'm skipping CI because I don't want 800 fontbakery messages in this PR. I have run `fontbakery check-googlefonts -c metadata/parses ofl/*/*ttf -n` to make sure that the METADATA.pb files have not been broken.

Rod has confirmed that we don't need to push these fonts because of this metadata change, so it should not cause an upset to the backend.


m4rc1e:
@emmamarichal @vv-monsalve We don't want this pr ending up on the traffic jam or push lists.

PR #7537, opened 2024-04-10
Fix typos

Akihiro-Nagai18120:
I found some typos.


vv-monsalve:
Thanks, @Akihiro-Nagai18120. Would you like to sign the CLA agreement?

davelab6:
Since this is very trivial, and since these are all from the same vendor (who uses many DBAs), I think this is good to merge, however, we ought to check - and if found hotfix - the binaries since I assume the same typo is in them too :)

m4rc1e:
I'm going to close this since we also need to modify the fonts and then regen the METADATA.pb files. This can be done when we do the collection wide improvements.

PR #7546, opened 2024-04-12
Update to_production.txt

nyshadhr9:


vv-monsalve:
The Specimen page of Playpen Sans seems to be broken also in the current version in Production.
Screen Shot 2024-04-11 at 20 39 14

nyshadhr9:
Ah I didn't realize it's broken on prod too. Ok, then we can add it back to the list and investigate in parallel why the specimen page is broken.

PR #7557, opened 2024-04-15
Remove VIC WA NT guide families

simoncozens:
We want to call these "guides" instead.


vv-monsalve (approved):

vv-monsalve:
From the PRs that have already been merged, would these be the only ones that need to be deleted now?

vv-monsalve:
Marking this one as "Blocked" since it has reached Sandbox, but it doesn't need to go any further. Previous PRs didn't reach Prod.
@simoncozens could you confirm?

PR #7571, opened 2024-04-18
Add articles

emmamarichal:
Add articles for Gabarito, Platypi, Teachers and Rethink Sans


kenmcd:
There appears to be a typo in line 3 of the Gabarito article. 
I think "work" should be "word".

vv-monsalve:
Thanks @kenmcd. However, the `work` word is also used in the [description of the family in the type foundry](https://store.naipe.xyz/fonts/gabarito), so I'm hesitant to change it.

PR #7650, opened 2024-05-10
 Subsets update: Noto Serif Hentaigana

emmamarichal:
@simoncozens, I'm not sure about the subsets order, please correct if I'm wrong


simoncozens:
We just keep them alphabetical. I'll fix it.

PR #7682, opened 2024-05-16
PWBR: added an empty description

emmamarichal:


emmamarichal:
just read the chat conversation, we actually don't need it. I'll remove it

PR #7746, opened 2024-05-17
Modify CJK subset definitions on Cactus, Chocolate, and LXGW Wenkai projects

chrissimpkins:
This reduces the CJK subsetter definitions to a single TC subset at Rod's request


vv-monsalve (approved):

chrissimpkins:
The CI fail is due to extremely long text in the fontbakery report that prevents us from being able to push the report to the PR thread on GitHub.  Nothing changed in the font binaries here. This is a simple metadata file change.  Will merge.  

cc @m4rc1e

PR #7763, opened 2024-05-23
Turning Markdown in `figcaptions` into HTML

elliotjaystocks:


nathan-williams (approved):

vv-monsalve:
@nathan-williams is going to be in charge of this one this time. So delisting it from the Traffic board now :)

PR #7782, opened 2024-05-30
Fix sample text of Cantonese; Add sample text for Hakka; Rename Chinese languages in a consistent format

laubonghaudoi:
This PR fixes the following issues:

1. The original sample text of `yue_Hant` (Cantonese) is actually in Simplified Chinese characters. This PR updates them into Traditional Chinese characters.
2. Sample text for Hakka Chinese was missing. This PR adds sample texts for Hakka.
3. The naming of Chinese languages was inconsistent. This PR renames them with the consistent `XX Chinese` format.


chrissimpkins:
Thank you very much for this contribution!

Going to close as a merge will break our git subtree.  As Simon mentioned, https://github.com/googlefonts/lang is the correct upstream location for these changes.  You'll find the same file path structure and it should be as simple as a copy/paste into the new repo.

laubonghaudoi:
Ah thank you! I will PR to that repo

PR #7792, opened 2024-05-30
Add `primary_language` field to Chinese fonts; Add languages to new Chinese fonts;

laubonghaudoi:
Current Traditional Chinese fonts don't have a primary language setting in their meta data, resulting in the default sample text to be in `wuu_Hant` which has the highest population. This PR specifies the `primary_language` fields of them. 

The PR also adds the Chinese languages to the metadata of new Chinese fonts.


chrissimpkins:
@simoncozens Mind reviewing this PR when you have an opportunity?

chrissimpkins:
cc @nathan-williams re: incoming changes to front end specimens

davelab6 (approved):

simoncozens:
The problem here is that fontbakery's version of the protobuf headers doesn't have a `primary_language` field, so it's not being parsed correctly there. I'll fix in fontbakery and try again.

nathan-williams (commented):
We should not specify languages explicitly in the METADATA.pb files. We currently do so, but only for Noto families. Please remove any specification of `languages` in the textprotos for non-Noto families.

Otherwise, PR looks good. Thanks again for the contribution!

nathan-williams (changes requested):
See [earlier comment](https://github.com/google/fonts/pull/7792#pullrequestreview-2094248810).

laubonghaudoi:
I see. But if I remove all `languages` in the non-Noto families, how to specify the default displayed sample text for them? Currently they are all showing Shanghainese texts by default which is what this PR is trying to fix.

laubonghaudoi:
I just updated the PR and removed the `languages` and `primary_language` for all non-Noto fonts. I see that https://github.com/fonttools/fontbakery/pull/4756 is working in progress, IIUC I should make another PR to add the languages back once that PR is merged and rolled out, so that the default display language issue can be fixed. Is this correct?

nathan-williams (approved):
The current set of changes looks good!

You can actually restore some of the reverted changes, though. Specifying `primary_language` for non-Noto families is fine – just not `languages`.

laubonghaudoi:
Okay done, thanks for letting me know!

PR #7802, opened 2024-05-31
Update lists of the 31th of May

vv-monsalve:


chrissimpkins (approved):

vv-monsalve:
The knowledge check is still failing on an internet connection.

Rod said in the chat:

> The offending content appears to be cc-by-sa/knowledge/modules/using_variable_fonts_on_the_web/lessons/optimizing_typographic_space_with_variable_fonts/content.md which includes "tools like [Font-To-Width](http://font-to-width.com/) or [fit-to-width](https://github.com/Lorp/fit-to-width) can make it relatively easy."

And indeed, the link `http://font-to-width.com` is not loading.

Merging this PR here

PR #7803, opened 2024-05-31
Fix link in optimizing_typographic_space_with_variable_fonts/content.md

davelab6:
https://web.archive.org/web/20240425214330/http://font-to-width.com/ works


rsheeter (approved):
LGTM though I do wonder, if the publisher has taken down the site are respecting their wishes by retaining an archive link? - my inclination would have been to delete the reference.

PR #7850, opened 2024-06-12
Laura Garcia Mut designer info

yanone:
- Added profile for Laura
- Put her second in Fustat's metadata since she designed while Khaled engineered (Section is called "designers")
- Added Arabic subset for Fustat


khaledhosny:
> * while Khaled engineered

While the order doesn't matter much, just to make things clear: I was also involved in the Arabic design both directly by drawing some glyphs and indirectly by giving feedback to Gaber during the design process.

yanone:
I drew my inspiration for this change from the font's [README](https://github.com/Kief-Type-Foundry/Fustat) which is very clear on who did what.
Thank you for the clarification.

khaledhosny:
I missed that. Fair enough.

PR #7875, opened 2024-06-21
Matemasie: Version 1.001; ttfautohint (v1.8.4.7-5d5b) added



n8willis:
Taken from the upstream repo https://github.com/YADAMSS/Matemasie-Font at commit https://github.com/YADAMSS/Matemasie-Font/commit/f8b28b0ecb71e4556e18b2f4af23ca151243cac0.
## PR Checklist:

- [x] Family categorization tags collected from the type design team with the Categories Form
- [x] Minisite_url definition in the METADATA.pb file for commissioned projects
- [x] Primary_script definition in the METADATA.pb file for all projects that have a primary non-Latin based language support target


m4rc1e:
@nathan-williams quite a few fails need addressing before we can merge this.

n8willis:
Updated

n8willis:
Updated

m4rc1e:
Looking good! Could you add description to the article file please?

n8willis:
Updated

n8willis:
Updated

m4rc1e:
The Article file is here, https://github.com/google/fonts/pull/7875/files#diff-5d21c92f508684d5a272559aa366d1a491ef96f1e057b6908d57c289f107973b and its content is the placeholder "N/A". Please populate the file.

n8willis:
The ARTICLE file is there: https://github.com/YADAMSS/Matemasie-Font/blob/main/ARTICLE.en_us.html ... it seems like perhaps the packager isn't grabbing the latest commit? I thought it was automatically going to do that (the packager was run from the same machine where the local commit was made in the upstream repo).

vv-monsalve:
The packager uses what is specified in the metadata file. It needs to include the source information for the article. See  [this](https://github.com/google/fonts/blob/main/ofl/playwritear/METADATA.pb)

n8willis:
Thanks for replying; I'm definitely glad to know there's a straightforward fix.

But booyhowdy, the error message sure does not point a person in the right direction at all.... Clarifying that would really make a big usability improvement for people.  Couple of quick thoughts there -- first, the root issue it's finding is actually the packager command not seeing a file where it expects to, but the error shown in the check is from FontBakery, about a missing HTML tag. If the packager needs the file to exist, then it could report that in its own output on the terminal, with the other stuff it reports. That would definitely be clearer, but it'd also be more helpful practically, because it'd let the user make that fix immediately, instead of pushing it to the PR and waiting to see the check output.

(Secondarily, if the presence of the file is mandatory, then could the packager look for it by name (like it does with OFL.txt) and handle it when constructing the boilerplate METADATA.pb?)

Also, separately, can a list of the required files please be added to the documentation? If Article is required, and it has to be specified in the .pb file, that goes somewhere. This detail is not on the page about the Article and not on the page about METADATA.pb. I looked in both places trying to figure out the fix; I don't know when I would have found it just by searching on my own.

n8willis:
Updated

n8willis:
Updated

vv-monsalve:
- For a few reasons, not every PR requires the addition of an Article (e.g., when updating a font), so the packager should not warn about it; the person making the PR knows the details.
- The CI Fontbakery run on each PR is reporting about the missing files at PR level 
- We are constantly evolving and improving the process, hence the tools. After this PR was created, the Packager now includes an empty article if it doesn't find one and warns you about it. This is why we don't recommend opening draft PRs.
- Similarly, the change to deprecate the original `.yaml` file and move to relay on the Metadata file is recent.
- The [Article](https://googlefonts.github.io/gf-guide/article.html) and [Metadata](https://googlefonts.github.io/gf-guide/metadata.html#example-variable-fonts-family) contents on the GF-Guide have been updated following the above
- All these changes in requirements and tools have been reported/discussed during the weekly meetings.
- Please feel free to open Issues or PRs in the repositories (GUide, gftools, or Fontbakery) with any improvement you would like to propose

vv-monsalve:
> The Article file is here, https://github.com/google/fonts/pull/7875/files#diff-5d21c92f508684d5a272559aa366d1a491ef96f1e057b6908d57c289f107973b and its content is the placeholder "N/A". Please populate the file.

Re-reading this now, it looks like the packager already included this feature at this PR time. AFAIR, it throws a message on the terminal about it.

m4rc1e:
@n8willis thank you very much. LGTM

n8willis:
@m4rc1e Can you add it to the traffic jam as well? I don't have permissions to access that

emmamarichal:
Hi @n8willis, could you ask the designer to fill out the [designer profile form](https://docs.google.com/forms/d/e/1FAIpQLSeMwHN8J213ZaxHrr5lHCrX56HY_NjGrWB8o604g98YxuMrdA/viewform), so we can add him/her to the catalog?

n8willis:
Sure thing @emmamarichal

PR #7992, opened 2024-08-05
Update tiny5/METADATA.pb

davelab6:
Per auto closed PR #7816 this should match upstream

### DIFF
--- a/ofl/tiny5/METADATA.pb
+++ b/ofl/tiny5/METADATA.pb
@@ -10,7 +10,7 @@ fonts {
   filename: "Tiny5-Regular.ttf"
   post_script_name: "Tiny5-Regular"
   full_name: "Tiny5 Regular"
-  copyright: "Copyright 2024 The tiny5 Project Authors (https://github.com/Gissio/font_tiny5)"
+  copyright: "Copyright 2022-2024 The Tiny5 Project Authors (https://github.com/Gissio/font_tiny5)"
 }
 subsets: "cyrillic"
 subsets: "cyrillic-ext"


emmamarichal (approved):

emmamarichal:
@m4rc1e I see "manifest" in these errors:
💥 ERROR Check if the vertical metrics of a CJK family are similar to the same family hosted on Google Fonts.
💥 ERROR Check if the vertical metrics of a family are similar to the same family hosted on Google Fonts.

I saw the manifest error last week with the traffic jam tool, but I'll not merge that PR without confirmation. What would be the best move here?

vv-monsalve:
> I saw the manifest error last week with the traffic jam tool

What do you mean by "manifest" error?

emmamarichal:
@vv-monsalve Inside the check: 

```
💥 ERROR
'manifest'


[code: error]
```


I don't know what "manifest" means, but I had it also with the traffic jam tool

m4rc1e:
Just fixed fontbakery https://github.com/fonttools/fontbakery/pull/4810. "manifest" appears when we're not able to download the family from a server. This should only happen if the family hasn't been pushed to the server yet, once the fb pr has been merged. I'll make a fix in manage traffic jam as well.

vv-monsalve:
This pull request was included in the 'to_sandbox' list and technically incorporated into the latest push. However, at present, it is not feasible to track this change through the servers, not even manually. Therefore, I am adding it to the "Blocked" list to prevent it from being added to the lists until we find a way to track these changes.

m4rc1e:
The download zip's OFL file has been changed so we can track it. 

Screenshot 2024-11-28 at 21 08 59

PR #8000, opened 2024-08-07
Switch Noto Color Emoji to Noto Emoji style description

rsheeter:


vv-monsalve:
Hi @rsheeter are you going to push this change through the servers on your end? Or should we add it to the Traffic board now so that we can track its movement through the pipeline?

rsheeter:
Ty for checking, I'll do it

vv-monsalve:
Okay

Issue #7545, opened 2024-04-11
COLRv1 font inconsistent color rendering


vv-monsalve:
**Describe the issue**

A COLRv1 project, which includes gradients and transparencies, shows rendering issues on the Dev server.

It doesn't consistently display the colors; it seems to need the page to be reloaded since it only shows them after switching between tabs.

- The first preview after searching works ok
<img width="1230" alt="Screen Shot 2024-04-11 at 16 41 03" src="https://github.com/google/fonts/assets/48698976/98648d60-c204-4ecd-a8a8-87bb089db855">    

- But then on the `Specimen` tab
    <img width="1471" alt="Screen Shot 2024-04-11 at 14 25 08" src="https://github.com/google/fonts/assets/48698976/eeaa099b-6ded-4b17-8f8b-b1678984e8c0">

- The `Type Tester` and `Glyphs` tabs seem to be working well always
    <img width="1496" alt="Screen Shot 2024-04-11 at 14 25 15" src="https://github.com/google/fonts/assets/48698976/a673f4a7-b520-446b-8b51-985248f778ae">

- Returning to the search results
    <img width="1477" alt="Screen Shot 2024-04-11 at 16 45 04" src="https://github.com/google/fonts/assets/48698976/b2e7cb79-2bf9-4939-b695-4bfe8017688f">

**To Reproduce**
Enter the server and follow the above process

**Expected behavior**
For it to work consistently 

Browser used: Chrome Version 123.0.6312.107 (Official Build) (x86_64)

cc @davelab6

tomasdev:
Debug info: color is retained when asking a static instance like wght:400. Color is lost when asking for a VF subset like wght:100..700

anthrotype:
> Color is lost when asking for a VF subset like wght:100..700

maybe a in hb-subset

davelab6:
Tracking internally at [b/334968053](https://b.corp.google.com/issues/334968053)

davelab6:
This is confirmed as a problem with the API and the team is looking into it; Viviana has marked the family as blocked until this is resolved.

vv-monsalve:
The family is being updated with an added Light color palette. It would be ideal for this fix to happen before the new font is ready. Is it feasible?

davelab6:
This was because hb-subset did not support partial instancing of COLRv1 tables but @qxliu76 implemented it! :) so this should be all working in all envs now

vv-monsalve:
[...]

PR #7919, opened 2024-07-04
Sixtyfour Convergence: Version 2.001 added



vv-monsalve:
Taken from the upstream repo https://github.com/jenskutilek/homecomputer-fonts at commit https://github.com/jenskutilek/homecomputer-fonts/commit/3e0b1194c1d2cc6bc66178908470a428f34349cd.
## PR Checklist:

- [x] Family categorization tags collected from the type design team with the Categories Form
- [ ] Minisite_url definition in the METADATA.pb file for commissioned projects
- [ ] Primary_script definition in the METADATA.pb file for all projects that have a primary non-Latin based language support target
- [ ] Fontbakery checks are reviewed and failing checks are resolved in collaboration with the upstream font development team
- [ ] Diffenator2 regression checks for revisions on all projects that are currently in production
- [x] Designers bio info have to be present in the designer catalog (at least an issue should be opened for tracking this, if they are not)
- [x] Check designers order in metadata.pb, since the first one of the list appears as “principal designer”
- [ ] Social media formatted visual assets for all new commissioned projects in the Drive directory, communicate with the repository Maintainer so that they can push this content to the Social Media tracker spreadsheet
- [ ] Social media content draft for all new commissioned projects in the Drive directory and Social Media tracker spreadsheet, communicate with the repository Maintainer so that they can push this content to the Social Media tracker spreadsheet

PR #8024, opened 2024-08-13
Touch Sixtyfour Convergence metadata file to get it to repush.

garretrieger:
To get subsetter fix: https://github.com/harfbuzz/harfbuzz/pull/4831.

For #7545


vv-monsalve:
This change will continue to be tracked from Sandbox by #7919

PR #8044, opened 2024-08-16
SUSE: Version 1.000 added



vv-monsalve:
Taken from the upstream repo https://github.com/SUSE/suse-font at commit https://github.com/SUSE/suse-font/commit/d5d0b353f3defb9dece73334a50cc4a62ae97e00.
## PR Checklist:

- [ ] `minisite_url` definition in the METADATA.pb file for commissioned projects
- [ ] `primary_script` definition in the METADATA.pb file for all projects that have a primary non-Latin based language support target
- [ ] `subsets` definitions in the METADATA.pb reflect the actual subsets and languages present in the font files (in alphabetic order). For **CJK fonts**, only include one of the following subsets `chinese-hongkong`, `chinese-simplified`, `chinese-traditional`, `korean`, `japanese`.
- [ ] Fontbakery checks are reviewed and failing checks are resolved in collaboration with the upstream font development team
- [ ] Diffenator2 regression checks for revisions on all projects that are currently in production
- [ ] Designers bio info have to be present in the designer catalog (at least an issue should be opened for tracking this, if they are not)
- [ ] Check designers order in metadata.pb, since the first one of the list appears as “principal designer”
- [ ] Social media formatted visual assets for all new commissioned projects in the Drive directory, communicate with the repository Maintainer so that they can push this content to the Social Media tracker spreadsheet
- [ ] Social media content draft for all new commissioned projects in the Drive directory and Social Media tracker spreadsheet, communicate with the repository Maintainer so that they can push this content to the Social Media tracker spreadsheet

PR #8079, opened 2024-08-23
Push lists August 23 update

vv-monsalve:
@chrissimpkins thi PR includes #7919 for a COLRv1 font and #8044 a `p2` project.

### DIFF
--- a/to_production.txt
+++ b/to_production.txt
@@ -1,5 +1,4 @@
 # New
-ofl/sixtyfourconvergence # https://github.com/google/fonts/pull/7919
 ofl/suse # https://github.com/google/fonts/pull/8044
 
 # Designer profile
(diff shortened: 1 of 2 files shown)


to_production.txt
```diff
@@ -1,3 +1,17 @@
-# Upgrade
+# New
+ofl/sixtyfourconvergence # https://github.com/google/fonts/pull/7919
```
chrissimpkins, line 2: Let's remove this COLRv1 family from the list.  We need time to prevent this from landing in Workspace editors

chrissimpkins (changes requested):

to_production.txt
vv-monsalve: Done!

chrissimpkins (approved):

PR #8086, opened 2024-08-26
Set primary scripts (4/16)

simoncozens:


emmamarichal:
Maybe these bugs are related to the lang PR, so it would be better just to wait for it?

simoncozens:
We don't need a new PR for these fonts, the metadata is now correct so there is nothing to fix on our end. The problem is with the front-end; best to ask the eng team.

emmamarichal:
Blocked because East Sea Dokdo / Gamja Flower / Gidugu and Gothic A1 aren't working on the dev-server
Screenshot 2024-09-13 at 15 04 12

PR #8091, opened 2024-08-26
Set primary scripts (9/16)

simoncozens:


emmamarichal:
Blocked, mogra doesn't work:

Screenshot 2024-09-13 at 16 26 55

emmamarichal:
@chrissimpkins That PR is on the to_sandbox list, but Mogra still doesn't work (see my previous comment).
I suggest we don't block these primary script PRs, and fix the non-working ones in a new PR later.

**Live**
 longcang
 mashanzheng

**Working on dev-sandbox**
metal
mirza
mochiypopone
mochiypoppone
modak

**Not working on dev-sandbox**
 mogra

**Font not even on dev-sandbox**
 lohitbengali
 lohittamil

PR #8092, opened 2024-08-26
Set primary scripts (10/16)

simoncozens:


emmamarichal:
Blocked since nanumbrushscript, nanumgothic don't work.

PR #8107, opened 2024-08-27
Adds fixes to typographic tag issues identified by Evan

kalapi:
@evanwadams Fixed the typographic tag issues you pointed out in the following link:

https://docs.google.com/spreadsheets/d/1cNUL2i6C5heKMD6v0W1jdUBluO37L9PshzrI7fXzPOc/edit?gid=1033747182#gid=1033747182

PR #8111, opened 2024-08-29
families.csv: sort alphabetically

m4rc1e:
@evanwadams Here's an updated csv that has been sorted using the tool.

At first it seems strange but you soon realise the original csv wasn't sorted correctly since ADLam Display should appear before Abel.


m4rc1e:
@evanwadams it probably makes sense for us to merge #8107 and then I redo this.

PR #8148, opened 2024-09-08
Fix broken link in figcaption

ethamck:
Broken link

at https://fonts.google.com/knowledge/designing_hangeul/an_introduction_to_hangeul

(I'm not familiar with the site's build system or style and haven't tested these changes outside of Pandoc, but editing the source of the site and adding correct `<p>` and `<a>` tags

Old figcaption

The new figcaption wrapped in a paragraph

does not break margins so I assume that this change will work.)


ethamck:
> Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

I don't have a Google account and won't make one. If that nullifies this PR, fine, but I don't think I could ever claim copyright over two newlines.

m4rc1e:
Thanks! I'll merge it and see if it solves the issue in our sandbox server.

m4rc1e:
It works in our sandbox server so thank you very much

Screenshot 2024-09-11 at 14 35 17

PR #8185, opened 2024-09-19
Use Kore for Hangul

simoncozens:
ISO15924 declares `Kore` as an alias for `Hang` + `Hani`, which is what we want for these fonts (and also what we have defined in the lang repo...)

Also, we need a fontbakery check to make sure that the languages and scripts defined in the metadata files actually exist...


emmamarichal (approved):

simoncozens:
Looks like that check is broken, because of the way it is trying to download the existing families from GF. Please note the font files haven't changed at all, just the METADATA.pb, so we should be fine...

PR #8207, opened 2024-09-26
families.csv: implement Evan feedback

m4rc1e:
@evanwadams I've also removed tags for the "Not text" families.


m4rc1e:
Evan was ok with me merging this pr.

Issue #7062, opened 2023-12-05
Update metadata in OFL fonts to point to new openfontlicense.org


jvgaultney:
**Describe the issue**
The SIL Open Font License web site has moved from both

http://scripts.sil.org/ofl
https://scripts.sil.org/ofl

to a new location:

https://openfontlicense.org

Please change the license URL metadata within OFL-licensed fonts (UFOs, Glyphs files, etc.) to point to the new site.

davelab6:
Thanks for posting this, the new site is great. We'll work though this in the new year :)

felipesanches:
I'll update FontBakery to also accept this new URL

PR #8208, opened 2024-09-26
Matemasie: Update article

n8willis:
Taken from the upstream repo https://github.com/YADAMSS/Matemasie-Font at commit https://github.com/YADAMSS/Matemasie-Font/commit/982fb1ec023f63c0da05a11a2bec15efd1470484.
## PR Checklist:


ofl/matemasie/OFL.txt
```diff
@@ -2,7 +2,7 @@ Copyright 2024 The Matemasie-Font Project Authors (https://github.com/YADAMSS/Ma
 
 This Font Software is licensed under the SIL Open Font License, Version 1.1.
 This license is copied below, and is also available with a FAQ at:
-https://openfontlicense.org
+https://scripts.sil.org/OFL
```
vv-monsalve, line 5: The url needs to go back to https://openfontlicense.org

vv-monsalve (changes requested):

ofl/matemasie/OFL.txt
n8willis: The upstream file has not changed: https://github.com/google/fonts/pull/7875/files#diff-5c8e4fe8229bd791faab95db2eeb75913e62cde987091ba4f3dde445e5c317e6

It has always had the SIL URL. 

If the check is saying that it was altered, then it sounds like the check has changed since the previous PR, or else packager did something different at the time of the original PR.

vv-monsalve: > It has always had the SIL URL.

It changed. Per [this request](https://github.com/google/fonts/issues/7062)

n8willis: Yes; we know there's a new URL and Fontbakery checks for it in the `name`` table of the binary. The question I was exploring was why it did not also check the URL in the contents of the OFL.txt file.

vv-monsalve: It's better for @felipesanches to answer that question.

n8willis: I discussed it with him on video call today, and it is the subject of the FB bug I opened after the call, which is listed in the links below.

vv-monsalve: Yes, I saw the FB Issue. However, the string still needs to be changed here.

n8willis: Well, the preceding comment inquired about talking to Felipe, therefore I replied to indicate I had already talked to Felipe. 🤷

> However, the string still needs to be changed here

It needs to be changed in the upstream repo. I've also already raised that need with upstream; when it is fixed there, it will propagate here.

n8willis:
Updated

m4rc1e:
Updated

vv-monsalve (approved):

vv-monsalve:
Image size fixed in #8229

PR #8225, opened 2024-10-01
Adds QA updates as agreed upon for tag: Cute

kalapi:
Updates tag data for 'Cute' with decisions made in [this sheet](https://docs.google.com/spreadsheets/d/15ivqgq3HHAfFqhHgbm7PfxJw71Dkxxa9o_ycZ_Wde_c/edit?gid=1899333079#gid=1899333079).

cc: @evanwadams @m4rc1e


kalapi:
@evanwadams you wrote:

>Cute (8225)
>- Akaya Kanadaka is lowered, but is not in the spreadsheet
>- Akaya Telivigala is lowered, but is not in the spreadsheet
>- Balsamiq Sans now has two Cute rows with values of 39 (old) and 34 (new). Is this bug in Marc's tool?
>- Dangrek has two rows
>- Sansita has two rows
>- Shrikhand was raised, but is not in the spreadsheet

- Reverted scores for those not in spreadsheet
- Removed the double tags; Yes, this is a bug in the tool. It allows multiple scores under the same category which should be disallowed. I noticed this early in the process and made sure to double check for all other categories.

PR #8246, opened 2024-10-03
Geist: Version 1.400 added



vv-monsalve:
Taken from the upstream repo https://github.com/vercel/geist-font at commit https://github.com/vercel/geist-font/commit/1bd07b141c309bae59a982e2a43eea96ae4cbdde.
## PR Checklist:


emmamarichal:
Hi @vv-monsalve:
I found almost the same issues than in the mono version! Here are some precisions:

**Interpolation**
- It's weird that the Behdad's tool didn't detect it:
Screenshot 2024-10-04 at 17 49 11

**Anchors**
- Same as Mono, with no top anchors on accents:
Screenshot 2024-10-04 at 18 12 05


**Cyrillic**
- I compared with some other GF fonts, and the placement seems to be wrong indeed:
Screenshot 2024-10-04 at 17 57 42
Screenshot 2024-10-04 at 17 58 14

- The issue we had in the mono isn't here, so I would follow this example, with the bars alignement:
Screenshot 2024-10-04 at 17 56 23

**Licence url**
We need to add back the check about the new URL. Here the licence field isn't in the source info, and the url is wrong. 
Screenshot 2024-10-04 at 15 48 13

**Small adjustments**
- (same as Mono) Alignements could be improved a bit (but I'm ok to merge it like this)
Screenshot 2024-10-04 at 15 30 33

**Kerning**
- The cyrillic kerning could be improved:
Screenshot 2024-10-04 at 18 07 38
Screenshot 2024-10-04 at 18 08 16
Screenshot 2024-10-04 at 18 08 22
Screenshot 2024-10-04 at 18 09 02

vv-monsalve:
@emmamarichal, Some of the reported kerning in Cyrillic, are working ok in the font.
I've reported this as a suggestion. However, this is a submitted font for which Latin Core is the base support requirement. This should not be considered a blocker.
Screenshot 2024-10-04 at 16 32 54

vv-monsalve:
Updated

vv-monsalve:
@emmamarichal Update for the update:

- 🔥 FAIL Check there are no overlapping path segments
   Reported ti upstream. I'll need to wait for them to make the changes >.<
- 🔥 FAIL METADATA.pb: Designers are listed correctly on the Google Fonts catalog?
   The designer didn't want to include an image. I would not consider this as a blocker.

vv-monsalve:
Updated

vv-monsalve:
@emmamarichal 

- > 🔥 FAIL Check there are no overlapping path segments
   
   The design team worked on the previously reported cases, but some new ones emerged. I tested the concerned glyphs on Chrome and Indesign, and there were no rendering issues. Also, Marc said about this check, "I think this check may already be redundant since Chrome 130 has fixed it." So I would not consider this a blocker.

emmamarichal:
Lgtm! 

I still see some small issues in cyr, but I'll not consider them as blockers, if we want to focus on latin fom now.
I just report them here, we can open an issue on the repo for the next update! 

I'll wait for your answer before merging it!

### **Cyrillique**

- **uni0409**: this thicker stem was not here in the previous version
Screenshot 2024-10-24 at 10 08 43

- **very small connexion issue with the components:**
Screenshot 2024-10-24 at 10 09 41
Screenshot 2024-10-24 at 10 10 12

- **alignement issue:**
Screenshot 2024-10-24 at 10 09 55

emmamarichal (approved):

vv-monsalve:
> I'll wait for your answer before merging it!

After the initial Cyrillic comments, designers said: " the cyrillic was not reviewed and requires a complete overhaul". 

So, the best option for a clean release would be to delist the Cyrillic subset. I'll do it in a quick follow-up PR today.
I'll also add the above comments to the upstream issue where this was discussed.

PR #8251, opened 2024-10-03
Unicode 16.0 emoji

rsheeter:
https://github.com/googlefonts/noto-emoji/releases/tag/v2.047

I have already taken this through the release process, no need to add to to_x.txt


vv-monsalve:
Noted! Thanks for the comment, @rsheeter.

PR #8253, opened 2024-10-03
Matemasie is failing to push

rsheeter:
Based on other examples I think it isn't supposed to have article/ in the img src


vv-monsalve:
Thanks, @rsheeter! Is Dev now unblocked?

rsheeter:
> Is Dev now unblocked?

Yes

Issue #8308, opened 2024-10-11
Add Edu VIC WA NT Arrows


emmamarichal:
**Font Project Git Repo URL:**

https://github.com/SorkinType/VICWANTSchoolHandAustralia

PR #8319, opened 2024-10-11
Edu AU VIC WA NT Arrows: Version 1.001 added

simoncozens:
Taken from the upstream repo https://github.com/SorkinType/VICWANTSchoolhandAustralia at commit https://github.com/SorkinType/VICWANTSchoolhandAustralia/commit/4d4efe710065410b34b66fa9d588498351642f41.

Adds arrows in numerals. See https://github.com/google/fonts/pull/7447#issuecomment-2299858628

See #8308.


emmamarichal:
@simoncozens These two fails seem to be a related to a width/spacing particularity:

🔥 FAIL MaxAdvanceWidth is consistent with values in the Hmtx and Hhea tables?
🔥 FAIL Checking correctness of monospaced metadata.

Should we correct that or this is related to the design and intentional?

emmamarichal:
Otherwise, the font looks good to me.

simoncozens:
This is a subsetter bug caused by https://github.com/harfbuzz/harfbuzz/issues/4931 in Harfbuzz.

However, we have been serving fonts over the web API for years using Harfbuzz subsetting, and so with the same problem, but nobody has complained, so maybe it's not a big deal. :-)

vv-monsalve (approved):

PR #8354, opened 2024-10-18
Update pushlists

m4rc1e:
cc @chrissimpkins

### DIFF
--- a/to_production.txt
+++ b/to_production.txt
@@ -1,6 +1,3 @@
-# New
-ofl/notoserifhentaigana # https://github.com/google/fonts/pull/7327
-
 # Upgrade
 ofl/miriamlibre # https://github.com/google/fonts/pull/7534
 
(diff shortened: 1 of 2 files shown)


to_production.txt
```diff
@@ -1,12 +1,13 @@
 # New
-
+ofl/notoserifhentaigana # https://github.com/google/fonts/pull/7327
```
chrissimpkins, line 2: There was an issue with primary_script in sandbox.  I think that we might still need to solve this issue before the prod push?

@simoncozens to confirm that this is not meant as a fix for the issue in the production push here?

chrissimpkins (changes requested):

to_production.txt
chrissimpkins: cc @nathan-williams

chrissimpkins: Let's remove the Noto Serif Hentatigana family from the to_prod list for now so that we can continue to move the other families.  Ty!

emmamarichal (approved):

PR #8476, opened 2024-11-08
Bad Script: Version 2.000; ttfautohint (v1.8.4.7-5d5b) added



emmamarichal:
Taken from the upstream repo https://github.com/alexeiva/badscript at commit https://github.com/alexeiva/badscript/commit/dca2962433a2b5817f1e716d6731a743440fbd79.
## PR Checklist:


emmamarichal:
⚠️ WARN Check if uppercase glyphs are vertically centered.
-> I suggested to update that WARN, so it concerns only new fonts: https://github.com/fonttools/fontbakery/issues/4877

vv-monsalve:
- > -> I suggested to update that WARN, so it concerns only new fonts: [fonttools/fontbakery#4877](https://github.com/fonttools/fontbakery/issues/4877)

It has a Warn status here. Being a font upgrade we know we'll need to accept the previous VM.

- The new `R` could be considered a regression for this script font. Would it remain accessible to users in a stylistic set?

bad-script-ba

emmamarichal:
@vv-monsalve 

> It has a Warn status here. Being a font upgrade we know we'll need to accept the previous VM.

Ah yes, I meant it would be great if we can update the warn itself, so it can check centered caps only on new fonts.

> The new R could be considered a regression for this script font. Would it remain accessible to users in a stylistic set?
Yeah, it is now in a .fina feature. I think it was actually the first intention, because it looks weird in the middle or beginning of a word!

vv-monsalve (approved):

PR #8479, opened 2024-11-12
add new GD article to modules list

shailenpatel2:


cc-by-sa/knowledge/modules/choosing_type/module.textproto
```diff
@@ -1,4 +1,5 @@
 name: "Choosing type"
+lessons: "choosing_web_fonts_beginners_guide"
```
nathan-williams, line 2: Order matters. Suggest checking in with Tobias/Dave/someone else on where they would expect this article to slot in. FWIW, the first position makes sense to me.

nathan-williams (approved):

PR #8489, opened 2024-11-14
Mona Sans: Updated metadata (designers)

emmamarichal:
We had a feedback from Tobias, with the right names that should appear. The designers profile addition will follow!


vv-monsalve (approved):

vv-monsalve:
- > 🔥 FAIL METADATA.pb: Check URL on copyright string is the same as in repository_url field.

Discussed during today's meeting. @emmamarichal I'll wait for you to update the metadata with this URL :)

PR #8493, opened 2024-11-14
Add Degarism Studio

emmamarichal:
Github fonts team (Mona Sans and Hubot Sans)


catalog/designers/degarismstudio/bio.html
```diff
@@ -0,0 +1,2 @@
+<p>Degarism Studio, established by Deni Anggara in 2015, is an independent design studio focused on typography, type, and editorial design—known for fonts such as Alliance and Biotif.</p>
```
vv-monsalve, line 1: "focused on typography, type," sounds redundant. Could it be just typography?

vv-monsalve (changes requested):

catalog/designers/degarismstudio/bio.html
vv-monsalve: I've changed it to speed up the process. Please feel free to update it later if you think it needs to be restored.

vv-monsalve (approved):

PR #8495, opened 2024-11-14
Update families.csv

nyshadhr9:
Remove header.


m4rc1e:
This has rekt the tagging tool, https://google.github.io/fonts/tags.html

PR #8496, opened 2024-11-15
ci: update `upload-artifact` and `download-artifact` actions to v4

Juneezee:
v3 of `actions/upload-artifact` and `actions/download-artifact` will be fully deprecated by 5 December 2024. Jobs that are scheduled to run during the brownout periods will also fail. See:

1. https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
2. https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions/

An attempt was made to update the artifact actions in PR https://github.com/google/fonts/pull/8159, but the version was eventually downgraded from `v4` to `v3` due to artifact name conficts in PR https://github.com/google/fonts/pull/8210. This is because v4 of `actions/upload-artifact` does not allow uploading to the same artifact name anymore and we need to give each artifact a unique name. See:

1. https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact
2. https://github.com/actions/upload-artifact?tab=readme-ov-file#not-uploading-to-the-same-artifact
3. https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/

/cc @simoncozens @m4rc1e

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -61,9 +61,9 @@ jobs:
 
       - name: Upload check results
         if: steps.check_files.outputs.files_exists == 'true'
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: qa
+          name: qa-diffbrowsers-${{ matrix.os }}
           path: out/
 
   diffenator:
@@ -111,9 +111,9 @@ jobs:
 
       - name: Upload check results
         if: steps.check_files.outputs.files_exists == 'true'
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: qa
+          name: qa-diffenator
           path: out/
 
   ftxvalidator:
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -59,7 +59,7 @@ jobs:
       # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
       # format to the repository Actions tab.
       - name: "Upload artifact"
-        uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8 # v3.1.0
+        uses: actions/upload-artifact@184d73b71b93c222403b2e7f1ffebe4508014249 # v4.4.3
         with:
           name: SARIF file
           path: results.sarif
--- a/axisregistry/.github/workflows/publish-release.yml
+++ b/axisregistry/.github/workflows/publish-release.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Build a binary wheel and a source tarball
         run: python3 -m build
       - name: Store the distribution packages
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: python-package-distributions
           path: dist/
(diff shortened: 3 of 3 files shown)


simoncozens (approved):
LGTM.

m4rc1e:
Thanks for the PR!

Slight issue we need to solve first is that v4 has the following breaking change:

> Due to how Artifacts are created in this new version, it is no longer possible to upload to the same named Artifact multiple times. You must either split the uploads into multiple Artifacts with different names, or only upload once. Otherwise you will encounter an error.

.github/workflows/test.yaml
```diff
@@ -61,9 +61,9 @@ jobs:
 
       - name: Upload check results
         if: steps.check_files.outputs.files_exists == 'true'
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: qa
+          name: qa-diffbrowsers-${{ matrix.os }}
```
Juneezee, line 66: RE: https://github.com/google/fonts/pull/8496#issuecomment-2478470250

.github/workflows/test.yaml
```diff
@@ -111,9 +111,9 @@ jobs:
 
       - name: Upload check results
         if: steps.check_files.outputs.files_exists == 'true'
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: qa
+          name: qa-diffenator
```
Juneezee, line 116: RE: https://github.com/google/fonts/pull/8496#issuecomment-2478470250

.github/workflows/test.yaml
Juneezee, line 1: @m4rc1e Only this workflow is affected by the breaking change. I have also updated both artifact names to ensure they are unique.

m4rc1e: Thank you very much!

In order to test this pr, we need to trigger the action to run on a family. Would you mind adding another commit that changes https://github.com/google/fonts/blob/main/ofl/abel/METADATA.pb#L4 from `SANS_SERIF` to `SERIF`. If we're happy with the results, we'll delete the test commit.

Juneezee: @m4rc1e Yup! Pushed a test commit 388f1bc589dedc64d1570d6c242604deec58819c

Juneezee: It looks like the secrets are not available in my forked repository.

m4rc1e:
We're also dealing with a failing CI due to a fontbakery error so this isn't your fault.

I'm happy to merge this and finish it off on Monday. From what I've read in the upload repo, I'll need to consolidate the zips in a separate job, https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact

I'm not going to merge it today since I adhere to "no deploy Fridays" for critical stuff ;-)

Juneezee:
Rebased.

Juneezee:
Rebased.

Juneezee:
https://github.com/google/fonts/pull/8970

m4rc1e:
@Juneezee I'm very sorry for forgetting about this PR and overwriting it with #8970.

PR #8526, opened 2024-11-20
Create tags_metadata.csv

nyshadhr9:


m4rc1e:
@nyshadhr9 Any reason why we need a separate file? Since I have been developing the app to input the tagging data, I would like to be involved in these decisions.

nyshadhr9:
This is a list of all the tags and their display names. Going forward lets add new tags here and that helps us internally to run checks like not having an invalid or misspelled tag for a family. This could be expanded further to add additional metadata about tags but we don't have any concrete plans yet.

I also added a tags_ui_details file and thats specifically for the frontend catalog. Having it in the github repo helps in surfacing tags on dev frontend. No action needed there.

m4rc1e:
Thanks for the explanation. I'll update https://google.github.io/fonts/tags.html so it'll also modify this file if a user adds a tag.

PR #8661, opened 2024-11-27
Playwrite ZA Guides: Version Version 1.003 added

vv-monsalve:
Taken from the upstream repo https://github.com/TypeTogether/Playwrite at commit https://github.com/TypeTogether/Playwrite/commit/0bd52a3a13b6f3492820c746fa4bfa5196005306


vv-monsalve:
> 🔥 FAIL METADATA.pb should contain at least "menu" and "latin" subsets.

- We use only the menu subset to avoid the subsetter issues identified with the PW fonts.

PR #8662, opened 2024-11-27
Playwrite AR Guides: Version 1.003 added



vv-monsalve:
Taken from the upstream repo https://github.com/TypeTogether/Playwrite at commit https://github.com/TypeTogether/Playwrite/commit/0bd52a3a13b6f3492820c746fa4bfa5196005306.

Resolves #8541
## PR Checklist:

- [ ] `minisite_url` definition in the METADATA.pb file for commissioned projects
- [ ] `tags` are added for NEW FONTS
- [ ] `primary_script` definition in the METADATA.pb file for all projects that have a primary non-Latin based language support target
- [ ] `subsets` definitions in the METADATA.pb reflect the actual subsets and languages present in the font files (in alphabetic order). For **CJK fonts**, only include one of the following subsets `chinese-hongkong`, `chinese-simplified`, `chinese-traditional`, `korean`, `japanese`.
- [ ] Fontbakery checks are reviewed and failing checks are resolved in collaboration with the upstream font development team
- [ ] Diffenator2 regression checks for revisions on all projects that are currently in production
- [ ] Designers bio info have to be present in the designer catalog (at least an issue should be opened for tracking this, if they are not)
- [ ] Check designers order in metadata.pb, since the first one of the list appears as “principal designer”
- [ ] Social media formatted visual assets for all new commissioned projects in the Drive directory, communicate with the repository Maintainer so that they can push this content to the Social Media tracker spreadsheet
- [ ] Social media content draft for all new commissioned projects in the Drive directory and Social Media tracker spreadsheet, communicate with the repository Maintainer so that they can push this content to the Social Media tracker spreadsheet

PR #8689, opened 2024-12-04
PW subsets fix

vv-monsalve:
HI @emmamarichal, this PR fixes the subsets for the CO and AR families to make them only `menu` as expected for these fonts.


emmamarichal:
great thank you @vv-monsalve!

vv-monsalve:
The changes introduced have reached Sandbox, and here will be moved through the pipeline by te original PRs #8662 and #8669.
I'm removing this PR from the traffic board now.

Issue #8416, opened 2024-10-24
Tracking primary_script PRs


emmamarichal:
## SEE UPDATES IN THE COMMENTS BELOW

@simoncozens 
cc @m4rc1e @vv-monsalve 
I put here the issues with the recent primary_script updates.

**Important!** 
- https://github.com/google/fonts/pull/8097
We have to remove `ofl/phetsarath` and `ofl/ponnala` from the push lists (never went in dev-sandbox or sandbox)

_____

### **DEV-SANDBOX**

**Working / On the to_sandbox list:**
- https://github.com/google/fonts/pull/8083 
- https://github.com/google/fonts/pull/8090
- https://github.com/google/fonts/pull/8092
- https://github.com/google/fonts/pull/8094
- https://github.com/google/fonts/pull/8095

**Not working in dev-sandbox / Blocked:**
- https://github.com/google/fonts/pull/8086
Gidugu:` primary_script: "Telu"`

- https://github.com/google/fonts/pull/8091
Mogra: `primary_script: "Gujr"`

- https://github.com/google/fonts/pull/8096
Noto Sans Pau Cin Hau / Noto Noto Sans Phags-pa / Noto Sans Symbols 2: nothing changed, but had a primary script working on the API. Since there are no differences, I guess it didn't work?
Noto Serif Khitan Small Script: `primary_script: "Kits"`
Noto Sans SignWriting: `primary_script: "Sgnw"`

- https://github.com/google/fonts/pull/8098
Poppins: p`rimary_script: "Deva"`

**New fonts, but are working:**
notosanstamilui 
notosansteluguui 
notosansthaiui

_____

### **SANDBOX**

**Working / On the to_production list:**
- https://github.com/google/fonts/pull/8089

All the "Hang" ones should be solved by: https://github.com/google/fonts/pull/8185

**Not working in dev-sandbox / Blocked:**
- https://github.com/google/fonts/pull/8084
Black And White Picture: `primary_script: "Hang"`
Black Han Sans: `primary_script: "Hang"`

- https://github.com/google/fonts/pull/8085
Cute Font: `primary_script: "Hang"`
Dokdo: `primary_script: "Hang"`
Dongle: `primary_script: "Hang"`

- https://github.com/google/fonts/pull/8087
Gowun Batang: `primary_script: "Hang"`
Gowun Dodum: `primary_script: "Hang"`
Gugi: `primary_script: "Hang"`
Hahmlet: `primary_script: "Hang"`
[...]

PR #8692, opened 2024-12-05
[Nanum Myeongjo] Hotfixed fonts to add a space character

simoncozens:
As per https://github.com/google/fonts/issues/8416#issuecomment-2520870616, Nanum Myeongjo is showing tofu on the Google Fonts front end because it has no space character. In normal usage, this is fine because the `.notdef` character is blank, so that can be used as a space. But on the frontend, we use a fancy notdef fallback font to explicitly show tofu, so the ".notdef as space" trick doesn't work.

The solution is to add an explicit space glyph. We don't have the sources for this font, so we hotfix it with some FontTools hackery. Hotfix script added in case we need to do this again with updated font binaries.


emmamarichal:
Thanks @simoncozens!
Could you bump the font version? 2.030 -> 2.031

I want to make sure this fail isn't a problem in our case: 
```
🔥 FAIL Font contains glyphs for whitespace characters?
Whitespace glyph missing for codepoint 0x00A0.
```


Also, diffenator2 doesn't work:

```
fonts marichalemma$ diffenator2 diff -fa /Users/marichalemma/Downloads/nanummyeongjo/fonts/NanumMyeongjo-Regular.ttf -fb /Users/marichalemma/Downloads/nanummyeongjo/fonts/before/NanumMyeongjo-Regular.ttf 
[2/2] _diffenator "{'fonts_before': ['...gjo/fonts/NanumMyeongjo-Regular.ttf'}"
FAILED: out/Regular 
_diffenator "{'fonts_before': ['/Users/marichalemma/Downloads/nanummyeongjo/fonts/before/NanumMyeongjo-Regular.ttf'], 'fonts_after': ['/Users/marichalemma/Downloads/nanummyeongjo/fonts/NanumMyeongjo-Regular.ttf'], 'diffbrowsers': True, 'diffenator': True, 'out': 'out', 'imgs': False, 'styles': 'instances', 'characters': '.*', 'user_wordlist': None, 'filter_styles': 'Regular', 'font_size': 28, 'pt_size': 20, 'threshold': 0.9, 'precision': 28, 'no_words': False, 'no_tables': False, 'diffenator_template': '/Users/marichalemma/Google/env/lib/python3.12/site-packages/diffenator2/templates/diffenator.html', 'command': 'diff', 'diffbrowsers_templates': [], 'debug_gifs': False, 'kwargs': {'no_diffenator': False, 'no_diffbrowsers': False}, 'no_diffenator': False, 'no_diffbrowsers': False, 'coords': 'wght=400', 'old_font': '/Users/marichalemma/Downloads/nanummyeongjo/fonts/before/NanumMyeongjo-Regular.ttf', 'new_font': '/Users/marichalemma/Downloads/nanummyeongjo/fonts/NanumMyeongjo-Regular.ttf'}"
 64%|██████▍   | 7657/11880 [00:02<00:01, 3121.52it/s]
Traceback (most recent call last):
  File "/Users/marichalemma/Google/env/bin/_diffenator", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/marichalemma/Google/env/lib/python3.12/site-packages/diffenator2/_diffenator.py", line 110, in main
    diff.diff_all()
  File "/Users/marichalemma/Google/env/lib/python3.12/site-packages/diffenator2/_diffenator.py", line 39, in diff_all
    self.diff_words()
  File "/Users/marichalemma/Google/env/lib/python3.12/site-packages/diffenator2/_diffenator.py", line 57, in diff_words
    self.glyph_diff = test_fonts(
                      ^^^^^^^^^^^
  File "/Users/marichalemma/Google/env/lib/python3.12/site-packages/diffenator2/shape.py", line 51, in test_fonts
    glyphs = test_font_glyphs(font_a, font_b, threshold=threshold, font_size=font_size)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/marichalemma/Google/env/lib/python3.12/site-packages/diffenator2/shape.py", line 72, in test_font_glyphs
    pc, diff_map = differ.diff(g)
                   ^^^^^^^^^^^^^^
  File "/Users/marichalemma/Google/env/lib/python3.12/site-packages/diffenator2/renderer.py", line 217, in diff
    img_a, left_a = self.renderer_a.render(string)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/marichalemma/Google/env/lib/python3.12/site-packages/diffenator2/renderer.py", line 57, in render
    return self.render_text_cairo(text)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/marichalemma/Google/env/lib/python3.12/site-packages/diffenator2/renderer.py", line 75, in render_text_cairo
    left_edge = (min(0, orig_bounds[0], orig_bounds[2]) - self.margin) * scaleFactor
                        ~~~~~~~~~~~^^^
TypeError: 'NoneType' object is not subscriptable
ninja: build stopped: subcommand failed.
```

simoncozens:
This bumps the version and adds the 0xA0 character.

emmamarichal:
Thank you @simoncozens!

Just want to confirm with you that we have the right name table since it changed a bit from the API version:
Screenshot 2024-12-11 at 15 24 35

simoncozens:
This is fine. It looks like the subsetter is stripping out a bunch of name table entries for different platform/encoding/language combinations and only keeping the most common one (Windows/Unicode/English); you will see the same thing if you compare the API version and the `main` branch version.

emmamarichal (approved):

coreyvan:
This change seemingly breaks the English representation of this font. There's wildly different spacing between all the words.

image

Do you have any advice on where to submit feedback on this problem?

cre8tivediva:
I can confirm the spacing issue. I've just encountered this. Hoping this is resolved soon as we have some tickets already on it. Thanks in advance!

KamilKeski:
> This change seemingly breaks the English representation of this font. There's wildly different spacing between all the words.
> 
> image
> Do you have any advice on where to submit feedback on this problem?

Bump.  This is affecting us as well.  We use this font in almost all of our business documents.  Would be nice to be able to select a previous version.

simoncozens:
The reason this is happening (and I know this doesn't directly help) is that the font literally *didn't have a space at all* before, so your browsers would have been taking a fallback space glyph from another family. That would look right for Latin based languages (but broken for Korean), and it's arguably the wrong way for a font to operate.

I thought that em space was the right spacing for Korean text - I thought that because my experience of Korean comes from CJK fonts which have em space to make the other scripts work - but looking at other Korean fonts, it should be more like 200-250 as you say. I'll modify the hotfix and PR again.

cre8tivediva:
Would there happen to be an update on this or possibly on another thread? We're receiving quite a number of support tickets on this now. Our next step might be to have them replace it all together. Thanks in advance.

coreyvan:
@cre8tivediva a hotfix was merged for this a few hours ago. @simoncozens do you know how long it generally takes for it to be available on fonts.google.com?

simoncozens:
I'd say a couple of weeks, to be honest, to go through staging and into production.

PR #8693, opened 2024-12-05
Update METADATA.pb

nathan-williams:
Specify primary language for Noto Serif Hentaigana (ja_Hira).

### DIFF
--- a/ofl/notoserifhentaigana/METADATA.pb
+++ b/ofl/notoserifhentaigana/METADATA.pb
@@ -44,6 +44,7 @@ source {
 }
 is_noto: true
 primary_script: "Hira"
+primary_language: "ja_Hira"
 sample_text {
   # The text is the traditional Japanese pangram poem "Iroha"
   # https://en.wikipedia.org/wiki/Iroha


emmamarichal:
@nathan-williams @simoncozens @aaronbell 

(all the screenshots come from the dev-sandbox)

I tested Hentaigana: no more latin, but the new sample text: great! 🎉 
But I just want to confirm some things with you:

### **1-** 
The two other fonts that contain "Hentaigana" in the name have a different sample text and look very different from the Noto version. The Noto string doesn't seem to be supported by Yuji family. But maybe there are several kind of Hentaigana? (it's very naive question, I know nothing about japanese fonts).

Screenshot 2024-12-13 at 10 10 00
Screenshot 2024-12-13 at 10 19 47

### **2-** 

Noto Serif Hentaigana only appear in the font list when `Japanese Hiragna` is selected, but with tofu.

Screenshot 2024-12-13 at 10 26 51

simoncozens:
OK, so background: Japanese has three writing systems - (Chinese) kanji, hiragana and katakana. Hiragana and katakana are phonetic (syllabic) alphabets, and their shapes are derived by simplifying a kanji character with the same sound. In modern Japanese, the kanji used to represent each sound is standarised, and so there is a standard way of writing each syllable. However, in Classical Japanese, people simpified *different* kanji to represent the same sound, and so there were many different ways of writing each syllable. These different, non-standard ways are called "hentaigana" ("kana alternates").

Yuji Hentaigana uses the normal Unicode codepoints for Hiragana, but instead of the modern shapes, they choose one  "kana alternate" for each codepoint; it's kind of like Montserrat Alternates - a font full of stylistic alternates for normal characters. But which alternate did they choose? It's kind of up to the Yuji designers.

Some users want to make sure they are using a *particular* alternate form for each kana so that they can encode Classical Japanese manuscripts the way they were written. ("I need to write the sound 'ni' as a simplified 兒, not a simplified 仁".) So Unicode encoded a *separate* set of codepoints: instead of HIRAGANA LETTER NI, you get HENTAIGANA LETTER NI-1 DERIVED FROM U+4E39, HENTAIGANA LETTER NI-2 DERIVED FROM U+4E8C, etc.

This is why in your question, Noto Serif Hentaigana has a different sample text - it uses complete different Unicode codepoints. Actually the sample text repeats the same sounds but shows off different alternates ("I-1 RO-1 HA-1 NI-1, I-2 RO-2 HA-2, NI-2, ...")

Because Noto Serif Hentaigana is designed to be used with Noto CJK, *it does not itself include any HIRAGANA codepoints*. It just contains the hentaigana alternates. So actually this PR is wrong :-) and that is why you get tofu when you choose Japanese Hiranaga - it doesn't actually support those codepoints at all.

emmamarichal:
Thank you very much @simoncozens for all these explanations! I'll save it for later ;)

chrissimpkins:
It sounds like we need a new front end specimen language definition to cover the new Hentaigana (**note: not Hiragana**) encodings that are being added in this family @nathan-williams

chrissimpkins:
Deleted my previous comment. It seems this tofu is actually WAI.  This family does not support the Hiragana encodings, only the Hentaigana extension encodings if I understand Simon's comment correctly.  Is the solution here to simply remove the Hiragana language definition so that the family does not show up with a Hiragana drop down filter?

simoncozens:
Correct.

PR #8590, opened 2024-11-25
Iansui: 1.003 added

aaronbell:
Taken from the upstream repo https://github.com/ButTaiwan/iansui at commit https://github.com/ButTaiwan/iansui/tree/41190e8094a9a5b47626c5d927daf0dcb226babf

New font project as part of the GF CJK Q4 onboarding effort. Resolves #7742. 

Some existing `FAILS` with FB that I do not believe are problematic at this time.
`com.google.fonts/check/glyph_coverage` - Font contains GF Core
`com.google.fonts/check/dotted_circle` - Font is primarily intended for CJK use
`com.google.fonts/check/case_mapping` - This is an upstream (Klee) issue that I don't think needs attending to for this project at this time.

PR #8697, opened 2024-12-06
Lists of the 6th of December

emmamarichal:
@chrissimpkins 

We should wait for Iansui https://github.com/google/fonts/pull/8590
Maybe we can solve the tofu later and push this one anyway


to_sandbox.txt
```diff
@@ -1,10 +1,50 @@
 # New
-
-# Upgrade
-
+ofl/agudisplay # https://github.com/google/fonts/pull/8487
+ofl/iansui # https://github.com/google/fonts/pull/8590
```
chrissimpkins, line 3: We'll move Iansui into sandbox and block there. We need to integrate the symbols2 subset definition into the production environment before this family moves into production.  

@rsheeter @nyshadhr9

chrissimpkins (approved):

PR #8748, opened 2024-12-18
Update families.csv

evanwadams:
Changed some of the tag values for /Seasonal/Lunar New Year from Zhao and Qunxin's input.  More emphasis on festive and bolder.


chrissimpkins:
Evan confirmed that this PR was manually added to sandbox outside of the to_sandbox list process.  I updated the project board data accordingly.

PR #8772, opened 2024-12-23
42dot Sans : v1.00 added

aaronbell:
Taken from the upstream repo https://github.com/42dot/42dot-Sans at commit https://github.com/42dot/42dot-Sans/commit/d23e87fe44d5b4f5afaa9dca9d5926d7c414d625


aaronbell:
Updated metadata to resolve FAILS. 

Regarding the font-specific fails:
🔥 FAIL Check font follows the Google Fonts CJK vertical metric schema
This font was intended to be metrics-compatible with Pretendard, so the metrics are set. Additionally, I believe we need to vary Korean vertical metrics from the GF CJK standard.

🔥 FAIL Check family name for GF Guide compliance.
🔥 FAIL Make sure family name does not begin with a digit. 
I'm not exactly sure what the origin of this is, but I couldn't find an application that failed with the numeric name.

m4rc1e:
@aaronbell I remember Windows didn't like families that started with numerals. Is this still the case?

aaronbell:
@m4rc1e it seemed to work fine in office and installed without issue in the system.  Maybe it was a legacy limitation?

aaronbell:
Sounds good to me! The company is pretty adamant about keeping the name as is, so that’s good news.

aaronbell:
@m4rc1e Checking in if this can get moved forward :)

emmamarichal:
I merge it! we can still check the dev-server then, to see if everything is good :)

PR #8809, opened 2025-01-07
ci: install cairo on ubuntu latest

m4rc1e:
Latest Ubuntu doesn't come with Cairo preinstalled so we have to do it manually.


m4rc1e:
Merging this since the CI fail is related to broken urls in the knowledge graph.

PR #8823, opened 2025-01-10
Playfair: Version 2.203 added



emmamarichal:
Taken from the upstream repo https://github.com/googlefonts/Playfair at commit https://github.com/googlefonts/Playfair/commit/a49f9f9dc9a2924641cefba3f5e33ac1c5aa4264.
## PR Checklist:


emmamarichal:
@m4rc1e
The fails are related to the unusual way to set up the opsz axis, but it has been decided that way a year ago, so we can go ahead.
We also have some fails and warns appearing because of different repo url used. But I think we don't have choice here, since the process is a bit unusual too. 

⚠️ WARN Each font in set of sibling families must have the same set of vertical metrics values.
-> This Warn seems too be a false positive, please confirm! (I didn't have it locally)

PR #8965, opened 2025-01-17
Bitcount Prop Double Ink: Version 1.001 added



yanone:


chrissimpkins:
Closing so that we can update the family to show foreground color at default axis locations

emmamarichal:
Updated

PR #8983, opened 2025-01-22
Create python-package-conda.yml

ossnop:


m4rc1e:
Thank you for the PR but bundling this repo as a Conda package feels out of scope.

PR #8989, opened 2025-01-23
Martel: Version 1.002 added



emmamarichal:
Taken from the upstream repo https://github.com/SorkinType/Martel at commit https://github.com/SorkinType/Martel/commit/3ac0dc1d1a3ce71199a2637f736d568442480af0.
## PR Checklist:


m4rc1e:
Devanagari mark positioning seems worse. Same situation for the Latin as well.

mt

emmamarichal:
@EbenSorkin, any idea of what happened here?

EbenSorkin:
The anchors themsel ves seem to be placed correctly still. So I looked at this today and I saw that the Deva OT code in the source is made to update automatically. So I think it is likely that when Glyphs changed their preferred way of doing things for Deva from the time this was made Glyphs 2? ( 8 years ago?) it broke the source when used in the Glyphs 3 app.  We could either 1 - go back to older sources ( or fonts ) and collect the Deva OT code from them and uncheck the auto-update box. This would I think return the old function the fastest. Or 2) Find someone who works on Devanagaris regularly ( Simon? Yannone? ) and the source updated so that it plays nicely with the new Glyphs expectations. I think I know how to do 1 even though it's a bit new to me. It seems like a quick thing to check. What would you ( Marc and Emma ) prefer to do?

emmamarichal:
@EbenSorkin option 1 sounds good! It looks like we just have to copy/past old features + make a diff to see if everything is ok. Let me know once it's updated :)

EbenSorkin:
Ok 👍

EbenSorkin:
The shipped font uses OT code that refers to the glyphs by unicode not
'nice names' So we could manually swap the unicodes for nice names with
search and replace. Or replace the Deva glyphs with unicodes.

[image: image.png]

But either might take longer than desired.

Simon, do you already have a process to do either of these rapidly?

Also the shipping font only has Prefix and CCMP no DevalMatra class or
Nukta Feature. Presumably it's all just mashed into the above.

-e.

emmamarichal:
cc @simoncozens?

@EbenSorkin the image you uploaded doesn't show up :/

simoncozens:
I've had a look at this. It is not an OT code issue specifically. There are a couple of different problems.

For example, UUE (ॷ) has automatic component alignment which is putting the UUE VOWEL SIGN onto the bottom anchor; it should be manually aligned.

Screenshot 2025-02-07 at 09 45 56

For tth_ra-deva uMatra-deva, everything looks right...

Screenshot 2025-02-07 at 09 52 44

But this is a [long-standing glyphsLib issue](https://github.com/googlefonts/glyphsLib/issues/756). glyphsLib falsely assumes that, effectively, any glyph name with an underscore it in must be a component glyph. When it constructs the mark attachment features, it looks for anchors named `bottom_1` or `bottom_2` and ignores any anchors named `bottom`.

To fix this issue, you must explicitly set the subcategory of any *base* glyphs with underscores in the name to  "Other", using the glyph info panel. Select the affected glyphs, press option-command-i, set subcategory and tick the box:

Screenshot 2025-02-07 at 09 54 33

emmamarichal:
@EbenSorkin, should I put that PR and postpone the upgrade, or do you have some time to try fixing this? thanks :)

EbenSorkin:
I’ll write the author and see if I can get a source file.

emmamarichal:
Needs some work, @EbenSorkin let me know if there are any updates, I'll re-open a fresh PR

Issue #9043, opened 2025-02-06
Athiti: incorrect position of U+201E (double low-9 quotation mark)


sebalis:
**Describe the bug**

In the Athiti font (all variants), the character U+201E (double low-9 quotation mark) is not correctly positioned. These is the opening quotation mark in German. The shape appears at the same height as the ones in U+201C and U+201D (left/right double quotation mark), while it should be near the baseline as the Unicode name suggests. The same problem does not occur with the single counterpart, U+201A (single low-9 quotation mark). On that basis I assume that this is an unintended mistake and not a misguided design choice similar to the one Microsoft made with Verdana, Tahoma and other fonts (where U+2019/U+201D, the right single/double quotation marks, are horizontally mirrored images of their left counterparts U+2018/U+201C, not 180-degree rotations as they should be, which also messes up German typographical quotation marks). This should be corrected and I am next going to submit a PR that tries that.

**To Reproduce**

Compare the glyphs U+201A and U+201E in Athiti, e.g. by placing them next to each other in a document, and add the other typographical quotation marks for reference:
‘ ’ ‚ „ “ ”

In Athiti, the fourth character (not counting the spaces) appears at the same height as the fifth and sixth.

**Expected behavior**

The middle two characters (when not counting spaces, numbers 3 and 4) should be near the baseline, as they are shown in the font used on this Github page.

PR #9044, opened 2025-02-06
Athiti font: correct positioning of U+201E (double low-9 quotation mark)

sebalis:
This PR is an attempt at correcting issue #9043 using FontForge. As I have no experience with FontForge, I will describe my process: I simply opened the TTF files for each variant, measured the height of the U+201A shape, and used the Transform dialog to move the U+201E shape to the same height. I then executed File → Generate Fonts to produce new versions of the TTF files. For every variant, there were two warnings. The first informed me about a convention that TrueType fonts should have an em-size that is a power of 2, while these fonts have 1000. The second was “Missing points at Extrema” for various glyphs (quotedbl, quotesingle, parenleft, parenright, braceleft, braceright, uni0E04, uni0E05, uni0E28, uni0E56, minute, second – so seemingly not the glyph that I was changing). I chose to ignore these.


sebalis:
I have tried to create a Google account in order to sign the CLA. But Google demands that I supply a mobile phone number, contrary to the information given in https://support.google.com/accounts/answer/1733224?hl=en, and that I refuse to do (“If you have a mobile phone, this info is optional but highly recommended”). I would otherwise be happy to sign the CLA as shown on https://cla.developers.google.com/about/google-individual. Please advise.

sebalis:
(force-pushed to correct author information, the previously mentioned problem with the CLA remains)

sebalis:
Thanks for the response. As I said above, I don’t really have the expertise to work on these things, this was a quick fix using FontForge. However, I have a strong suspicion that running this FontBakery check on the original font would produce the exact same results, because I didn’t consciously change any of these things. Could you try that, @emmamarichal?

emmamarichal:
@sebalis, yes, this is what I did (you'll find two fontbakery reports above).
For example, for ExtraLight in the current version on the API, there is 39 issues, whereas there are 47 for the version in this PR. It's maybe because of the upgrade of our tools. 
I'll open an issue for this upgrade and we will plan it later!

PR #9056, opened 2025-02-08
Update METADATA.pb

evanwadams:
Merriweather is converting to a VF.  It's Fallbacks need to include the additional axes of wdth and opsz.


nyshadhr9:
LGTM

PR #9066, opened 2025-02-11
Bitter: Version 3.021 added



yanone:


emmamarichal:
There is now an interpolation issue in `uni25C7`:

https://github.com/user-attachments/assets/7ee63f0a-ab4c-4182-8a2c-4dd24a87f735

Also, the apostroph position changed in `napostroph` and became a bit too spaced in my opinion:

https://github.com/user-attachments/assets/b46063f2-d7fd-4f1e-8437-2afb245565da

PR #9076, opened 2025-02-13
Readex Pro: Version 1.205 added



yanone:


emmamarichal:
@yanone nice upgrade!
just catch an interpolation issue:

https://github.com/user-attachments/assets/4eee17b7-74f8-4386-892b-0a9854591c15

yanone:
Will fix it

yanone:
Updated

PR #9127, opened 2025-02-28
update pushlists

m4rc1e:
cc @chrissimpkins

### DIFF
--- a/to_production.txt
+++ b/to_production.txt
@@ -11,6 +11,7 @@ ofl/buenard # https://github.com/google/fonts/pull/7573
 ofl/cormorantgaramond # https://github.com/google/fonts/pull/8593
 ofl/cormorantinfant # https://github.com/google/fonts/pull/8594
 ofl/englebert # https://github.com/google/fonts/pull/8688
+ofl/iansui # https://github.com/google/fonts/pull/9022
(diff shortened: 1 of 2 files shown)


to_production.txt
```diff
@@ -11,6 +11,7 @@ ofl/buenard # https://github.com/google/fonts/pull/7573
 ofl/cormorantgaramond # https://github.com/google/fonts/pull/8593
 ofl/cormorantinfant # https://github.com/google/fonts/pull/8594
 ofl/englebert # https://github.com/google/fonts/pull/8688
+ofl/iansui # https://github.com/google/fonts/pull/9022
```
chrissimpkins, line 14: New project that requires file size review before we push to Workspace editors

chrissimpkins (changes requested):
@m4rc1e Iansui (new project) was listed in the updates list. Will you please confirm that there are no other new projects in the updates list?  I'm going to look into file sizes with @aaronbell today to decide if we need to exclude this from the Workspace editors.  Guessing this is likely to be necessary until they have an opportunity to test with this CJK area family.

to_production.txt
chrissimpkins: Aaron confirmed that this is a single instance static format family that has a size of 9.2MB. This should be ok for Workspace.

PR #9147, opened 2025-03-05
Update families.csv Special Use

EbenSorkin:
Special Use


EbenSorkin:
Special use is divided into three categories : 
100 - Things that serve a purpose requiring expert consideration and which are not general use enough to be safe for AI to recommend. This is often because their use case is a small niche. These 100 ranked fonts may not be perfect but they reach a fairly high standard of design and usefulness. 
50 - These are designs that are flawed enough either in their spacing or glyph design or both that it would be embarrassing for AI to recommend them. Some of these designs could be fixed and leave this category to be recommendable. Some are too banal or dull to be saved but at the same are not unusable.
0 - These designs are as above but where the flaws are more intense. Some of these are also worth saving/improving. These tend to fonts that seem to have been casually or carelessly made or which are very visually noisy.

In many cases fonts in a given category could be argued to belong in the next one up or down if they might have been a 20 30 or a 70 or 80 in a more granular scale.

PR #9163, opened 2025-03-06
Correct Iansui's metadata

bestpika:
All modifications to the writing of this font are based on the standards of the "Ministry of Education of Taiwan," so it should be categorized as Taiwanese Chinese rather than Hong Kong Chinese.

### DIFF
--- a/ofl/iansui/METADATA.pb
+++ b/ofl/iansui/METADATA.pb
@@ -12,13 +12,16 @@ fonts {
   full_name: "Iansui Regular"
   copyright: "Copyright 2022 The Iansui Project Authors (https://github.com/ButTaiwan/iansui)."
 }
-subsets: "chinese-hongkong"
+subsets: "chinese-traditional"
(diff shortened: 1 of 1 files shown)


bestpika:
It seems that after signing the CLA, a rescan is needed, it's done.

aaronbell:
@m4rc1e I think all the TC fonts were changed to the Hong Kong subset, but honestly I don’t know why. I figured it was something to do with the GF backend systems, but I think it has other effects, like with how these fonts are showing up in workspace. I’m not even sure why we can only assign one CJK subset…

@chrissimpkins

chrissimpkins:
@emmamarichal is this proposal now outdated based on the latest changes in Iansui?

emmamarichal:
Yes, it has been done here: https://github.com/google/fonts/pull/9296/files

chrissimpkins:
Thanks for the prompt to make this change @bestpika

PR #9184, opened 2025-03-11
Adding 119 automatically generated `config.yaml` files, for font repos that have `.glyphs` files in their `sources/` directory.

felipesanches:
These files were auto-generated by the python script below, working on the git repo cache dir that results from running the [google-fonts-sources](https://github.com/googlefonts/google-fonts-sources) tool.

```python
import os
import sys

if len(sys.argv) != 3:
	sys.exit(f"usage: {sys.argv[0]} gfonts_dir cache_dir")
	
gfonts_repo = sys.argv[1]
cache_dir = sys.argv[2]

SOURCE_EXT = ".glyphs"
#SOURCE_EXT = ".ufo"
#SOURCE_EXT = ".vfb"
#SOURCE_EXT = ".designspace"

count = 0

for subdir, _, files in os.walk(f'{gfonts_repo}/ofl'):
	commit = None
	repo_url = None
	if "METADATA.pb" in files:
		for line in open(subdir+"/METADATA.pb",'r').readlines():
			if "commit" in line:
				commit = line.split("\"")[1].strip()
			elif "repository_url" in line:
				repo_url = line.split("\"")[1].strip()

	if commit and repo_url and "github.com/" in repo_url:
		repo_checkout = cache_dir + "/" + repo_url.split("github.com/")[1]
		if os.path.isdir(repo_checkout) and os.path.isdir(repo_checkout + "/sources") and not os.path.exists(repo_checkout + "/sources/config.yaml"):
			for repo_subdir, _, repo_files in os.walk(repo_checkout + "/sources"):
				source_files = [f for f in repo_files if f.endswith(SOURCE_EXT)]
				if not source_files:
					continue

				config_yaml_content = "sources:\n" + "\n".join([f"  - {sf}" for sf in sorted(source_files)])
				
#				print(subdir + "/config.yaml")
				open(subdir + "/config.yaml", "w+").write(config_yaml_content)
				print(f"[{repo_checkout}] {repo_url} {commit} {repo_subdir}\n")
				count += 1
				break

print(count)
```

### DIFF
--- a/ofl/aboreto/config.yaml
+++ b/ofl/aboreto/config.yaml
@@ -1,0 +1,2 @@
+sources:
+  - Aboreto.glyphs
--- a/ofl/akshar/config.yaml
+++ b/ofl/akshar/config.yaml
@@ -1,0 +1,2 @@
+sources:
+  - Akshar.glyphs
--- a/ofl/anekbangla/config.yaml
+++ b/ofl/anekbangla/config.yaml
@@ -1,0 +1,2 @@
+sources:
+  - AnekGurmukhi.glyphs
--- a/ofl/anekdevanagari/config.yaml
+++ b/ofl/anekdevanagari/config.yaml
@@ -1,0 +1,2 @@
+sources:
+  - AnekGurmukhi.glyphs
--- a/ofl/anekgujarati/config.yaml
+++ b/ofl/anekgujarati/config.yaml
@@ -1,0 +1,2 @@
+sources:
+  - AnekGurmukhi.glyphs
--- a/ofl/anekgurmukhi/config.yaml
+++ b/ofl/anekgurmukhi/config.yaml
@@ -1,0 +1,2 @@
+sources:
+  - AnekGurmukhi.glyphs
--- a/ofl/anekkannada/config.yaml
+++ b/ofl/anekkannada/config.yaml
@@ -1,0 +1,2 @@
+sources:
+  - AnekGurmukhi.glyphs
--- a/ofl/aneklatin/config.yaml
+++ b/ofl/aneklatin/config.yaml
@@ -1,0 +1,2 @@
+sources:
+  - AnekGurmukhi.glyphs
--- a/ofl/anekmalayalam/config.yaml
+++ b/ofl/anekmalayalam/config.yaml
@@ -1,0 +1,2 @@
+sources:
+  - AnekGurmukhi.glyphs
--- a/ofl/anekodia/config.yaml
+++ b/ofl/anekodia/config.yaml
@@ -1,0 +1,2 @@
+sources:
+  - AnekGurmukhi.glyphs
--- a/ofl/anektamil/config.yaml
+++ b/ofl/anektamil/config.yaml
@@ -1,0 +1,2 @@
+sources:
+  - AnekGurmukhi.glyphs
--- a/ofl/anektelugu/config.yaml
+++ b/ofl/anektelugu/config.yaml
@@ -1,0 +1,2 @@
+sources:
+  - AnekGurmukhi.glyphs
--- a/ofl/arima/config.yaml
+++ b/ofl/arima/config.yaml
@@ -1,0 +1,2 @@
+sources:
+  - Arima.glyphs
--- a/ofl/artifika/config.yaml
+++ b/ofl/artifika/config.yaml
@@ -1,0 +1,2 @@
+sources:
+  - Artifika.glyphs
--- a/ofl/average/config.yaml
+++ b/ofl/average/config.yaml
@@ -1,0 +1,2 @@
+sources:
+  - Average.glyphs
--- a/ofl/baloobhaijaan2/config.yaml
+++ b/ofl/baloobhaijaan2/config.yaml
@@ -1,0 +1,11 @@
+sources:
+  - Baloo2.glyphs
+  - BalooBhai2.glyphs
+  - BalooBhaijaan2.glyphs
+  - BalooBhaina2.glyphs
+  - BalooChettan2.glyphs
+  - BalooDa2.glyphs
+  - BalooPaaji2.glyphs
+  - BalooTamma2.glyphs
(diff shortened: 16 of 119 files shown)


cmyr:
One thing to consider: some fonts use `sources` and some fonts use `Sources` for the name of the sources directory. It looks like this PR is only checking for `sources`?

One other issue here is that one of the justifications for this mechanism is to support situations where multiple families are in the same repo. In this case. Imagine a repo with the following structure:

```
colinFont/
  - colinSans/
    - sources/
      - colinSans.glyphs
  - colinSerif/
    - sources/
      - colinSerif.glyphs
```

in _this_ repo, we would have `ofl/colinsans/METADATA.pb` and `ofl/colinserif/METADATA.pb`, both of which would include something like,

```
source {
    repository_url: "https://github.com/cmyr/colinfont"
}
```

... and the problem here is that the included config file isn't really enough for me to determine (without heuristics) where my sources are, since it isn't pointing to the correct subdirectory.

Without giving this a whole lot of thought, it feels to me like the easiest solution would be to add a field to the `sources` mapping in METADATA.pb that is something like `sources_dir`: "colinSans/sources"`.

felipesanches:
Future batches will come after this one based on removing those self-imposed restrictions, such as by accepting other source types (UFO, designspace, etc)

Also, @simoncozens also sent me the following messages as feedback on a private chat-group

**felipesanches**: Does it look reasonable?

**simoncozens**: For the fonts with single sources, yes. For fonts with multiple sources, you need to be more careful. Roman and Italic can be in a family's config file, but there are a bunch of cases where you have multiple different families in the same config. So you need to check the filename (or better, the family name in the source itself) and match it against the family name in the METADATA.pb for the directory you're iterating on.

Baloo Baijan is an example of scooping up all the sources in a repo when you only really want the source that applies to the family.

cmyr:
> > One thing to consider: some fonts use `sources` and some fonts use `Sources` for the name of the sources directory. It looks like this PR is only checking for `sources`?
> 
> Yes, that was intentional. As I'm currently processing 361 repos identified with google-fonts-sources, I decided to work on a smaller subset of those as a first batch.
> 
> So my self-imposed set of restrictions for this batch are:
> 
> * only repos with GlyphsApp source files
> * located on a directory called "sources" at the root of the repo (even though we know there are other possibilities)
> * only the OFL ones (which is almost everything, but I didn't bother this time looking up the apache and ufl dirs, event though that's very easy to do)

Sounds good (and google-fonts-sources also only looks at OFL) but we do still need some mechanism here to determine what the `sources` directory is, i.e. where we are supposed to pretend the config lives, within the repo (even if we know for this particular batch of additions that it's always at `$repo/sources`, I would prefer it to just always be explicit).

felipesanches:
> Without giving this a whole lot of thought, it feels to me like the easiest solution would be to add a field to the `sources` mapping in METADATA.pb that is something like `sources_dir`: "colinSans/sources"`.

While working on this PR yesterday, my reasoning was that:
* **fontc_crater** reads entries from its targets.json file and, for each entry, it checks out the repo at the specified revision (that's current implementation behavior)
* if looks for a config.yaml at **google/fonts** repo. And only runs the next step if the file is not found there (proposed at https://github.com/googlefonts/fontc_crater/issues/18)
* look for the config.yaml file at the place described on that `tagets.json` entry's `config_files` field. All such entries are relative to a **sources/** dir at roo of font repo. (that's current implementation behavior)


I also proposed at  that same issue (https://github.com/googlefonts/google-fonts-sources/issues/53) that when copying config.yaml from google/fonts into a checked out repo, it should be also placed on a sources directory at the root of the repo.

And the file paths listed on "sources" field inside the config.yaml file are relative to the location of that file (that's current gftools-builder behavior).

For multiple sets of sources, we'll have separate config-something.yaml files, where the "config" prefix is mandatory and the "something" part is any meaningful keyword. Here's an example of an existing repo that does that: https://github.com/CatharsisFonts/Cormorant

Screenshot From 2025-03-11 12-05-51

What do you think?

felipesanches:
> if looks for a config.yaml at google/fonts repo.

I guess that for this to be implemented in fontc_crater, it would be good to store the google/fonts subdir as a field on target.json (so google-fonts-sources should add those entries)

Instead of:

```
  {
    "repo_url": "https://github.com/googlefonts/Bruno-ace",
    "rev": "b6fea3bcf7edde0fb4f468502a1c81fdbba2e182",
    "config_files": [
      "config-regular.yaml",
      "config-small-caps.yaml"
    ]
  },
```

We could have this:
```
  {
    "repo_url": "https://github.com/googlefonts/Bruno-ace",
    "rev": "b6fea3bcf7edde0fb4f468502a1c81fdbba2e182",
    "config_files": [
      "config-regular.yaml",
    ],
    "gfonts_dir": "ofl/brunoace",
  },
  {
    "repo_url": "https://github.com/googlefonts/Bruno-ace",
    "rev": "b6fea3bcf7edde0fb4f468502a1c81fdbba2e182",
    "config_files": [
      "config-small-caps.yaml"
    ],
    "gfonts_dir": "ofl/brunoacesc",
  },
```

Or maybe an even better approach could be something like this:

```
  {
    "repo_url": "https://github.com/googlefonts/Bruno-ace",
    "rev": "b6fea3bcf7edde0fb4f468502a1c81fdbba2e182",
    "config_files": {
      "ofl/brunoace": "config-regular.yaml",
      "ofl/brunoacesc": "config-small-caps.yaml"
    }
  },
```

m4rc1e:
Apologies if this going slightly off topic but it may be good to know. The `gftools packager` can make a family pr to this repo by building the fonts from sources e.g

`gftools packager "Aboreto" ~/Type/fonts/ --build-from-source`

If the repo doesn't contain a config.yaml, it'll create a mock config file. It is pretty cool, https://github.com/googlefonts/gftools/blob/main/Lib/gftools/packager/build.py

felipesanches:
I'll soon submit a PR to replace this one with improvements based on the feedback received here.

Issue #9176, opened 2025-03-07
[tag tool] Save state in localStorage()


davelab6:
To prevent user silliness of closing tabs or browser crashes nuking minutes to hours of painstaking work before making a PR commit, it [should be simple](https://blog.logrocket.com/localstorage-javascript-complete-guide) to store the current state in localStorage()

Issue #8031, opened 2024-08-15
[tag tool] Feature requests


davelab6:
Small UX tweaks:

- [x] Use Roboto not Avenir (traitor!)
- [x] move the 'save csv' button to the right to separate it from the 'open pull request' button
- [x] give the 'add' button a blue background to raise it in the visual hierarchy as the default end action
- [ ] horizontally center align each row

Medium UX tweaks:

- [x] add the https://github.com/adobe-fonts/adobe-notdef font for fallback
- [x] add a simple autocomplete to the family name input

m4rc1e:
Added a line to break up the panel a bit into sections

<img width="457" alt="Screenshot 2024-08-27 at 11 35 36" src="https://github.com/user-attachments/assets/25920a06-3bee-49b5-9e10-05285ebe0a0a">

davelab6:
>  give the 'add' button a blue background to raise it in the visual hierarchy as the default end action

You ticked it but its not in your screenshot or live in https://google.github.io/fonts/tags.html

davelab6:
> Autocomplete is a great suggestion:

The weird placement to the left jumping to the right suggests to me the 'add' panel should be made center aligned, either still floating, or made fix full width

davelab6:
When loading the new Halloween tags to review today (https://google.github.io/fonts/tags.html?commit=refs%2Fheads%2Fmain&category=%2FSeasonal%2FHalloween) then some rows' fonts previews didn't load. Waiting and refreshing the page allowed them time to download. 

It appears this is because the page is loading 100s to 1,000s of fonts in full, with [tags.html#L216](https://github.com/google/fonts/blob/9b0024e5803fb451202a432eb614238ccd66a937/tags.html#L216):

    return "https://fonts.googleapis.com/css2?family=" + Family.replace(" ", "+") + "&display=swap"

This can use the `text=` API feature to massively reduce the filesize of the fonts, but I guess that requires a method to concatenate the characters in all the [Pangrams](https://github.com/google/fonts/blob/c8db839e73c32d7f18a01ab998e05650bd7567f9/.ci/tags.html#L92-L107), convert them to URL-encoded strings (https://www.w3schools.com/tags/ref_urlencode.ASP) and include them all in the CSS API URL.

davelab6:
The tool's export is now broken. The copy paste result starts with these 3 lines:

```
Family,Group/Tag,Weight
ABeeZee,/Expressive/Calm,"91.6
"
[...]

PR #9189, opened 2025-03-12
tags.html: further improvements

m4rc1e:
Users can now reverse a sort, https://github.com/google/fonts/issues/8031#issuecomment-2706938390 and they're able to load their previous session in case they close the browser tab, #9176


davelab6:
Session save/load looks like what I hoped for  :) Thanks!

PR #9157, opened 2025-03-05
Playpen Sans: Version 2.000 added



vv-monsalve:

PR #9207, opened 2025-03-13
Add designer: Vera Evstafieva

vv-monsalve:
Vera Evstafieva is one of the designers of [Playpen Sans (LGC)](https://github.com/google/fonts/pull/9157)

@m4rc1e I created this PR by collecting information from different sources, but there is still a chance the designer could provide it herself through the form. I put the `needs confirmation` label on it and will let you know when it's ready for revision.


vv-monsalve:
@m4rc1e I updated the PR with the information provided by the designer. This is now ready for review.

PR #9224, opened 2025-03-18
poppins: add devanagari subset.

m4rc1e:


m4rc1e:
Closing since this is covered in #9221

PR #9226, opened 2025-03-18
Cascadia Code: version 2407.024

aaronbell:
Taken from the upstream repo https://github.com/microsoft/cascadia-code at commit https://github.com/microsoft/cascadia-code/commit/56bcca3f2c1e4cb19458954f0e2bb4635960df91


aaronbell:
Sorry for the many fails :)

LMK if any are particularly concerning, and we can discuss!

PR #9235, opened 2025-03-19
Kanchenjunga: add kirat-rai subset

m4rc1e:
kirat-rai subset is already served.


m4rc1e:
Just realised this has been defined in Simon's subset update, #9221

PR #9232, opened 2025-03-19
Cal Sans: Version 1.000; ttfautohint (v1.8.4.7-5d5b) added



emmamarichal:
Taken from the upstream repo https://github.com/calcom/font at commit https://github.com/calcom/font/commit/b833fb1129ba8c62c29b1d9f70861c77204affe2.
## PR Checklist:

- [ ] `minisite_url` definition in the METADATA.pb file for commissioned projects
- [x] `tags` are added for NEW FONTS
- [x] `primary_script` definition in the METADATA.pb file for all projects that have a primary non-Latin based language support target
- [x] `subsets` definitions in the METADATA.pb reflect the actual subsets and languages present in the font files (in alphabetic order). For **CJK fonts**, only include one of the following subsets `chinese-hongkong`, `chinese-simplified`, `chinese-traditional`, `korean`, `japanese`.
- [ ] Fontbakery checks are reviewed and failing checks are resolved in collaboration with the upstream font development team
- [ ] Diffenator2 regression checks for revisions on all projects that are currently in production
- [x] Designers bio info have to be present in the designer catalog (at least an issue should be opened for tracking this, if they are not)
- [x] Check designers order in metadata.pb, since the first one of the list appears as “principal designer”
- [x] Social media formatted visual assets for all new commissioned projects in the Drive directory, communicate with the repository Maintainer so that they can push this content to the Social Media tracker spreadsheet
- [x] Social media content draft for all new commissioned projects in the Drive directory and Social Media tracker spreadsheet, communicate with the repository Maintainer so that they can push this content to the Social Media tracker spreadsheet

PR #9240, opened 2025-03-20
Cal Sans: Add designer: Mark Davis

emmamarichal:
https://github.com/google/fonts/pull/9232


m4rc1e:
failing url check is fine.

PR #9255, opened 2025-03-24
test rust ci

simoncozens:
- **Attempt to use Rust QA tooling**
- **Test pushster**


simoncozens:
Pushed to wrong repo.

PR #9269, opened 2025-03-27
Update Anton families to display

tomasdev:
Per talk with @davelab6 these should not be categorized as sans-serif but rather display fonts. They don't look great for small body text.


davelab6 (approved):

davelab6:
They are Sans Serif Display;)

PR #9284, opened 2025-03-31
[Batch 1/4] port info from fontc_crater targets list

felipesanches:
Port most of the repo_url & commit info from fontc_crater's target.json file. (https://github.com/googlefonts/fontc_crater/commit/ee7a65d433882450efa1d8418ed746bf07c05642)

These correspond to all target.json entries with a single config_yaml each. There are also 100 entries with multiple config_yaml values that I'll submit later because they need more careful review.

(issue #2587)


felipesanches:
Thanks, @simoncozens ;-)

PR #9291, opened 2025-04-02
XL Lubrifont TC: 2.100 added

aaronbell:
Taken from the upstream repo https://github.com/NightFurySL2001/WD-XL-font at commit https://github.com/NightFurySL2001/WD-XL-font/tree/75c4a3bb2fffefa0f0aae470c2c0a73d6f79d0b9

Font is 10.3MB, single weight.


NightFurySL2001:
I wonder how many things did the dash broke to had to drop "WD-" here 🤔

aaronbell:
@NightFurySL2001 Oh, I just handwrote that. Don't recall why I didn't put the "WD-" in there.  It is still in the font.

NightFurySL2001:
I see. Metadata.pb also had the name dropped too so might need a fix.

emmamarichal (approved):

aaronbell:
@NightFurySL2001 So.... it turns out the hyphen does in fact break everything. Sigh. 
Would you prefer "WD XL" or "WDXL"?

NightFurySL2001:
I think without space is better, WDXL. (This assumption might need to be check in detail in the future though)

PR #9292, opened 2025-04-02
XL Lubrifont SC: 2.100 added

aaronbell:
Taken from the upstream repo https://github.com/NightFurySL2001/WD-XL-font at commit https://github.com/NightFurySL2001/WD-XL-font/tree/75c4a3bb2fffefa0f0aae470c2c0a73d6f79d0b9

Font size: 10.3MB, single weight


emmamarichal:
@aaronbell 

hyphen needs to be removed in the family name

Issue #9173, opened 2025-03-07
Iansui subsetting is not working properly


aaronbell:
**Describe the bug**

As reported by @ButTaiwan [here](https://github.com/ButTaiwan/iansui/pull/47#issuecomment-2705658064), it appears that the Google Fonts subsetter is stripping out crucial glyphs for Taiwanese Hokkien.  

**To Reproduce**

See [this codepen](https://codepen.io/Ki-n-tiong-T-n/pen/ogNwWoV)

![Image](https://github.com/user-attachments/assets/6dbddbdb-3f13-4870-99fe-fec1e8f0f43e)

(Ignore the issues with the accented glyphs as that is unrelated to this bug report). There are many Chinese characters for which we see fallback and appear not to render. However, I can verify that these glyphs are present in the font and render as expected on the GF website as well as in the font source / binary.

From the codepen:
<img width="214" alt="Image" src="https://github.com/user-attachments/assets/80d3e7ce-cc47-4a25-8dd6-388383aabfc0" />

From GF website:
<img width="266" alt="Image" src="https://github.com/user-attachments/assets/97257c09-5cc4-49e3-913b-9eb368c37099" />

ButTaiwan:
It looks like there is same problem in combining mark symbols too.

Google fonts subsetter puts `U+0301`, `U+0300`, `U+0302` and `U+0304` correctly, but `U+030C`, `U+030D`, `U+0306` and `U+0358` are missing.

![Image](https://github.com/user-attachments/assets/43b07067-a0ff-495e-b85d-2384b6e009bc)

https://codepen.io/ButTaiwan/pen/YPzQBQg

Using Iansui installed in PC:

![Image](https://github.com/user-attachments/assets/c8ac7aec-5d96-47f4-8200-7e577a8ef599)

aaronbell:
We just pushed a new update to the slices for TW / HK which resolves this. It may still show up due to caching, but I've confirmed resolution on https://fonts.google.com/specimen/Iansui

<img width="2262" height="188" alt="Image" src="https://github.com/user-attachments/assets/28b88ee6-abc5-42cb-8b26-6063a4ad167f" />

<img width="1650" height="206" alt="Image" src="https://github.com/user-attachments/assets/26766820-7797-4175-a3cc-d74414d44cf8" />

<img width="1644" height="174" alt="Image" src="https://github.com/user-attachments/assets/e3295758-6a9f-4a21-a200-205b795eb89a" />

PR #9296, opened 2025-04-02
Update Iansui-Regular.ttf to 1.012

aaronbell:
Minor update to Iansui, now 1.1012

Font file updates:
- OO-poj glyphs now classified as spacing (these were rendering as non-spacing)

- added a post-production script:
apply GASP settings for rendering symmetric blur 
correct the vmtx / vhea table to align with the emBox.

Additionally, the language subset in the METADATA file has changed to use `chinese-traditional` instead of `chinese-hongkong`

### DIFF
--- a/ofl/iansui/METADATA.pb
+++ b/ofl/iansui/METADATA.pb
@@ -12,7 +12,7 @@ fonts {
   full_name: "Iansui Regular"
   copyright: "Copyright 2022 The Iansui Project Authors (https://github.com/ButTaiwan/iansui)."
 }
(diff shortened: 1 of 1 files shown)


emmamarichal:
Will this PR solve https://github.com/google/fonts/issues/9173? @aaronbell (so we can link the issue to the PR)

aaronbell:
@emmamarichal I don't believe so. My understanding is that the issues called out by #9173 are related to the subsetting lists and not to the font itself.

PR #9307, opened 2025-04-04
Bios for upcoming Ancizar fonts

yanone:


m4rc1e (changes requested):
http://www.unal.edu.co/ produces a 404 for me.

yanone:
I searched and found https://unal.edu.co/ which produces 403 instead. I'll ask by email for the correct URL.

PR #9309, opened 2025-04-04
push lists updated

m4rc1e:
cc @chrissimpkins


to_production.txt
```diff
@@ -1,5 +1,6 @@
 # New
 ofl/nationalpark # https://github.com/google/fonts/pull/9174
+ofl/playpensanshebrew # https://github.com/google/fonts/pull/9273
```
chrissimpkins, line 3: I think that Viv wants to ship all of PPS i18n at once

chrissimpkins (changes requested):
We need Viv's input before we merge this.

I tested the other families in WS and this is otherwise gtg.

m4rc1e:
@chrissimpkins Playpen is now removed in the to_prod file.

to_production.txt
chrissimpkins: Reverted in https://github.com/google/fonts/pull/9309/commits/8a73d9789b3ea3ae3c87db75b2ef4fe9dcd6d671

chrissimpkins (approved):

PR #9328, opened 2025-04-09
Figtree: Version 2.002 added

emmamarichal:


emmamarichal:
Updated

emmamarichal:
@m4rc1e In this update, I merged some of the main problematic glyphs, to avoid hinting and other rendering issues (see issues related)

PR #9337, opened 2025-04-11
Update Noto Znamenny Musical Notation metadata to classify as symbols

chrissimpkins:


nathan-williams (approved):

chrissimpkins:
Thanks Nate!

Any concerns about this Simon? We are trying to have this family show up under the Not Text filter on the front end.

Issue #4264, opened 2022-02-01
Add WD-XL Lubrifont SC/TC


NightFurySL2001:
**Font Project Git Repo URL:**

https://github.com/NightFurySL2001/WD-XL-font

Note: Please only add *Simplified Chinese (SC)* and *Traditional Chinese (TC)* version only; Japanese (JPS/JPN) version is a recent contribution and requires further testing.

**Super short description of the Font Family:**

WD-XL Lubrifont is a Chinese display font that emphasize on a compat yet welcoming experience, expanded for daily Chinese typography with professional typesetting in mind. Transcription system support includes Hanyu Pinyin, Taiwan Minnanyu Luomazi Pinyin Fang'an (or Tâi-lô), Pe̍h-ōe-jī and bopomofo (or zhuyin). This font is an update and expansion to [ZCOOL Qingke Huangyou](https://github.com/googlefonts/zcool-qingke-huangyou).

**Requirements:**

I understand that Google Fonts will publish only fonts that matches its requirements, and I can confirm the project meets them (by ticking the cases, or putting x between the square brackets in text mode):

- [x] The entire font project is licensed under the [OFL](https://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL)
- [x] There is no "Reserved Font Name" mention in the OFL license
- [x] Family name is unique according to namecheck.fontdata.com
- [x] If the font targets latin based written languages, it supports at least [GF Latin Core glyphset](https://github.com/googlefonts/glyphsets/blob/main/Lib/glyphsets/encodings/GF%20Glyph%20Sets/GF-latin-core_unique-glyphs.nam) - **Font does not target Latin based written languages, but support GF Latin Core**
- [x] I read and agree with the full [contributing requirements](https://github.com/google/fonts/blob/main/CONTRIBUTING.md)
- [x] The repository respect the [upstream repo structure](https://github.com/googlefonts/gf-docs/tree/main/Spec#upstream-repo-structure)
- [x] I will maintain the repository

RosaWagner:
Thanks for your submission. It looks promissing :) Would you consider exporting TTF files?
These is a page that can help you further to match GF expectations: https://googlefonts.github.io/gf-docs/Spec/

NightFurySL2001:
Thank you. I am not considering to export TTF files since AFDKO `makeotf` directly generates the `CFF ` table for OTFs, but if required I can add a conversion step using AFDKO `otf2ttf`.

[...]

PR #9380, opened 2025-04-24
WDXLLubrifontTC: Updating to remove the hyphen from the name

aaronbell:
Apparently including a hyphen in the font name causes Bad Things. So I've renamed "WD-XL Lubrifont" to "WDXL Lubrifont" per the author's preference.


emmamarichal:
fixes #4264

PR #9384, opened 2025-04-25
Huninn: change primary script Hani to Hant

emmamarichal:
@nyshadhr9 @nathan-williams: about Huninn issue, I'm doing a test with another primary_script. It has been validated by the designer team, it will maybe fix the tofu and rendering issue we have currently

cc @aaronbell @chrissimpkins

### DIFF
--- a/ofl/huninn/METADATA.pb
+++ b/ofl/huninn/METADATA.pb
@@ -18,7 +18,7 @@ subsets: "latin"
 subsets: "latin-ext"
 subsets: "menu"
 subsets: "vietnamese"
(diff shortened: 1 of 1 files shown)


chrissimpkins:
Let's try moving this into sandbox to confirm that the specimen text shows up as expected as the eng team works on the specimens for Hani

chrissimpkins (approved):

chrissimpkins:
@m4rc1e this is currently a test.  Let's discuss it next week before we move it further through the pipeline

PR #9398, opened 2025-04-30
Update Iansui METADATA to add localized name

aaronbell:
For testing purposes, we are adding a localized name to the display name of Iansui. 

@chrissimpkins 
@m4rc1e


chrissimpkins:
Thanks Aaron!

PR #9400, opened 2025-04-30
WDXL Lubrifont metadata.pb 

emmamarichal:
primary_script: "Hant" instead Huni


aaronbell:
Is there a reason this is still listed as "blocked"?

aaronbell:
Or should this one just be cleared because it appears to be superseded by #9463

PR #9412, opened 2025-05-02
Nokora: Version 9.000 added



### LINKED ISSUE #7941: Update Nokora font
I redesigned and updated Nokora to include a full set of 9 weights.

yanone:


yanone:
Updated

yanone:
Dave says since it's a voluntary submission, it's okay to ignore that the Latin doesn't even cover GF Latin Kernel.

The Latin bolder weights are spaced closer, which I think is an improvement because the color now matches the Khmer.

The check about uni17BE being zero-width is a false positive imo. It's defined as a mark in GDEF, so it's expected to have zero width.

yanone:
Updated

emmamarichal:
Hey @yanone! Yes the spacing diff is quite big but the font isn't so distributed, so I would be ok to merge it. 
However, I see a difference of weight, the new Light looks bolder, is it intentional?
https://github.com/user-attachments/assets/3dfcb33c-d7f5-4127-8a47-f525ccbbd45b

yanone:
> is it intentional?

Not by me

emmamarichal (approved):

PR #9422, opened 2025-05-07
Documenting Font Sources, 2025-05-06

felipesanches:
* Font families with full source info: 779 of 1901 (40.98% of GF Library)
* 49 more (2.58% more) than [previous batch](https://github.com/google/fonts/pull/9323)
* Progress is being tracked at https://docs.google.com/spreadsheets/d/1ao3k56FwQy6W0Ll5QbU_wpuKEvNPYcn8YyEU9_L8O4Q/edit?usp=sharing

### DIFF
--- a/ofl/notorashihebrew/METADATA.pb
+++ b/ofl/notorashihebrew/METADATA.pb
@@ -25,6 +25,7 @@ axes {
 source {
   repository_url: "https://github.com/notofonts/hebrew"
   commit: "02fcd18e295ebadcda7bcaa774c107538fe2a5ee"
+  config_yaml: "sources/config-rashi-hebrew.yaml"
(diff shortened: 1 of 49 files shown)


felipesanches:
@simoncozens, @davelab6, @rsheeter, @cmyr: 

## The batch in this PR is an important milestone

After merging this batch, there will be no font listed in fontc_crater's `targets.json` file that is not also documented (in terms of commit hash and repo url) at google/fonts `METADATA.pb` files. So it will enable us to move forward with https://github.com/googlefonts/google-fonts-sources/pull/59 (Metadata as sole source of truth)

felipesanches:
> * Font families with full source info: 779 of 1901 (40.98% of GF Library)

Also worth mentioning that fontc_crater's targets.json currently lists 618 families. So, after merging this batch, METADATA.pb files will include all of these families and also 161 additional ones (with much more to be gradually added afterwards).

Issue #8911, opened 2025-01-14
C/J vertical metrics requirements


aaronbell:
Over the years, Google has established [clear guidance for setting font vertical metrics](https://googlefonts.github.io/gf-guide/metrics.html) to ensure consistent rendering across a variety of environments (with web as a particular focus).

As a short summary, this guidance makes use of the two core vertical metric values, `sTypo` (sTypoAscender / sTypoDescender / sTypoLineGap) to establish typographic metrics (which are set for ideal line height for reading) in conjunction with `usWin` (usWinAscent / usWinDescent) to establish clipping metrics (where MS Word will not display any glyph paint that extends beyond these metrics), and setting the `USE_TYPO_METRICS` flag to ensure that they are used for such. The `hhea` metrics follow the sTypo (as the `USE_TYPO_METRICS` flag is set) values for consistency on Mac. If you need further explanation of these metrics and how to set them, I suggest reading the spec at the link above.

Interestingly, where the metrics requirements are responsive to the design of the typeface, in the case of CJK they are [locked to specific values](https://googlefonts.github.io/gf-guide/metrics.html#cjk-vertical-metrics) based on the [metrics established by Ken Lunde](https://github.com/source-foundry/font-line/issues/2#issuecomment-201830252) for the Source Han Sans / Noto CJK project: 

Attrib | Value | Example using 1000upm font
-- | -- | --
OS/2.sTypoAscender | 0.88 * font upm | 880
OS/2.sTypoDescender | -0.12 * font upm | -120
OS/2.sTypoLineGap | 0 | 0
hhea.ascender | Set to look comfortable (~1.16 * upm) | 1160
hhea.descender | Set to look comfortable (~0.288 * upm) | -288
hhea.lineGap | 0 | 0
OS/2.usWinAscent | Same as hhea.ascent | 1160
OS/2.usWinDescent | abs(value) of hhea.descent | 288
OS/2.fsSelection bit 7 (Use_Typo_Metrics) | Do not set / disabled | 0

In this case, since the `USE_TYPO_METRICS` flag is disabled, the `usWin` metrics act as both typographic vertical metrics and also clipping metrics.

The setting for the `sTypo` values is of particular note, and comes from the [MS OpenType Spec](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#stypoascender) requirements for the sTypo values of CJK fonts. 

[...]

Issue #7741, opened 2024-05-17
Add Designer and Article: LXGW WenKai TC fonts


emmamarichal:
@aaronbell 

Add designer: LXGW

- [ ] Create the designer profile: Form to send
- [ ] Collect images: Images specs to send
- [ ] Collect a longer text for description

chrissimpkins:
@aaronbell do you have this documenation from the upstream development team?

aaronbell:
It looks like we have a DESCRIPTION.en_us file that can be used for the article. 
This image can be included as well: https://raw.githubusercontent.com/lxgw/LxgwMarkerGothic/main/documentation/images/markerg-1.png

aaronbell:
I've provided the form to the designer if they want to fill out their profile

chrissimpkins:
We have not received additional information from the upstream design team. We will close this issue and welcome a bio if they want to provide it in the future

PR #9429, opened 2025-05-08
LXGW Marker Gothic : 1.001 added

aaronbell:
Taken from upstream repro https://github.com/aaronbell/LxgwMarkerGothic at fe8357007423a983e696d2d3ff545ac9bb1bb89e

This is an adaptation of LXGW Marker Gothic with the following changes:
- Cleared unicode from softhyphen
- Removed GPOS data from legacy marks
- Updated font metadata and font info to follow GF expectations
- Updated the vertical metrics to follow GF's https://github.com/google/fonts/issues/8911. 
- added GASP table set to full blur
- adjusted vhea table slightly to follow the center line

Single weight font, 3.2MB


emmamarichal:
Hi @aaronbell!

do you think you can add the image to the article? https://raw.githubusercontent.com/lxgw/LxgwMarkerGothic/main/documentation/images/markerg-1.png

About that issue: https://github.com/google/fonts/issues/7741 Are we good? Let me know if there are any images, descriptions or designer bio/avatar available so I can update all of these new fonts :)

aaronbell:
@emmamarichal Sure! Added it in. 

As for #7741, I pinged the form over to the designer, but they never replied. I think we should just move forward with the stub bio for the time being and if they want to expand it out at some point, they can get in touch.

PR #9430, opened 2025-05-08
Chiron Hei HK : 2.525 added

aaronbell:
Taken from upstream https://github.com/aaronbell/chiron-hei-hk at 9cba188cc74bd43ab7021e7a62d68ba970e71ebd

At present, the PADG axis has been stripped, but it would be fairly easy to re-add. 

*NOTE* Font files are 31MB / 33MB. Must be blocked from workspace.

Chiron Hei HK project created by a modified build script of the original. Upstream, the build system is stored in the "source" branch, and the output files are in the "release" branch. If you are planning to rebuild this in the future, please note that the build process is somewhat complex as it runs AFDKO first to generate OTF files, then uses those to generate the variable TTFs, then copies over the OT tables, then processes the TTFs to produce the GF version.

### DIFF
--- a/ofl/chironheihk/METADATA.pb
+++ b/ofl/chironheihk/METADATA.pb
@@ -1,0 +1,45 @@
+name: "Chiron Hei HK"
+designer: "Tamcy"
+license: "OFL"
+category: "SANS_SERIF"
+date_added: "2025-05-08"
+fonts {
+  name: "Chiron Hei HK"
+  style: "normal"
+  weight: 400
+  filename: "ChironHeiHK[wght].ttf"
+  post_script_name: "ChironHeiHK-ExtraLight"
+  full_name: "Chiron Hei HK ExtraLight"
+  copyright: "Copyright 2025 The Chiron Hei HK Project Authors (https://github.com/chiron-fonts/chiron-hei-hk)"
+}
+fonts {
+  name: "Chiron Hei HK"
+  style: "italic"
+  weight: 400
+  filename: "ChironHeiHK-Italic[wght].ttf"
+  post_script_name: "ChironHeiHK-ExtraLightItalic"
+  full_name: "Chiron Hei HK ExtraLight Italic"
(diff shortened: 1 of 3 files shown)


aaronbell:
@emmamarichal @m4rc1e I've opened this in draft mode as the upstream author hasn't had a chance to respond on the PRs as of yet, but if you wanted to take an initial pass, feel free. I don't expect any significant changes from what's on here.

aaronbell:
BTW, the ttx roundtrip fail here is somewhat complicated. When the fonts are built with AFDKO, a name table entry is generated for the stylistic sets. However, those entries are not being generated when the `ttf` is built as the names don't survive the conversion process of static `otf` to `UFO` (which is needed for fonttools to run). But in order to have the full OT functionality, the OT features are being copied directly from the AFDKO `otf` output file to the FontTools `ttf` output file as a post-production step. 

So the GSUB table has references to name table entries that don't exist. AFAIK this isn't really an issue, but it is an irritant.

aaronbell:
Ready for review!

emmamarichal:
Hi @aaronbell!

- There are some small interpolations issues: 
[ChironHeiHK-Italic[wght].pdf](https://github.com/user-attachments/files/20208774/ChironHeiHK-Italic.wght.pdf)
[ChironHeiHK[wght].pdf](https://github.com/user-attachments/files/20208775/ChironHeiHK.wght.pdf)

- Do we want some localized names for the family name and designer?

emmamarichal:
@aaronbell wait, forget my question about localised family names, we decided to not implement them yet!

aaronbell:
@emmamarichal Thanks for catching the interpolation error! The font data from this project originates in Source Han Sans and it looks like this bug is replicated there as well:
Screen Shot 2025-05-14 at 9 19 12 AM

So we'll need to report it to Adobe to fix upstream. Given that the interpolation issue is in the Korean glyph set and this font is intended for TC, might we just let it go through for now while we wait for Adobe to update?

emmamarichal (approved):

PR #9462, opened 2025-05-15
to lists of the 15th of May

emmamarichal:
@chrissimpkins I can add the to-delist with all the concerned projects. Let me know!

### DIFF
--- a/to_delist.txt
+++ b/to_delist.txt
@@ -1,6 +1,2 @@
-# will be replaced by UNAL Ancizar Sans and UNAL Ancizar Serif (upcoming PRs)
-https://fonts.google.com/specimen/Ancizar+Sans
-https://fonts.google.com/specimen/Ancizar+Serif
-
 # Will be replaced by https://github.com/google/fonts/pull/9413
 https://fonts.google.com/specimen/42dot+Sans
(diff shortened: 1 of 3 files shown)


to_delist.txt
```diff
@@ -0,0 +1,6 @@
+# will be replaced by UNAL Ancizar Sans and UNAL Ancizar Serif (upcoming PRs)
+https://fonts.google.com/specimen/Ancizar+Sans
```
chrissimpkins, line 2: Let's wait on these Ancizar delistings in the to_delist until we have the new families in production.

to_production.txt
```diff
@@ -1,15 +1,114 @@
 # New
+ofl/notoserifdivesakuru # https://github.com/google/fonts/pull/9047
+ofl/wdxllubrifonttc # https://github.com/google/fonts/pull/9380
```
chrissimpkins, line 3: Will need a file size assessment on this one before the prod push.  @garretrieger have you seen any flags that this CJK familiy is too large for Workspace?

chrissimpkins (dismissed):

to_production.txt
garretrieger: The sandbox presubmit didn't raise any errors for size so I think it should be OK.

PR #9463, opened 2025-05-16
Update primary script for CJK fonts

emmamarichal:
cc @aaronbell @chrissimpkins @garretrieger 

Update `primary_script`: `hani` -> `hunt` for **WDXL Lubrifont SC** and **WDXL Lubrifont TC**

This solved the specimen issue we had with **WDXL Lubrifont JP N**.

Screenshot 2025-05-16 at 10 14 16

But, @aaronbell, maybe we should use different primary scripts for these 3 different fonts?

### DIFF
--- a/ofl/wdxllubrifontjpn/METADATA.pb
+++ b/ofl/wdxllubrifontjpn/METADATA.pb
@@ -18,7 +18,7 @@ subsets: "latin"
 subsets: "latin-ext"
 subsets: "menu"
 subsets: "symbols2"
-primary_script: "Hant"
+primary_script: "Jpan"
 source {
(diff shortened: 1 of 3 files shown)


aaronbell:
@emmamarichal I updated the SC and JP ones to follow the conventions in the gflanguages list. Let's see how this works?

emmamarichal:
Great, thank you @aaronbell! I'm waiting for the jobs running, and I'll merge it

emmamarichal:
Hi @aaronbell! 

All seems to work well for JP N and SC, but not for TC (`sandbox`)

Screenshot 2025-05-22 at 14 27 24

I'll investigate and let you know if I find anything!

PR #9467, opened 2025-05-16
Update Chocolate Sans to 1.005

aaronbell:
This is an update to the Chocolate Sans project with improvements made upstream since the original submission, but also now building with a new system per the upstream maintainer's preference. 

This version features a number of changes (some vs the original source):
- Bug fixes and other modifications to to TC source
- Integration of better latin coverage.
- vertical metrics updated to follow our new standard. 
- renamed some very long glyph names:
* katakanasmallhu_katakanahiraganasemivoicedsoundmarkcombkana.ccmp
* katakanasmallhu_katakanahiraganasemivoicedsoundmarkcombkana.ccmp.1

- Adjust anchor in glyphs and add anchors to dotted circle ($25CC)
* uni0300
* uni0301
* uni0307
* uni030C

- uni9E44 - fixed overlap issue due to overlapping contours where is quadratic and one is cubic.

- Cleared overlapping segments / duplicate contours on various glyphs
* uni50C6
* uni5966
* uni79BB
* uni7E34
* uni8506
* uni8867
* uni9C40

- Switching metadata to point to the upstream repository.

### DIFF
--- a/ofl/chocolateclassicalsans/METADATA.pb
+++ b/ofl/chocolateclassicalsans/METADATA.pb
@@ -19,11 +19,11 @@ subsets: "latin-ext"
 subsets: "menu"
 subsets: "vietnamese"
 source {
-  repository_url: "https://github.com/aaronbell/ChocolateSans"
-  commit: "91c9c5179adc1b2af2de1e3972d91e5e62c7bbc4"
+  repository_url: "https://github.com/MoonlitOwen/ChocolateSans"
+  commit: "0446a76d969f9788b0b339a9923f6541699df903"
   config_yaml: "sources/project.yaml"
   files {
-    source_file: "OFL.txt"
+    source_file: "license.md"
     dest_file: "OFL.txt"
   }
   files {


NightFurySL2001:
I've updated the build system to use the config version number @ 1.005, please help to update here so that we can still do a reasonable version bump in the future.

aaronbell:
@NightFurySL2001 Done.

emmamarichal:
Hi @aaronbell, we have some significant v-metrics changes, is it intentional? 
Sorry, with all the CJK new fonts, I forgot which ones were concerned by the v-metrics discussions. In any case, we usually don't allow a lot of difference like this one 



https://github.com/user-attachments/assets/dee5b91a-e38b-479d-8380-ef2477fefdc8

NightFurySL2001:
Sorry, this was unintentional when updating the Latin part of the font. Will revert.

NightFurySL2001:
After investigating, there are two issues: one is the incorrect metrics in the FCP/UFO source file, and second is turning on useTypoMetrics flag (which had been closed before). @aaronbell can I comment out the useTypoMetrics flag to ensure consistency with the old font?

aaronbell:
Since Chocolate Sans was part of the priority list for this round of work, I’d assumed that we would update the metrics the follow the new model. These new metrics have been set in the upstream repro as well via the pre-production build script (FYI @NightFurySL2001)

IMO it would be better to update to keep consistency with the upstream repro.

But we can reverse that decision.

NightFurySL2001:
> These new metrics have been set in the upstream repro as well via the pre-production build script 

I don't think the metrics were included correctly in the source files as the FCP had incorrect vertical metrics after replacing the LCG with Source Sans 3. For now I have reverted the source files back to using 880/-120 for sTypo; no vertical metrics override is provided in the Python script except BASE table and fsSelection bit 7 which is turned on.

> IMO it would be better to update to keep consistency with the upstream repro.

what values should the upstream repro use? I'm thinking of keeping the old metrics before revamp where sTypo* reflects the ideo bbox and turn off bit 7.

emmamarichal:
Hi @NightFurySL2001 @aaronbell  We need to keep the values from the currently live version! Indeed, we have now a new model for CJK v-metrics, that will be apply to all new CJK projects. We decided to not upgrade the catalog and existing projects, it would break too many layers. So, in that case, I think we should keep exactly the same set up as the live on!

aaronbell:
Thanks @emmamarichal. In that case I think it also makes sense to set the same metrics for ThenKhung so there is similar performance between Cactus, Chocolate, and ThenKhung. I’ll look to update these values.

NightFurySL2001:
@aaronbell Cactus Serif and Chocolate Sans have different metrics, although both follow the old metrics guideline. Which one will be preferable?

NightFurySL2001:
@aaronbell  got it. This branch can update the file to revert the metrics, I'll check ThenKhung tomorrow.

NightFurySL2001:
@aaronbell both Chocolate Sans and Then Khung have been updated back to their original metrics. You can update the files here.

aaronbell:
@NightFurySL2001 Thanks! I'll do the update tomorrow morning so we can get these moving along.

NightFurySL2001:
@aaronbell please do a quick update from the latest build for some vertical advance fixes too. Should fixed everything with full support for complete CJK URO and Ext A blocks (20992 + 6592 = 27584 ideographs).

aaronbell:
Updated per @NightFurySL2001's request from commit 624ecb8064d34258383bcbb08521f9fa2af00124 in https://github.com/MoonlitOwen/ChocolateSans

emmamarichal (approved):

PR #9468, opened 2025-05-16
UoqMunThenKung: version 1.005

aaronbell:
Taken from upstream repro https://github.com/MoonlitOwen/ThenKhung at a9fd850b5af389a764742cf34839254bba7baf1f

Font size: 4.9MB


NightFurySL2001:
I've updated the build system to use the config version number @ 1.005, please help to update here so that we can still do a reasonable version bump in the future.

aaronbell:
@NightFurySL2001 Done.

PR #9478, opened 2025-05-18
Chiron Sung HK added: version 1.019

aaronbell:
Taken from https://github.com/chiron-fonts/chiron-sung-hk 
(source fork) at commit aff5b9c0be78f8c96d90b07acb50fa4102351b30.

*NOTE* Font files are 50.7MB and 51.9MB 😬

PR #9493, opened 2025-05-22
Lists of the 22th of May

emmamarichal:
@chrissimpkins Here you can find the lists!

@evanwadams, some notes for you:
- Savate (font + designers) needs to reach prod before the 18th of June
- [Chiron Sung HK](https://github.com/google/fonts/pull/9478) needs to stay in Sandbox once it's pushed in it
- 42dot has to be delisted, since it will be replaced by #9413
- We have an issue with https://github.com/google/fonts/pull/9380: Primary_script should be `Hant` and not `Hani`. It has been changed here, but not pushed, unlike the other ones: https://github.com/google/fonts/pull/9463/files

Let me know if you have any questions!

### DIFF
--- a/to_production.txt
+++ b/to_production.txt
@@ -1,26 +1,42 @@
 # New
+ofl/astasans # https://github.com/google/fonts/pull/9413
+ofl/huninn # https://github.com/google/fonts/pull/9228
 ofl/notoserifdivesakuru # https://github.com/google/fonts/pull/9047
-ofl/wdxllubrifonttc # https://github.com/google/fonts/pull/9380
+ofl/savate # https://github.com/google/fonts/pull/9453
+ofl/wdxllubrifontjpn # https://github.com/google/fonts/pull/9294
+ofl/wdxllubrifontsc # https://github.com/google/fonts/pull/9292
 
 # Upgrade
-# Deleted: apache/robotomono # https://github.com/google/fonts/pull/9282
 ofl/baskervville # https://github.com/google/fonts/pull/9375
+ofl/baskervvillesc # https://github.com/google/fonts/pull/9376
+ofl/fustat # https://github.com/google/fonts/pull/9417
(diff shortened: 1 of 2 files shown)


to_production.txt
```diff
@@ -1,26 +1,42 @@
 # New
+ofl/astasans # https://github.com/google/fonts/pull/9413
+ofl/huninn # https://github.com/google/fonts/pull/9228
 ofl/notoserifdivesakuru # https://github.com/google/fonts/pull/9047
-ofl/wdxllubrifonttc # https://github.com/google/fonts/pull/9380
+ofl/savate # https://github.com/google/fonts/pull/9453
+ofl/wdxllubrifontjpn # https://github.com/google/fonts/pull/9294
```
chrissimpkins, line 6: Need size checks before this moves to production

chrissimpkins:
> Chiron Sung HK needs to stay in Sandbox once it's pushed in it

@emmamarichal please block it on the repository side once you confirm that it reaches sandbox and let me know.  That will trigger the review of the file sizes to decide if we can push it.  Fonts eng will not be tracking this.

emmamarichal:
@chrissimpkins I removed Lubrifonts from the prod list, and also Huninn since it's a CJK too.

to_production.txt
chrissimpkins: Reached out to Evan today for a size check

chrissimpkins:
> @chrissimpkins I removed Lubrifonts from the prod list, and also Huninn since it's a CJK too.

Thanks!  Let's see if we can resolve it before we merge these list updates.  The change to deny list the families in Workspace is straightforward and fast, we just need the data on static instance sizes.

chrissimpkins (approved):

chrissimpkins:
@emmamarichal didn't get more information about the CJK families today.  Let's go ahead with the merge without those included. We'll add them back when we have more information about sizes.  Ty!

PR #9495, opened 2025-05-22
Add UNAL to display_name of Ancizar Sans + Serif

davelab6:
Per request of upstream organization

### DIFF
--- a/ofl/ancizarsans/METADATA.pb
+++ b/ofl/ancizarsans/METADATA.pb
@@ -30,6 +30,7 @@ axes {
   min_value: 100.0
   max_value: 1000.0
 }
+display_name: "UNAL Ancizar Sans"
 source {
   repository_url: "https://github.com/UNAL-OMD/UNAL-Ancizar"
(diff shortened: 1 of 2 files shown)


yanone (approved):

emmamarichal:
Hi @simoncozens, regarding the CI report:

```
🔥 FAIL Check for presence of an ARTICLE.en_us.html file (googlefonts/description/has_article)
🔥 FAIL Check font has a license. (googlefonts/family/has_license)
```
Here we only have a metadata.pb change, but it looks like the checks are looking for a license and article (the font already have them)

We have a false positive here I think:

```
💥 ERROR Check if the vertical metrics of a family are similar to the same family hosted on Google Fonts. (googlefonts/vertical_metrics_regressions)
💥 ERROR Error: Could not get remote style: Expected a string
```

davelab6:
@emmamarichal I suggest to ignore that for now, file it as an issue, and merge and push this through to prod :)

PR #9545, opened 2025-06-09
Updating Cactus Serif to version 1.005

aaronbell:
Fairly significant update to Cactus Classical Serif including expanded Latin language coverage, and bug fixes / improvements for the Chinese portion. 

Font is now building directly in the upstream repository (https://github.com/MoonlitOwen/CactusSerif) 
Thanks to @NightFurySL2001 for assistance in getting this project ready.


emmamarichal:
fontspector report doesn't show up because it's too long. I checked it locally, we don't have any ERROR or FATAL. Since the font has been onboarder a few months ago, we should be ok. I'll merge it and will check it in dev-server

PR #9561, opened 2025-06-12
intelonemono: Add new hinted statics

m4rc1e:


m4rc1e:
@emmamarichal We don't want to enable smart dropout for these fonts since they are hinted with VTT so ignore this fail for the statics.

emmamarichal:
🔥 FAIL Each font in a family must have the same set of vertical metrics values. (family/vertical_metrics)

-> It concerns win metrics, is it ok to ignore?

@m4rc1e

m4rc1e:
Good catch. Let me take a look.

emmamarichal (approved):

PR #9572, opened 2025-06-16
Update Bitcount Ink 1.001

simoncozens:
This is a rebuild of Bitcount Single Ink and Bitcount Prop Single Ink to place the zero (default) position of the element position axes in the middle of the designspace, so that the default instance shows something interesting rather than black pixels.


simoncozens:
Name table version FAIL is a problem, I'll fix. Others look like what we expect for this particular font.

emmamarichal:
@simoncozens ping me once the version issue is solved! I'll merge it right away so we can test it in dev-sandbox

PR #9596, opened 2025-06-22
Chiron GoRound TC v1.002

aaronbell:
Release v1.002
*NOTE* Variable font is 50MB

### DIFF
--- a/ofl/chirongoroundtc/METADATA.pb
+++ b/ofl/chirongoroundtc/METADATA.pb
@@ -1,0 +1,31 @@
+name: "Chiron GoRound TC"
+designer: "Tamcy"
+license: "OFL"
+category: "SANS_SERIF"
+date_added: "2025-06-21"
+fonts {
+  name: "Chiron GoRound TC"
+  style: "normal"
+  weight: 400
+  filename: "ChironGoRoundTC[wght].ttf"
+  post_script_name: "ChironGoRoundTC-ExtraLight"
+  full_name: "Chiron GoRound TC ExtraLight"
+  copyright: "Copyright 2025 The Chiron GoRound TC Project Authors (https://github.com/chiron-fonts/chiron-go-round-tc)"
+}
+subsets: "chinese-traditional"
+subsets: "cyrillic"
+subsets: "cyrillic-ext"
+subsets: "latin"
+subsets: "latin-ext"
+subsets: "menu"
+subsets: "vietnamese"
+axes {
+  tag: "wght"
+  min_value: 200.0
+  max_value: 900.0
+}
+source {
+  repository_url: "https://github.com/chiron-fonts/chiron-go-round-tc"
+  branch: "source"
+}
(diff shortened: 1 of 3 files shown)


aaronbell:
@emmamarichal 

> 🔥 FAIL Check font has a license. (googlefonts/family/has_license)

Not sure why it is getting this fail when `OFL.txt` is present. 

>🔥 FAIL Check base characters have non-zero advance width. (base_has_width)

The glyphs in question are combining (intended to enclose numbers), and should have a zero advance width. 

> 🔥 FAIL Ensure the font supports case swapping for all its glyphs. (case_mapping)

Unlikely to be fixed.

> 🔥 FAIL Ensure files are not too large. (file_size)

Can't do much about that. 

> 🔥 FAIL Glyph names are all valid? (valid_glyphnames)

Is this a major issue? 

> 🔥 FAIL Shapes languages in all GF glyphsets. (googlefonts/glyphsets/shape_languages)

Font is not intended for these languages.

> 🔥 FAIL Check family name for GF Guide compliance. (googlefonts/name/family_name_compliance)

CamelCase name. Should we whitelist?

> 🔥 FAIL Check font follows the Google Fonts CJK vertical metric schema (googlefonts/cjk_vertical_metrics)

These match the metrics of the existing Chiron Hei HK family. I do note that the vmtx / vhea tables are missing, which should probably be investigated. Will do. 

> 🔥 FAIL Font has all mandatory 'name' table entries? (googlefonts/name/mandatory_entries)

Will look to add.

aaronbell:
@emmamarichal 

Changelist:
- VMTX and VHEA tables are now being generated. 
- NID9 is now added, matching Chiron Hei HK.

emmamarichal:
Hi @aaronbell,
I'm taking a look, and our interpolation check tool found a lot of kink issues like this:

Screenshot 2025-07-09 at 15 45 42

I guess we can ignore them, since there are a lot. Can you confirm?

emmamarichal:
@simoncozens @m4rc1e it seems we have an issue with the CI. I thought it was because of a very long fontspector report, but [it already worked previously](https://github.com/google/fonts/pull/9596#issuecomment-2993917075), with version 1.1.0

aaronbell:
@emmamarichal the rounding is done via an automated tool, so I’m not surprised if there are issues. Not much that can be done at such a scale unfortunately.

emmamarichal:
@aaronbell Ok perfect, I'm just waiting for Simon or Marc answer about the CI issue, and we should be good!

simoncozens:
OK, this is happening because there's a fontspector ERROR googlefonts/cjk_vertical_metrics_regressions. And *that* is happening because we don't skip this check when the font is new! Will fix in fontspector.

emmamarichal (approved):

emmamarichal:
I've tested it locally, let's see how it appears in dev-sandbox

PR #9597, opened 2025-06-22
Add Bluecurve

Halfpenny37:


emmamarichal:
Hi @Halfpenny37!
Thank you for this PR, we appreciate your submission. However, it doesn't follow all our specs, and we also need to have a look before deciding if we onboard it or not. I close that PR, and open an issue.

We will let you, thank you!

PR #9639, opened 2025-07-06
Add BlueCurve

Halfpenny37:
Added sources


emmamarichal:
Hello @Halfpenny37, thank you for your submission!
Can you please open [an issue on the google fonts repo](https://github.com/google/fonts/issues) so we can have a look at the fonts, and then decide if we onboard it? This is the usual process :)

Here you can find the template to set up the repo: https://github.com/googlefonts/googlefonts-project-template

PR #9650, opened 2025-07-09
Initial push of TikTok Sans

emmamarichal:
TikTok Sans reached the API through the internal pipeline. This PR adds the font to the repo, but doesn't need to be pushed and tracked through the servers


emmamarichal:
@davelab6 @m4rc1e @nyshadhr9?
It's exactly the same folder we sent internally, and that has been pushed on the API (**v4.000**) I'm adding it to the repo, so I can then update the designers order in metadata.pb.
- Is that ok to let the fail about the font versioning `🔥 FAIL Version number has increased since previous release on Google Fonts? ` I don't want to push this again, since the font is the same. 
- I need to update `metadata.pb`, should I open a new PR if I only want `metadata.pb` to be pushed, and not the font?

PR #9656, opened 2025-07-10
Add a PAT to the auto-commit-action

simoncozens:
Now that `main` is protected (which is good!) our actions can't just auto-commit stuff to it anymore. Which is good, I guess, but a but annoying. So we need to do the commit as a user who has permission to bypass the branch protection rules. We have a `GH_TOKEN` secret defined already, which I *guess* is for admin, so that should do the job.

### DIFF
--- a/.github/workflows/report.yaml
+++ b/.github/workflows/report.yaml
@@ -55,6 +55,7 @@ jobs:
         name: Commit any updated data files
         with:
           file_pattern: ".ci/dashboard/src/data/"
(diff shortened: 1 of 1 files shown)


simoncozens:
This isn't the right fix, not least because `git-auto-commit-action` doesn't have a `token` parameter.

m4rc1e:
Thank you! Report action generated correctly so I'm merging. Will force rerun the report job now as well.

PR #9679, opened 2025-07-16
Update Google profile

davelab6:
@MariannaPaszkowska mentioned that the Google profile is empty, so I threw this together to start :)


emmamarichal (approved):

davelab6:
Tobias confirmed this is good to go :)

PR #9681, opened 2025-07-16
Add designer: Jakob Ekelund

emmamarichal:
Designer of Zalando 
https://github.com/google/fonts/pull

cc @RosaWagner


emmamarichal:
@RosaWagner The image was the same as KH Type, so I didn't add it. Let me know if you have a different one! :)

PR #9728, opened 2025-08-07
Science Gothic: Version 1.026 added



emmamarichal:
Taken from the upstream repo https://github.com/googlefonts/science-gothic at commit https://github.com/googlefonts/science-gothic/commit/b12c900788e515909dee58d14437379bf6022cb1.
### PR
- [ ] The PR is in the Traffic Jam Board (Project > Traffic Jam Board)
- [ ] It has the correct labels (upgrade, new font, specific language)
- [ ] The PR is linked to the corresponding issue

### QA
- [ ] Fontspector checks are reviewed and failing checks are resolved in collaboration with the upstream font development team
- [ ] Diffenator2 and 3 regression checks for revisions on all projects that are currently in production
- [ ] The article has no typos, images are properly linked and correctly sized
- [ ] OFL.txt contains the correct license URL (https://openfontlicense.org)

### Specifics

### Release


emmamarichal:
@tphinney @kateliev
We need to wait for the axis registration. The PR is done and in Sandbox, but I think we need to wait it goes live.
In the meantime, let's add the article here. I'm waiting for your confirmation for which images and text we can use

emmamarichal:
Updated

emmamarichal:
🔥 FAIL Validate defaults on fvar table match registered fallback names in GFAxisRegistry. (googlefonts/axisregistry/fvar_axis_defaults)
🔥 FAIL The defaul value CTRS:85 is not registered as an axis fallback name on the Google Axis Registry.
You should consider suggesting the addition of this value to the registry or adopted one of the existing fallback names for this axis:
-100, 0, 100 [code: not-registered]

@tphinney @evanwadams @rsheeter 

Do we need to update something in the new axis CTRS? Or is it something to solve in sources?

tphinney:
Unfortunately, I discovered a few days ago that the Contrast axis is… reversed. When I swapped to the new scale and axis I got it wrong.  :( (https://github.com/googlefonts/science-gothic/issues/373)

**The correct/new fallback for Science Gothic will be 18 after I finish reconfiguring the source files and we rebuild.**

But that value is ALSO not yet in the registry.

Why? One side effect of the contrast axis scale as defined: it is extremely likely that each font has a different default, and it is calculated.

emmamarichal:
Updated

emmamarichal:
Updated

PR #9757, opened 2025-08-26
[dashboard] Adjust fontc source key regex

simoncozens:
The way that fontc_crater results indicate the targets being built has changed; update our regex so that we recognise the sources again.


felipesanches:
Thanks, @simoncozens!

PR #9791, opened 2025-09-05
Sources info 2025-09-04

felipesanches:
@davelab6 and @simoncozens: This is a small set of source-repo info for 12 more families.

Overall progress still being tracked at [this spreadsheet](https://docs.google.com/spreadsheets/d/1ao3k56FwQy6W0Ll5QbU_wpuKEvNPYcn8YyEU9_L8O4Q/edit?gid=0#gid=0) and a while later it ends up reaching the [dashboard](https://google.github.io/fonts/source_check.html) (a bit after this PR is merged).

This PR was generated automatically by a script (a web crawler) that I am starting to develop to partially automate collection of sources. The goal here was to automatically fill-in the fields in the cases that do not require a human to review them. (even though I did review these ones to troubleshoot my initial crawler implementation)

The reasoning is that whenever the log message of the commit that last updated the font files points to an upstream repo commit that is still nowadays the last commit in such repo, then we can blindly trust the data. On the other hand, when additional commits were made on the upstream repo, we have to more carefully inspect the git history to discriminate whether those are additional upstream work done after the last onboarding of a family, or if those were also included in the onboarded files without updating the log message automatically generated by gftools-packager.

Even though the initial set of families that fit the description above was estimated last week as a set of 315 families, only 12 of them (the ones included in this PR) have their gftools-builder config file stored at `sources/config.yaml`.

In the next week, I'll make additional PRs for families (from this initial set of 315) that keep their config in other places, or with different filenames. To do so, I'll most likely update my helper script to actually git-clone the upstream repos and query them locally, instead of fetching github webpages and parsing them, for two reasons: (a) I don't want to get my IP blocked by github and (b) it is probably easier and faster to write a script to find yaml files with unusual names by navigating the filesystem locally, than to continue with the hacky webcrawling approach.

For that reason, I expect to be able to find larger sets of source-info on the next PRs.

My goal is to document **at least** an average of 15 families per day. So if the scripting approach does not yield better results within one or two weeks, I'll revert to doing things manually so that I can reach my self-assigned goal.

### DIFF
--- a/apache/kosugi/METADATA.pb
+++ b/apache/kosugi/METADATA.pb
@@ -19,6 +19,8 @@ subsets: "latin-ext"
 subsets: "menu"
 source {
   repository_url: "https://github.com/googlefonts/kosugi"
+  commit: "75171a2738135ab888549e76a9037e826094f0ce"
+  config_yaml: "sources/config.yaml"
   files {
     source_file: "fonts/ttf/Kosugi-Regular.ttf"
     dest_file: "Kosugi-Regular.ttf"


felipesanches:
Please also note that the actual content of the diffs in this PR is very very simple. But what is really important here is to know that the scripting is validating the `repository_url`, `commit` and `config_yaml` fields. So, even if the values were already set, now we can trust them. If values did not change, it means they were correct (but we did not know it). If values changed, it means they were wrong (or missing) and we fixed them.

felipesanches:
Also, every commit here includes (in their auto-generated log messages) references to the original commits that were parsed to gather the info, so we keep a track record for the future, in case we may want to double-check the data.

felipesanches:
After this PR is merged, I'll re-run the [google-fonts-sources](https://github.com/googlefonts/google-fonts-sources) tool and update [fontc_crater](https://google.github.io/fonts/fontc_crater.html)'s [`targets.json`](https://github.com/googlefonts/fontc_crater/blob/main/targets.json) file and open a PR for @cmyr to review.

My current plan is to try to do it (submit a PR with targets.json updates) weekly on Thursday nights or Friday mornings, in the hope that the [fontc](https://github.com/googlefonts/fontc) team can have a fresh crater run by Monday (or as soon as such weekly PRs are merged).

davelab6:
Thanks Felipe! I defer to Simon on the details but this all sounds good to me

felipesanches:
Closing this one, because it is being replaced by #9801 with identical contents.

PR #9801, opened 2025-09-05
Sources info 2025-09-04

felipesanches:
This PR is identical to #9791 that was opened yesterday. The only difference is that this one proposes merging from a branch of google/fonts repo instead of from a branch, and by doing so, enables access to ghaction secrets necessary to properly run all CI jobs.

### DIFF
--- a/apache/kosugi/METADATA.pb
+++ b/apache/kosugi/METADATA.pb
@@ -19,6 +19,8 @@ subsets: "latin-ext"
 subsets: "menu"
 source {
   repository_url: "https://github.com/googlefonts/kosugi"
+  commit: "75171a2738135ab888549e76a9037e826094f0ce"
+  config_yaml: "sources/config.yaml"
   files {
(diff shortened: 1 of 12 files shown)


simoncozens:
Thanks for this; my interpretation of `config_yaml` is that it should provide an override for the default value, `sources/config.yaml`. So in my view, most of this is not needed, but some tool which consumes this data needs to be smarter. It's not *wrong*, and I'm not opposed to merging it, but it feels redundant.

felipesanches:
Unfortunately, placing config files at "sources/config.yaml" does not seem to be extremely common. There are way too many different names people have used for this file, across almost 2 thousand font projects. So I'd prefer not to rely on a default value, but instead keep it explicit.

felipesanches:
Please merge this, so that I can move on to the next step (updating targets.json on fontc_crater)

felipesanches:
thanks!

PR #9824, opened 2025-09-11
google sans code: Update METADATA.pb

nyshadhr9:
Import sample_text changes which are currently used in sandbox/prod. Set the correct font category.


davelab6:
This is a no-op for onboarding since its to back-propagate current server state

PR #9865, opened 2025-09-23
drop families_new.csv and use families.csv again

m4rc1e:
Improves #9863


m4rc1e:
CI fails because the tagging data has errors. Will merge this PR then open another to fix these issues.

PR #9877, opened 2025-09-25
Fix compaction of fontspector database

simoncozens:


simoncozens:
CI fail is a gftools issue, not my bug.

PR #9882, opened 2025-09-26
Tags update first batch

emmamarichal:
Adds:

- [x]   SUSE Mono
- [x]   Tirra
- [x]   Nata Sans
- [x]   Sirivennela
- [x]   Story Script
- [x]   Vend Sans
- [x]  Alan Sans

https://fonts.google.com/share?selection.family=SUSE+Mono|Tirra|Nata+Sans|Sirivennela|Story+Script|Vend+Sans|Alan+Sans

### DIFF
--- a/tags/all/families.csv
+++ b/tags/all/families.csv
@@ -1,4 +1,3 @@
-
 42dot Sans,,/Quality/Concept,70
 42dot Sans,,/Quality/Drawing,70
 42dot Sans,,/Quality/Spacing,70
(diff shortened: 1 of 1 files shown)


davelab6:
We discussed this in chat and the large number of line changes are due to the tags sort order here becoming what the https://google.github.io/fonts/tags.html tool produces; the currently file was sorted by me with macOS `sort -u` when I was debugging a previous submission by Eben, so, that's why this is unusual.

PR #9909, opened 2025-10-09
Libre Baskerville: Version 2.005; added

emmamarichal:


m4rc1e:
@emmamarichal any reason why "fj" lig is now gone?

Screenshot 2025-10-14 at 15 14 27

emmamarichal:
@m4rc1e mmh, I had some trouble with ligatures for that font. I managed to solve the big part of issue by changing the order of feature, but apparently, it's not totally fixed... 
let me have a look!

emmamarichal:
Updated

m4rc1e:
Screenshot 2025-10-22 at 10 10 55

Ligs still broken

PR #9910, opened 2025-10-09
Zain: Version 2.000; ttfautohint (v1.8.4) added



emmamarichal:


emmamarichal:
I tried to PR it, since we decided to ignore v-metrics changes, but there are some other fails to solve - still in progress

emmamarichal:
Updated

emmamarichal:
@simoncozens, we agree here to make an exception here and change the v-metrics. However, I find an inconsistency in the fontspector report and between what diffenator tells me. Also, in fontspector, I only have FAIL on Regular, whereas all styles are affected (the change should be detected everywhere). Any idea?

Screenshot 2025-10-10 at 11 30 34

Screenshot 2025-10-10 at 11 30 43

Screenshot 2025-10-10 at 11 31 27

m4rc1e:
Here's a before and after of the v metrics:

rt

simoncozens:
Fontspector (and fontbakery) only check the regular style for metrics regressions. Maybe we could improve this, but it's not unexpected.

emmamarichal:
@simoncozens 

> Fontspector (and fontbakery) only check the regular style for metrics regressions. Maybe we could improve this, but it's not unexpected.

Okay, that makes sense. And in any case, if any of the other styles, other than Regular, change, we'll get the fail indicating that the v-metrics are not consistent across all styles. So I'm okay with keeping it that way. This Fail will make us check all the v-metrics in any case.

However, do you know why we have:
**🔥 FAIL** hhea Ascender is 869 when it should be 720 [code: bad-hhea-ascender]

But if I check diffenator, I see:
before: 900 and after: 869?

It should be:
**🔥 FAIL** hhea Ascender is 869 when it should be 900 [code: bad-hhea-ascender]

right?

simoncozens:
> However, do you know why we have:
> 🔥 FAIL hhea Ascender is 869 when it should be 720 [code: bad-hhea-ascender]

> But if I check diffenator, I see:
> before: 900 and after: 869?

> It should be:
> 🔥 FAIL hhea Ascender is 869 when it should be 900 [code: bad-hhea-ascender]

> right?

We have two different checks:

* Are the vertical metrics *correct* according to the current GF spec?
* Are the vertical metrics *consistent* with production?

In the incoming font, the hhea ascender is 869. According to the GF spec, it should be 720. But what we have in production is 900! So this font is both incorrect *and* inconsistent! (Oh well.)

PR #9914, opened 2025-10-14
Noto Sans Sinhala: Version 3.000 added



simoncozens:


simoncozens:
> Axis 'wdth' min value is out of range. Expected '62.5', got '75'. [code: axis-min-out-of-range]

Urgh. Interesting, debugging...

simoncozens:
Updated

simoncozens:
Updated

emmamarichal:
Looks good @simoncozens, thank you for that update!

Just to check:

- Is it normal for these shapes to touch each other? In other styles, they are separated

Screenshot 2025-10-15 at 13 49 34

- I saw this in the before/after, I prefer to check if it's ok!

Screenshot 2025-10-15 at 13 56 54

https://github.com/user-attachments/assets/1834a001-080d-4d29-8476-85c0ba50453c

simoncozens:
I think it's ok but @pathumego can you answer?

pathumego:
> Is it normal for these shapes to touch each other? In other styles, they are separated

No, that is a mistake that escaped us. They shouldn't touch, @simoncozens will you fix this on your end? 

> I saw this in the before/after, I prefer to check if it's ok!

Yes, this is okay :) @emmamarichal that specific string you are looking at is 'invalid' and would not occur in Sinhala. We added a feature to process rakar by themselves and hence the change. Out of curiosity, what is this test text? I see more invalid combinations there. Is it meant to text random edge cases

emmamarichal:
It's that list: https://github.com/googlefonts/diffenator2/blob/main/src/diffenator2/data/wordlists/Sinhala.txt but I don't know how it has been made, @simoncozens can answer better than me on that point.

Let me know once it's fixed, I can do a last review before merging :)

pathumego:
@simoncozens Should I fix this in my repo?

PR #9960, opened 2025-10-30
Update Malay (Arabic) text example

niskala5570:
https://github.com/google/fonts/issues/9959


emmamarichal:
Hi @niskala5570, thank you very much for that PR, but for all the language/script changes, we do it on the lang repo (and then merge changes here): https://github.com/googlefonts/lang

Please open an issue there so we can review it and make the change, thank you!

PR #10044, opened 2025-12-02
Update OFL.txt to use expected phrasing

nathan-williams:


davelab6 (changes requested):
I suggest holding this until we confirm with upstream

emmamarichal (approved):

aaronbell:
I believe we resolved this on the Eng side, and no change is necessary. Closing.

PR #10090, opened 2025-12-21
Noto Sans Bengali: Version 3.011 added



simoncozens:


emmamarichal:
@simoncozens looks good except an interpolation issue on `u0983`:
Screenshot 2026-01-07 at 15 23 21

+ is that fail ok?
🔥 FAIL Check samples can be rendered (googlefonts/metadata/can_render_samples)
Font NotoSansBengali[wdth,wght].ttf cannot render Mundari sample text: সোবেন হোড়োকো কে মনরংগ ওড়োও অক্তিয়ার কো রেঅ͚ পইতিবাবত রে জনম জোরোংগ্এতে অহ্ডানড ওড়োও বরাবরী [code: sample-text]

simoncozens:
Updated

simoncozens:
> FAIL The font has the following codepoints needed to render the specimen 16 sample text for language Mundari, but although the font contained all the codepoints for the sample text, tofu will still be produced because the codepoints do not appear in any of the subsets bengali, latin-ext, latin:
͚ (U+035A) [code: missing-subsetted]

Bah, even if we fix the tofu glyph, it won't get shown because there is no subset including it. Well, the font is good anyway.

PR #10256, opened 2026-02-20
Add Kaitag (`xdq`) language

alkaitagi:
This PR adds complete Kaitag language info (`xdq_Cyrl.textproto`). I did not find any formal specification, so I filled the fields after the existing languages. Please point out anything that isn't right.

One thing that might be off is `exemplar_chars.index`. The Kaitag alphabet contains digraphs which appear in `exemplar_chars.base`. Should `index` only list base characters?

### DIFF
--- a/ofl/allkin/METADATA.pb
+++ b/ofl/allkin/METADATA.pb
@@ -26,7 +26,7 @@ source {
   }
   branch: "main"
 }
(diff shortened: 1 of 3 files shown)


simoncozens:
Thanks for this - changes to the languages database should really go in googlefonts/lang, which is the source of the lang subtree. I've made an equivalent PR over there. (https://github.com/googlefonts/lang/pull/253)

PR #10322, opened 2026-03-05
Update families.csv (missing tags)

emmamarichal:
Adds missing tags for last new fonts @m4rc1e


emmamarichal:
@m4rc1e For variable fonts like Science Gothic, for now, I added tags according the style displaying on the main page

PR #10325, opened 2026-03-05
Remove redundant font entries from families.csv

emmamarichal:
Removed multiple entries for Ek Mukta, Noto Sans Arabic UI, Noto Sans Bengali UI, Noto Sans Devanagari UI, Noto Sans Gujarati UI, Noto Sans Gurmukhi UI, Noto Sans Kannada UI, Noto Sans Kharoshthi, and Noto.


m4rc1e:
Excellent! Thank you so much.

PR #10431, opened 2026-04-15
/Expressive/Loud VF tags

m4rc1e:


emmamarichal:
@m4rc1e 

### CI ERRORS:

**1. test_families_missing_tags — 8 families without tags**
These fonts exist in Google Fonts but have no classification tags: Akt, BJCree, Estedad, Finlandica Headline, Finlandica Text, M PLUS U, STIX Two Math, Strichpunkt Sans. They need to be tagged via the dedicated webapp.
-> I'll make a PR to add them

**2. test_no_duplicate_families — Duplicate entries in /Expressive/Loud**
Hundreds of families appear twice under the /Expressive/Loud category. This looks like a data import or merge that duplicated the entire category.
-> I updated the checks, but it doesn't seem to be working? https://github.com/google/fonts/pull/10489

**3. test_tag_vals_in_range — Values at 0.0 (out of range 1–100)**
Those same duplicated families have a value of 0.0 for /Expressive/Loud, which is invalid (expected range is 1–100). This is directly linked to the previous issue: the duplicate entries were created with a null value.
-> Can you confirm that we now allow 0 value so I can update the check too?

**4. test_tags_for_unknown_families — 2 families tagged but not found**
Finlandica and Saira Stencil One have tags but no longer exist in the sandbox metadata — likely renamed or removed fonts (Finlandica was split into Headline/Text variants, and Saira Stencil One was renamed to Saira Stencil).
-> They just have been delisted, I'll remove them

PR #10476, opened 2026-04-27
Upgrade Arimo: OFL release

simoncozens:
This includes the new release of Arimo from the googlefonts/ repo, and moves from Apache to OFL.


emmamarichal:
There is a small interpolation issue, but since that's not the focus of this quick update, I'm fine with merging the PR and updating that later! @simoncozens 

Screenshot 2026-04-29 at 10 19 50

simoncozens:
Agreed. It was there already. Thanks!

PR #10523, opened 2026-05-13
ci: declare contents: read in four pure-CI workflows

arpitjain099:
Four workflows currently run with no  block, so  picks up whatever the repository default is. Each of them only checks out the repo and runs a Python validator (gftools / knowledge_graph.py / designer profiles / pushlist linter). None push commits, comment, or call any GitHub write endpoint.

`scorecard.yml` in this repo already declares `permissions: read-all`; this PR brings the four data-check workflows in line with that explicit-scope convention.

YAML validated locally before commit.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,6 +16,9 @@ on:
     - "to_*.txt"
     - ".github/workflows/ci.yaml"
 
+permissions:
+  contents: read
+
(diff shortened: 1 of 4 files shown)


emmamarichal:
@m4rc1e can you have a look? thanks!

m4rc1e:
Thanks and sorry for the delay.

PR #10528, opened 2026-05-14
Workflow fix - Uploard artifact

emmamarichal:
[CI is failing](https://github.com/google/fonts/actions/runs/25808417937/job/75949197647?pr=10527), I tried to fix it by changing artifact versions (using Claude)

```
Run actions/upload-artifact/merge@v4
Found 0 artifact(s)
Error: No artifacts found matching pattern 'qa-*'
```

@m4rc1e or @simoncozens can you have a look?

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -66,7 +66,7 @@ jobs:
 
       - name: Upload check results
         if: steps.check_files.outputs.files_exists == 'true'
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@v4
         with:
           name: qa-${{ matrix.os }}
           path: out/
@@ -128,7 +128,7 @@ jobs:
 
       - name: Upload check results
         if: steps.check_files.outputs.files_exists == 'true'
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@v4
         with:
           name: qa-diffenator
           path: out/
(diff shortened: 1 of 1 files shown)


simoncozens:
I'm not sure this is the correct fix. The new version of the action is fine; the question is, why are there no artefacts beginning with qa-? Looking in the "Diffenate Fonts" job, it seems like no artefacts were uploaded because the `out/` directory was empty:

Screenshot 2026-05-15 at 09 13 19

Looking further back in that job, the "Run diffenator and fontspector" section doesn't have any diffenator or fontspector output, only the output from the `youseedee` call; and it ran in 4 seconds. It didn't do anything!

Screenshot 2026-05-15 at 09 17 08

When I run the `.ci/run.py` line myself on my computer, the first thing I see is:

```
$ python3 .ci/run.py --pr-number 10527
Checking new family: ofl/geomini
```

So we should see that output at least. But we don't. So this means that it didn't detect the family as being new or changed. I'm not sure why. But the problem is likely to be either in what GitHub is checking out, or in `.ci/whatchanged.py` which works out what is different about the branch.

emmamarichal:
@simoncozens I don't know if it can help, but it failed here too: https://github.com/google/fonts/pull/10507

m4rc1e:
Closing. Fixed in https://github.com/google/fonts/commit/0d746b26076ef8b2fe03907559beb0d940f082f9. Updating checkout to v6 doesn't fetch all branch and tags by default so you need to explicitly enable it.

PR #10541, opened 2026-05-20
[Knowledge] Add check for URLs

emmamarichal:
@m4rc1e can you review that?
Unlike the others, which are .yaml files, this one is a .py file that isn't linked to CI. Do I need to add something to CI for the actions to trigger? Because when I create a PR, the checks don't show up.

### DIFF
--- a/.github/workflows/knowledge_graph.py
+++ b/.github/workflows/knowledge_graph.py
@@ -1,10 +1,11 @@
 from pprint import pprint
+
 from absl import app
 from absl import flags
 from gftools import knowledge_pb2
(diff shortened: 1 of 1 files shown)


simoncozens:
Hi Emma, this looks fine apart from one thing. There are a lot of changes which are just adding or removing blank lines. We use the `black` tool to keep Python code formatted consistently. Run `black knowledge_graph.py` and then commit the result.

emmamarichal:
@simoncozens, ah yes sorry, it's corrected!

simoncozens (approved):

PR #10542, opened 2026-05-20
Renaming Huerta tipografica to HT Fonts

andrestelex:
After some years, our foundry changed its name to HT Fonts, has a new website url (https:///www.htfonts.com) and new designers emails under the same domain.

We take this oportunity to
- Update all mentions to Huerta Tipografica to HT Fonts
- Update Carolina Giovagnoli's bio text and picture (to be in line with www.htfonts.com)
- Add Andres Torresi bio and links to it
- Add all relevant links to HT Fonts as designer
- Remove some entries linking Sol Matas to Huerta Tipografica (she hasn't been working with us for a while)

The links to github.com/huertatipografica are preserved as we haven't migrated that yet.
We will do a new pull request when we are ready to update that.

The metadata information within the fonts is not updated, but our intention here is to just update the info in Google Fonts catalog. We will take opportunity to update fonts when we update them.

I'm doing this with a pull request to be clear with my intentions, but feel free to reject it and propose other method.
I think this is more clear than opening an issue explaining what we want to do.
But I'm not sure if the update is fully correct. Just let me know.


davelab6:
Looks great! Thanks 

Would you be open to remove the RFNs?

emmamarichal:
Hi @andrestelex,
Thank you very much! 

Actually, opening an issue with this to-do list would be ideal; I could create sub-issues based on the type of change (designer, OFL, etc.). I think it’s better to have separate PRs based on the files we’re changing, that will make things clearer. 

I’m closing this PR for now, but I’ll reuse what you’ve done. 
I’ll take this opportunity to review it and see if there are any changes needed to align with our latest specifications (like removing the RFN, for example).

Let me know if you have any questions, and feel free to ping me in the issue and mention this PR there :)

andrestelex:
Thanks Emma, and sorry if I complicated things.
I will create the issue with the list of changes and mention this PR. 
Also mention that we are open for removing RFNs @davelab6 (I need to double check with the team today , but I think it's yes. I'll add tot he list. Mentioning this in case we need to do something else apart from the code.

PR #10562, opened 2026-05-26
Update skip.csv

evanwadams:
Fix some errors in skip.csv with the wrong name being used for some fonts:

Playwrite Argentina Guides -> Playwrite AR Guides
Playwrite Österreich Guides -> Playwrite AT Guides Noto Emoji Color -> Noto Color Emoji
Edu Australia VIC WA NT Hand Arrows -> Edu AU VIC WA NT Arrows Edu Australia VIC WA NT Hand Dots -> Edu AU VIC WA NT Dots

Deleted Edu SA Dotted Hand Guidelines


m4rc1e:
@evanwadams I can happily delete these from families.csv if you'd like? I can also do a pass over every tag and remove any for families that we don't serve.

PR #10600, opened 2026-06-06
feat: add script and taxonomy tags for Noto Sans Tirhuta

RedSaurb:
### Description
This PR explicitly registers script classification and taxonomy metadata for the 'Noto Sans Tirhuta' family within the tags collection layout. 

### Justification
While 'Noto Sans Tirhuta' already possesses expressive and quality metric tags within `families.csv`, it currently lacks explicit script categorization hooks (`/Script/Historic` and `/Script/Tirh`). 

Because the Tirhuta script handles the Maithili language—which historically diverges from standard regional Devanagari automation pipelines—adding these specific taxonomy markers is necessary. This configuration change explicitly maps the font family properties to the ISO 15924 'Tirh' script tag designation and ensures correct indexing for upstream platform compilation targets (such as system-wide font fallback configurations in AOSP).

### Changes Matrix
- Added `/Script/Historic` tag to classify the family under the historical scripts collection matrix.
- Added `/Script/Tirh` tag to programmatically link the family asset to the core Unicode script block definition.

### DIFF
--- a/tags/all/families.csv
+++ b/tags/all/families.csv
@@ -16015,6 +16015,8 @@ Noto Sans Tirhuta,,/Quality/Drawing,70
 Noto Sans Tirhuta,,/Quality/Spacing,70
 Noto Sans Tirhuta,,/Quality/Wordspace,100
 Noto Sans Tirhuta,,/Sans/Humanist,100
+Noto Sans Tirhuta,,/Script/Historic,100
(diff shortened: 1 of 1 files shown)


emmamarichal:
Hello @RedSaurb, thank you for your submission, but unfortunately I can't accept it since these tags are not registered.
Have a great day!

Issue #10573, opened 2026-06-01
Add Hibur Mono


simoncozens:
**Font Project Git Repo URL:**

https://github.com/typehabesha/HiburMono

**Super short description of the Font Family:**

Hibur Mono is a monospaced styled font for the Ge’ez (Ethiopic) script. Latin letters of Hibur Mono have been borrowed from the Noto Sans Mono font. Many of the Hibur Mono geometric properties are also conforming to the Noto Sans Mono design.

**Requirements:**

By opening this issue, I confirm the project meets the following requirements:

- [x] The entire font project is available in a Github repository (repo) and licensed under the [OFL](https://openfontlicense.org/open-font-license-official-text/)
- [x] The source files are available in the repo
- [x] I am the sole copyright author of the entire project, or all other copyright authors have licensed their work to me under the OFL, and I commit to clearly disclosing if AI tools were used in the creation of this project.
- [x] There are no "Reserved Font Names" in the OFL license information, or in the project documentation of any known upstream projects. If there are RFNs, they are not used in whole or in part in this family name, or, I want to discuss how Google can work with my use of them.
- [x] The family name is unique according to [namecheck.fontdata.com](https://namecheck.fontdata.com/)
- [x] The name of the font family expected to appear on app menus must be very clearly communicated and definitive. It should not include any copyright holder's full names or acronyms.
- [x] The font supports at least the Google Fonts 'Latin Core' glyphset from [github.com/googlefonts/glyphsets](https://github.com/googlefonts/glyphsets) ([direct link](https://github.com/googlefonts/glyphsets/blob/main/data/results/txt/nice-names/GF_Latin_Core.txt))
- [x] The repo has the [Google Fonts preferred upstream repo structure](https://googlefonts.github.io/gf-guide/upstream.html)
- [x] I have read, agree with, and comply with, the full [Google Fonts contributing requirements](https://googlefonts.github.io/gf-guide/index#pre-production-getting-your-fonts-ready-for-gf)
- [x] I will maintain the repository and participate in the onboarding process (addressing, solving, and responding to issues, merging pull requests, etc)

**Image:**

![](https://typehabesha.github.io/HiburMono/documentation/image1.png)

dyacob:
Thank you for adding the issue. We are touching up the ARTICLE, and expect to create the PR within a day.

dyacob:
[...]

PR #10607, opened 2026-06-09
Add foundry: TypeHabesha

typehabesha:
Foundry for typeface Hibur Mono 
https://github.com/google/fonts/issues/10573


emmamarichal:
Hello @typehabesha! I just want to check with you, should we write Type Habesha, with a space, instead? Or is it intentional? It seems they are two different words :)

typehabesha:
> Hello @typehabesha! I just want to check with you, should we write Type Habesha, with a space, instead? Or is it intentional? It seems they are two different words :)

Hello, TypeHabesha was written intentionally to make it seemless and can be read in one go, please continue with TypeHabesha

emmamarichal:
@typehabesha, please check if the image is following the specs (same for Behailu Berento):
`Should be a PNG or JPEG image between 100px and 300px. It will be cropped as a circle, so the image must be submitted as a square (1:1).`

typehabesha:
> @typehabesha, please check if the image is following the specs (same for Behailu Berento): `Should be a PNG or JPEG image between 100px and 300px. It will be cropped as a circle, so the image must be submitted as a square (1:1).`

I realized the resolution is too high, Now I made a new push to it fixing the image

emmamarichal (approved):

PR #10612, opened 2026-06-10
Typehabesha

typehabesha:
Fixing update info.pb.
This should fix everything


emmamarichal:
already merged, I've put .png since the image in now a png

typehabesha:
> already merged, I've put .png since the image in now a png

Thank you so much.

PR #10615, opened 2026-06-10
Update families.csv VF tagging Fancy complete

bghryct:
There are only two "Fancy" vfs.


m4rc1e:
@bghryct This pr mainly touches vintage tags.

bghryct:
you're right this is wrong.

PR #10623, opened 2026-06-12
docs(montserrat): add Design Applications section to DESCRIPTION

webdevpraveen:
## Description

Adds a "Design Applications" section to the Montserrat font DESCRIPTION.en_us.html file.

## Change

Appends information about recommended use cases for Montserrat:
- Display typography
- Headlines & branding
- Posters & web design
- Modern, geometric design projects

## Rationale

Montserrat is one of Google Fonts' most popular families. The existing description covers design history and characteristics, but lacks practical guidance. This section helps designers understand optimal use cases.

## Type

Documentation

Signed-off-by: webdevpraveen <pr4veensingh@proton.me>

### DIFF
--- a/ofl/montserrat/DESCRIPTION.en_us.html
+++ b/ofl/montserrat/DESCRIPTION.en_us.html
@@ -17,3 +17,7 @@ In fall, Julieta Ulanovsky, Sol Matas, and Juan Pablo del Peral, led the develop
 The Montserrat project is led by Julieta Ulanovsky, a type designer based in Buenos Aires, Argentina. 
 To contribute, see <a href="https://github.com/JulietaUla/Montserrat">github.com/JulietaUla/Montserrat</a>
 </p>
+
+
+<h3>Design Applications</h3>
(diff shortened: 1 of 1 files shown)


emmamarichal:
Hi @webdevpraveen, thanks for the initiative, this is a thoughtful contribution and we appreciate the effort.

That said, recommended use cases are already covered by the catalog's tags. The filtering system on Google Fonts lets users select families precisely by use case (display, business, etc.), so adding a "Design Applications" section to the description would duplicate information the tags already provide.

There's also a broader concern: adding this kind of section to one family invites the same change across every description in the catalog, and maintaining that consistently isn't where we want to invest time, especially since the tags are designed to handle exactly this.

For those reasons we'll pass on this one, but thanks again for taking the time to contribute! :)

webdevpraveen:
Hi @emmamarichal,

Thank you for taking the time to review my PR and explaining the rationale so clearly. That completely makes sense—I didn't realize the tagging system was meant to handle this use case exclusively. I appreciate the feedback and the amazing work you all do maintaining the catalog!

I would love to contribute to the repository in other ways if there are any beginner-friendly areas or documentation tasks you’d recommend focusing on. Thanks again! :)

PR #10638, opened 2026-06-17
Update gflanguages to v0.7.8

simoncozens:
This PR updates the `gflanguages` subtree with the latest release (v0.7.8) from [googlefonts/gflanguages](https://github.com/googlefonts/gflanguages). **Important:** When merging, use the **"Create a merge commit"** button, **not** "Squash and merge", to preserve subtree history and avoid future merge conflicts.


Lib/gflanguages/data/languages/ak_Latn.textproto
moyogo, line 1: @simoncozens This is a duplicate of fat_Latn, with samples with cut words or different samples, beside the name and ISO 639 code.
The rationale for not including it in the first place was likely that fat or twi are more appropriate.

PR #10639, opened 2026-06-17
EB Garamond v1.003

emmamarichal:
previous update never pushed, because of a version issue. 
this PR bumps version to 1.003


emmamarichal:
@m4rc1e 
I see some errors here: 
Screenshot 2026-06-17 at 17 52 48

PR #10642, opened 2026-06-18
Update Kazakh examples

MuratKaribay:
Please update the data and examples for Kazakh, specifically for the Arabic and Latin variants, since I am the creator and contributor of kk-Arab in the Unicode CLDR, and having noticed some errors, I have corrected them.


emmamarichal:
@simoncozens what do you think? :)

simoncozens:
@MuratKaribay Thanks so much for this! I will need to replicate this in the "googlefonts/lang" repository since this is where we pull that data from. Can you confirm why the Latin text should be all uppercase? Can you show me some examples of this in use?

MuratKaribay:
@simoncozens This is especially important for languages that rely on Turkic localized forms such as `i.loclTRK`.

In uppercase-only or caps-style fonts, such as Bebas Neue and similar typefaces, the dotted and dotless I forms are often not displayed as distinct forms. As a result, users of languages such as Turkish, Azerbaijani, Crimean Tatar, and Kazakh cannot properly distinguish between `İ/i` and `I/ı`.

simoncozens:
OK; but that's a problem with the fonts, and we can test for that and fix it. My question is, should text in this language be displayed in uppercase normally?

MuratKaribay:
To be honest, there are no rules in the Kazakh language that require everything to be in uppercase, and since users don't have a Kazakh Latin keyboard, I think it would be better if the text were in both uppercase and lowercase. If fonts start displaying all `locl` variants.

And if possible, I'd like to use text that includes all the letters from the extended Latin alphabet.

PR #10658, opened 2026-06-26
Update to_sandbox.txt

m4rc1e:


dyacob:
Seeing the CLA error, I'm speculating that it might stem from Behailu Berento creating two CLAs, one for himself as a designer, and for his foundry (TypeHabesha) with separate email addresses.  But then  creating the two PRs using only a single github account account that uses the foundry's email address.

If this is the case, what do you recommend?  Should Behailu (@typehabesha ) change the email address in the designer CLA to match the address used in the github account?  The designer and foundry contact email addresses would then be the same (which might be common anyway).

dyacob:
**UPDATE:** 

- Behailu added his designer account as a secondary email address to his @typehabesha github account. 
- The two CLAs were updated to reference the _same_ github account while still having _separate_ email addresses.
- Behailu reran the CLA scan, from the link provided by the error message. However, after 1.5 hours  (now 2+ days), it remains running when a normal run takes 6s.  Perhaps it got hung on a permissions issue or is stuck in a loop?

PR #10674, opened 2026-07-03
Update families.csv

emmamarichal:
Added new expressive styles for multiple font families including Akt, Alien Block, Estedad, Finlandica Headline, Geist Pixel, and M PLUS U.

@m4rc1e


m4rc1e:
Thank you!

PR #10689, opened 2026-07-11
Update test_font_tags.py

aaronbell:
In adding a bunch of new tag data to the `families.csv`, I realized that there are situations where a tag may be active at the default location (MONO@0), but be inactive at a different location (MONO@1). "Monospace/Monospace" is a great example which can be set to 100 at the default location but needs to be set to 0 at other instances on the variable axis. 

This PR implements a change to the test_font_tags file to allow this to occur. It is required for https://github.com/google/fonts/pull/10688 to move forward.


aaronbell:
I pushed the commit to the wrong place. Reverted. PR is now only related to the zero state update.

m4rc1e:
I see the issue but I'm not too sure if this is the best solution. I can raise this in +oncall.

PR #10711, opened 2026-07-17
quant.csv: Add min and max pt size

m4rc1e:
cc @davelab6


davelab6:
@m4rc1e Please rename `pt` to `opsz`, e.g. from

    Family,,/quant/pt_min,5
    Family,,/quant/pt_max,24

to

    Family,,/quant/opsz_min,5
    Family,,/quant/opsz_max,24

and merge & queue to push :)

PR #10726, opened 2026-07-23
Update MaShanZheng-Regular.ttf

### LINKED ISSUE #10722: [Ma Shan Zheng] U+6DC0 (淀) glyph is visually confusable with U+7EFD (绽)
Ma Shan Zheng Version 2.002 contains an ambiguous glyph for U+6DC0 (淀). In words such as “白洋淀”, the character is easily misread as “绽”.

**Describe the bug**

In Ma Shan Zheng Version 2.002, U+6DC0 (淀) is visually confusable with U+7EFD (绽).

The font’s cmap entries are distinct and correct (`U+6DC0 → uni6DC0` and `U+7EFD → uni7EFD`), so this is not a Unicode mapping error. The problem appears to be the outline of `uni6DC0`: its left-side 氵 component looks too similar to 纟, causing “白洋淀” to be read as “白洋绽”.

**To Reproduce**

1. Download the current `MaShanZheng-Regular.ttf`.
2. Render `淀` (U+6DC0) and `绽` (U+7EFD) side by side at the same font size.
3. Render the place name `白洋淀`.
4. Observe that the final character can easily be read as `绽`.

**Expected behavior**

The U+6DC0 glyph should have a clearly recognizable 氵 component and remain visually distinct from the 纟 component in U+7EFD, while preserving the typeface’s brush style.

**Screenshots**

The attached image directly renders U+6DC0 and U+7EFD from the original TTF at the same size, without system-font fallback.

**Additional context**

- Font: Ma Shan Zheng Regular
- Version: 2.002
- TTF SHA-256: `b844c59bf20bf530e41c20d6ff12b383b23a2e553b9b68cc89f070869213155d`
- Verified against the upstream `googlefonts/mashanzheng` TTF, the current `google/fonts` TTF, and a local copy; all three files are byte-identical.
- Reproduced in Chromium on macOS using the original TTF loaded directly through `@font-face`.
- U+6DC0 and U+7EFD map to distinct glyph outlines. This report therefore concerns glyph design and readability, rather than character encoding.

Image

aaronbell:
Fixing a single glyph (6DC0) in response to: https://github.com/google/fonts/issues/10722#issuecomment-5047782865


aaronbell:
The CL is pretty tiny, just a modification to the single glyph. No other changes made to the font.

emmamarichal (approved):

PR #10728, opened 2026-07-23
Yomogi Primary Script

aaronbell:
Changing primary script from Hiragana to Japanese, as it has broad Kanji coverage.

### DIFF
--- a/ofl/yomogi/METADATA.pb
+++ b/ofl/yomogi/METADATA.pb
@@ -35,4 +35,4 @@ source {
   branch: "ver3.00"
   commit: "8551d68706679b88ce6ff40e093a993e546e593e"
 }
-primary_script: "Hira"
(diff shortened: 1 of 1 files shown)


emmamarichal (approved):

emmamarichal:
Screenshot 2026-08-07 at 12 35 51

@aaronbell I think we need to fix the primary script, the new one is rendering this in the sandbox!

aaronbell:
It looks like the version that we have is out of date and is missing the 々 glyph, which upstream has. The notdef character has also been modified though, so I'm going to submit a PR to fix that, then onboard the updated font. Nate also suggested switching these to use "primary_language", which I think I will do moving forward.

aaronbell:
https://github.com/satsuyako/YomogiFont/pull/17

PR #10732, opened 2026-07-23
update name designer

emmamarichal:
small update= Fontrix -> RixFont


emmamarichal:
@aaronbell, name updated, according their last email :)

Issue #7056, opened 2023-12-04
Audit of Japanese fonts


simoncozens:
Some of our Japanese fonts display kana-only (which might be OK because they are kana only fonts, but also might indicate a problem with kanji coverage). Some of them display Latin sample text. I'm proposing to restrict the `ja_Jpan` exemplar kanji to [Joyo kanji only](https://github.com/googlefonts/lang/pull/130), but I'm going to use this issue to document whether or not that would actually fix the problem.

* [Noto Sans Japanese](https://fonts.google.com/noto/specimen/Noto+Sans+JP) shows a kana-only specimen when the `subset` filter is set to Japanese. This is wrong, and I'm not sure why this is the case especially given that [Noto Serif Japanese](https://fonts.google.com/noto/specimen/Noto+Serif+JP) does not. METADATA.pb languages and primary script are the same in both fonts.
* Display fonts. The following fonts show kana-only in the samples, but they are kana-only fonts so that's *fine*:
  * [Shizuru](https://fonts.google.com/specimen/Shizuru)
  * [Aoboshi One](https://fonts.google.com/specimen/Aoboshi+One?subset=japanese)
  * [Monomaniac One](https://fonts.google.com/specimen/Monomaniac+One)
  * [Tsukimi Rounded](https://fonts.google.com/specimen/Tsukimi+Rounded?subset=japanese&noto.script=Hira)
  * [Cherry Bomb One](https://fonts.google.com/specimen/Cherry+Bomb+One?subset=japanese&noto.script=Hira)
  * [Chokokutai](https://fonts.google.com/specimen/Chokokutai?subset=japanese&noto.script=Hira)
  * [Darumadrop One](https://fonts.google.com/specimen/Darumadrop+One?subset=japanese&noto.script=Hira)
  * [Slackside One](https://fonts.google.com/specimen/Slackside+One?subset=japanese&noto.script=Hira)
  * [Palette Mosaic](https://fonts.google.com/specimen/Palette+Mosaic?subset=japanese&noto.script=Hira)
  * [Yuji Hentaigana Akari](https://fonts.google.com/specimen/Yuji+Hentaigana+Akari?subset=japanese&noto.script=Hira) (Specialist, non-Unicode font)
  * [Yuji Hentaigana Akebono](https://fonts.google.com/specimen/Yuji+Hentaigana+Akebono?subset=japanese&noto.script=Hira) (Specialist, non-Unicode font)
[...]

PR #10735, opened 2026-07-23
Update ja_Jpan.textproto

aaronbell:
Updating Japanese text strings to ensure that Kanji and Kana are present for Noto fonts


aaronbell:
@simoncozens Referencing your old Japanese font audit (#7056) I'm updating the strings for Japanese to include both Kana and Kanji. LMK if you'd suggest other strings.

simoncozens:
I don't understand this change - both before and after samples contain kana and kanji.

aaronbell:
@simoncozens If you look at https://fonts.google.com/noto/specimen/Noto+Sans+JP?preview.layout=grid, the top specimen only shows Hiragana "すべての". So this change is making sure there is Kanji & Kana present in the first 4 characters to ensure the top specimen shows both in all scenarios.

simoncozens:
Only problem now is that changes to lang should be made in googlefonts/lang (which gets imported into here). But I can make the change.

aaronbell:
@simoncozens I've submitted the PR over in googlefonts/lang for this (and hawaiian). Can you get it merged and sent over here?

PR #10775, opened 2026-08-05
Update families.csv Dancing Script Formal Tag

### LINKED ISSUE #10770: Dancing script should be tagged as Calligraphy
**Describe the bug**

Dancing script should be tagged with Calligraphy, presumably, Formal.

bghryct:
This is in response to https://github.com/google/fonts/issues/10770  and should solve it.


m4rc1e:
These tags are not ordered!

m4rc1e:
@bghryct if you're going to use your tagger, please make the output a 100% match against the official. This is the absolute ball ache of rolling your own. We now have two stupid taggers to maintain.

PR #10816, opened 2026-08-18
Ogonek 64 Sans: Version 1.003 added

arekbr:
## Ogonek 64 Sans

New pixel typeface on an 8x12 grid, covering the full **GF Latin Core** glyph set (319 glyphs) with complete Polish diacritics — which 8-bit character sets never had.

**Source repository:** https://github.com/arekbr/ogonek64
**Designer:** SMOK - Arek Bronowicki
**License:** SIL OFL 1.1, no Reserved Font Name
**CLA:** signed

### QA

`fontbakery check-googlefonts` — **0 FAIL** across 455 check executions.

Remaining WARNs are inherent to a pixel design (`contour_count`, `unreachable_subsetting`) or resolved during onboarding (`designer_profiles`).

### Design notes

Accents occupy two dedicated rows plus one of clearance. In a single row of eight pixels the circumflex, caron, breve, ring and diaeresis are indistinguishable from one another, so the cell is twelve pixels tall rather than eight.

Accent placement is computed from each letter's actual top edge, which keeps capitals, ascenders and x-height glyphs consistent; the accent over `i`/`j` replaces the dot, as convention requires.

Glyph sources are plain text files, one line per pixel row, so the build is reproducible and reviewable without a font editor.


emmamarichal:
Hello @arekbr!
Thank you very much for your submission, we appreciate the initiative of preparing a PR! :)
However, the submission process is a bit different: you need to open an[ issue first](https://github.com/google/fonts/issues), and then we review the font to decide whether to add it to the catalog. If the decision is favorable, we then carry out a full QA process and make any necessary modifications. Don't hesitate to reach out if you have any questions, I'll be happy to help!

I'm closing the other issues you've opened as well.

PR #10818, opened 2026-08-18
Delete tags/tags_metadata.csv

davelab6:
Afaik this file is obsolete


m4rc1e:
This file is used in the backend since I've been asked to add some tags to it so I'm going to close this.

PR #10823, opened 2026-08-19
Update gflanguages to v0.7.10

simoncozens:
This PR updates the `gflanguages` subtree with the latest release (v0.7.10) from [googlefonts/gflanguages](https://github.com/googlefonts/gflanguages). **Important:** When merging, use the **"Create a merge commit"** button, **not** "Squash and merge", to preserve subtree history and avoid future merge conflicts.


emmamarichal:
@simoncozens, small check from Claude: 

**1. cnr_Latn.textproto — Đ missing from the index**
The base field correctly lists D {DŽ} Đ E …, but the index field jumps straight from {DŽ} to E, skipping Đ. Since Đ is a distinct letter of the Montenegrin alphabet (present in the exemplar_chars), leaving it out of the index could throw off tools that rely on that field for sorting or glyph-coverage testing.

**2. cnr_Cyrl.textproto — no sample_text**
Unlike cnr_Latn.textproto (which has a full sample_text block), the Cyrillic version has none. This might be intentional (translation not ready yet), but it's worth double-checking it isn't an oversight before this language gets exposed to tools that consume that field.

**3. haw_Latn.textproto — worth double-checking: \n vs \\n**
In the new specimen_21 and specimen_16 lines, the rendering I fetched shows a double backslash (\\n), while everywhere else in the diff (including right above, in the same fields) it's a single \n. If this is real, it would mean the line break might no longer be parsed correctly in the textproto — but it could also just be an artifact of the HTML-to-text conversion from my page fetch rather than an actual change. I'd recommend checking directly on GitHub (Split view or the raw diff) before drawing any conclusions.

simoncozens:
That last one is a definite issue, don't merge this. I can't confirm (1) - as I don't have a Montenegrin dictionary to hand(!) - but I don't think we use the index anyway. (2) is fine.

PR #10964, opened 2026-09-14
Updating Manrope URL link

aaronbell:
Now points to https://github.com/googlefonts/manrope instead of a non-existent repo. 
All files updated (including fonts). Version string bumped.


emmamarichal:
There is still this FAIL:

```
🔥 FAIL METADATA.pb: Check URL on copyright string is the same as in repository_url field. (googlefonts/metadata/consistent_repo_urls)
🔥 FAIL Repository URL is https://github.com/googlefonts/manrope. But:
HTML description has 'href="[https://github.com/googlefonts/manrope">github.com/googlefonts/manrope](https://github.com/googlefonts/manrope%22%3Egithub.com/googlefonts/manrope).
' [code: mismatch]
```


But I think it's a false positive, because I don't see where the issue could be.

emmamarichal (approved):

